PR #33, opened 2017-02-13
First look at HFP and mSBC support

kuikka:
Hi,

Here is my first cut at HFP and mSBC support. Please take a look and let me know if this approach looks good to you.

It is not yet ready for merging, one big issue being that I haven't been able to make the mSBC output actually work. My headsets just ignore the frames I send them. The input through microphone does work as does standard CVSD audio.

### DIFF
--- a/src/io-msbc.c
+++ b/src/io-msbc.c
@@ -1,0 +1,239 @@
+/*
+ * BlueALSA - io-msbc.c
+ * Copyright (c) 2017 Arkadiusz Bokowy
+ *
+ * This file is a part of bluez-alsa.
+ *
+ * This project is licensed under the terms of the MIT license.
+ *
+ */
+
+#include "io.h"
+
+#include <ctype.h>
+#include <errno.h>
+#include <fcntl.h>
+#include <math.h>
+#include <poll.h>
+#include <signal.h>
+#include <stdlib.h>
+#include <string.h>
+#include <time.h>
+#include <unistd.h>
+#include <sys/eventfd.h>
+#include <sys/socket.h>
+#include <sys/types.h>
+
+#include <sbc/sbc.h>
+#include "hfp.h"
+#include "bluealsa.h"
+#include "transport.h"
+#include "utils.h"
+#include "io-msbc.h"
+#include "shared/log.h"
+
+#if ENABLE_MSBC
+
+#define SCO_H2_HDR_0             0x01
+#define MSBC_SYNC                0xAD
+/* We seem to get the data in 24 byte chunks
+ * even though the SCO MTU is 60
+ * bytes. Use the same to send data
+ * TODO: Figure out why.
+ */
+#define MSBC_MTU		24
+
+#if defined (SILENCE)
+uint8_t msbc_zero[] = {
+	0xad, 0x0, 0x0, 0xc5, 0x0, 0x0, 0x0, 0x0, 0x77, 0x6d, 0xb6, 0xdd,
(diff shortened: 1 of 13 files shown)


Arkq:
Hi,

As a first cut it looks great. From your previous comment, I though, that the mSBC is handled internally by the bluez, but right now I see, that there is a lot of work :)

Just few minor issues regarding coding style (but it is not really important - I will correct this while merging).  Regarding more important things (e.g. io-msbc.c), I will have to read some specs, because right now I don't know much.

src/io-msbc.c
```diff
@@ -0,0 +1,253 @@
...
+
+#include <sbc/sbc.h>
+#include "hfp.h"
+#include "bluealsa.h"
+#include "transport.h"
+#include "utils.h"
+#include "io-msbc.h"
+#include "shared/log.h"
+
+#if defined(ENABLE_MSBC)
```
Arkq: For feature preprocessor flags please use single "if" (e.g. `#if ENABLE_MSBC`). Using "if define" might lead to unintentional code inclusion if one sets `#define ENABLE_MSBC 0`. I know, that someone might use `-DENABLE_MSBC` as an argument for gcc, but since we are using autotools for configuration, it's not the case here.

src/io-msbc.c
```diff
@@ -0,0 +1,253 @@
...
+		}
+	}
+	memmove(sbc->dec_buffer, p, bytes_left);
+	sbc->dec_buffer_cnt = bytes_left;
+}
+
+struct sbc_state* iothread_initialize_msbc(struct sbc_state *sbc)
+{
+	if (!sbc) {
+		sbc = malloc(sizeof(*sbc));
```
Arkq: Why not using memory allocated by the caller? It would be more consistent with the sbc library. And the size of the sbc_state structure is constant, so one might allocate it on the stack.

src/io.c
```diff
@@ -1059,18 +1066,22 @@ void *io_thread_rfcomm(void *arg) {
...
+	struct at_command at;
 
 	struct pollfd pfds[] = {
 		{ t->event_fd, POLLIN, 0 },
 		{ t->bt_fd, POLLIN, 0 },
 	};
 
+	/* HSP only supports CVSD */
+	if (t->profile == BLUETOOTH_PROFILE_HSP_HS || t->profile == BLUETOOTH_PROFILE_HSP_AG)
+		t->rfcomm.sco->sco.codec = TRANSPORT_SCO_CODEC_CVSD;
```
Arkq: This value should be set during the transport creation, since it is the default (and the only) value for HSP.

src/transport.c
```diff
@@ -537,7 +537,14 @@ unsigned int transport_get_sampling(const struct ba_transport *t) {
 	case TRANSPORT_TYPE_RFCOMM:
 		break;
 	case TRANSPORT_TYPE_SCO:
-		return 8000;
+		switch (t->sco.codec) {
+			case TRANSPORT_SCO_CODEC_UNKNOWN:
+				return 0;
```
Arkq: `break` would be more consistent here

src/utils.c
```diff
@@ -96,7 +96,7 @@ int hci_devlist(struct hci_dev_info **di, int *num) {
  * @param ba Pointer to the Bluetooth address structure for a target device.
  * @return On success this function returns socket file descriptor. Otherwise,
  *   -1 is returned and errno is set to indicate the error. */
-int hci_open_sco(const struct hci_dev_info *di, const bdaddr_t *ba) {
+int hci_open_sco(const struct hci_dev_info *di, const bdaddr_t *ba, int transparent) {
```
Arkq: Please, use `bool` type, since this argument is a simple feature flag.

PR #44, opened 2017-04-19
Add --disable-payloadcheck configure option for Pulseaudio compatibility

guillaumezin:
This way, no need to keep pulseaudio branch


Arkq:
Squashed and merged. Thanks!

PR #83, opened 2018-01-27
fix possible usage of already finished pthread_t

dambaev:
transport_set_state() can use pthread_cancel() and pthread_join() on thread, that had been already finished. At least, with musl-libc this will cause segfault. 
Transport threads, in theirs turn, has release variable, which will be changed to NULL value when thread's cleanup routine will be executed, so this can be used as a sign, that thread had been already finished and no need to call pthread_join()

### DIFF
--- a/src/transport.c
+++ b/src/transport.c
@@ -691,8 +691,11 @@ int transport_set_state(struct ba_transport *t, enum ba_transport_state state) {
 	switch (state) {
 	case TRANSPORT_IDLE:
 		if (created) {
-			pthread_cancel(t->thread);
-			ret = pthread_join(t->thread, NULL);
+			if( t-> release != NULL)
+			{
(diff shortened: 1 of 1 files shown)


Arkq:
Hi,

Thanks for pointing that out (that the pthread_cancel() is called for already terminated thread). However, using `t->release` variable for some side effect logic might be somehow misleading and might cause errors in the future. There is already a "way" of telling, that the thread has been terminated - the `thread` variable should be set to `main_thread`. I've made proper changes in the [musl](https://github.com/Arkq/bluez-alsa/tree/musl) branch. Please, check if that branch will solve this issue for you.

dambaev:
Hi,

yes, I haven't noticed, that config.main_thread is accessable from cleanup routine: it is a better approach indeed. Checked and working fine

PR #101, opened 2018-04-23
rfcomm: extend indicators

wberrier:
some (old) bluetooth headsets require additional indicators.  One of
these headsets is a Plantronics 510.

Otherwise, the headset never requests indicator statii, halting state
transitions to the "connected" state.

I noticed that this headset requested the indicator statii when connected to an android phone.  When inspecting the bluetooth capture, this was the difference.  Doing the same in bluealsa fixed the issue.


Arkq:
Many thanks!

PR #95, opened 2018-03-28
Build and install shared code as a shared library

wberrier:
Thoughts?  Changes?

Issue #85, opened 2018-02-14
Sending RING Indicate to Connected Headsets


bcroxall:
Arkq,

We're using bluealsa to send audio to a connected headset with gstreamer but now need to be able to send a ring indicate from our application. I cannot find this functionality in your code but I do see the ability to send a custom AT command. As you are probably aware, to do this we need to send /r/nRING/r/n to the rfcomm file descriptor for each connected headset.

Is this something you plan to implement in bluealsa? If so, do you have an expected timeline to do so? If not, will you provide some guidelines on how to send a ring indicate command to any connected headsets from a separate application via bluealsa?

Thanks for your help.

Arkq:
Hi,

> We [...] need to be able to send a ring indicate from our application. Is this something you plan to implement in bluealsa?

Yes, I do have.

> If so, do you have an expected timeline to do so?

I'm afraid it is rather a distant future. Currently, I want to close/fix issues related to audio transport within HFP/HSP profiles.

> If not, will you provide some guidelines on how to send a ring indicate command to any connected headsets from a separate application via bluealsa?

My general idea is to expose RFCOMM via bluealsa itself as a "proxy" socket. By doing so, bluealsa will be able to filter audio related commands and forward unrelated ones. In this way, it will be possible to establish Service Level Connection, and allow other application to communicate with BT headset without the requirement of implementing RFCOMM "stack". But it is just an idea. I don't have any prof-of-concept implementation.

bcroxall:
Hey,

Thanks for the quick response. We'd really like to work through this. It's a requirement for our product.

I have a few questions.

By "proxy" socket, I presume you mean that an application would get a file descriptor for a device from bluealsa and the application would then "write" to that file descriptor with an api call to bluealsa, which would write to the actual file descriptor. Is that kind of what you are thinking? I'm wondering if it might look very similar to how bluealsa now registers a profile with bluez and gets a new connection callback with a file descriptor when a device is added. Could an application register for new connections of specific profiles with bluealsa and get callbacks with proxy file descriptors when those devices are added? We couldn't write directly with those descriptors but would make a call to bluealsa to write for us?

How would the application communicate with bluealsa? Would there be a network socket or would the application and bluealsa communicate over dbus or some other way?

Again, thanks for your help.

Arkq:
[...]

PR #102, opened 2018-04-23
Allow external rfcomm

wberrier:
NOTE: this pull request depends on pull request #95 , but I wanted to publish this to start getting feedback.  To review, only the latest commit should be evaluated, and not the commits from #95.

This allows sending arbitrary rfcomm commands to the device.  I think a corresponding (although unimplemented) feature would be able to subscribe to all rfcomm commands from a device.

One advantage of doing it this way (opposed to sharing a "raw" file descriptor) was:

* It's pretty simple
* it coincides with the daemon's own rfcomm communication

Thoughts/feedback is welcome.

A prior discussion happened in issue #85 .

### DIFF
--- a/Makefile.am.include
+++ b/Makefile.am.include
@@ -1,0 +1,1 @@
+LOCAL_BLUEALSA_INCLUDES = -I$(top_srcdir)/src/shared
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -2,12 +2,11 @@
 # Copyright (c) 2016-2017 Arkadiusz Bokowy
 
 bin_PROGRAMS = bluealsa
-SUBDIRS = asound
+# put . so current directory is built before subdirectory for dependencies
+# TODO: wouldn't be necessary with non-recursive make or meson
+SUBDIRS = . asound
 
 bluealsa_SOURCES = \
-	shared/ffb.c \
-	shared/log.c \
-	shared/rt.c \
 	at.c \
 	bluealsa.c \
 	bluez.c \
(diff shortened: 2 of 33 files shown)


Arkq:
Hi, I think that this design is good enough. Like you've said, it's pretty simple :)

My first thought, though, was, that with this design receiving response (and commands) might be troublesome, but maybe you are right, that subscribing for RFCOMM notification will be good enough.

There is one thing, that I have to think about for a while - the message queue. In general it is OK. However, the requirement of such a communication between threads, might indicate, that the simple eventfd synchronization is too simple right now. So, prior to merging this PR, a little refactoring around eventfd might be required.

Many thanks for your contribution!

PS.
Sorry for not responding properly to your #95 PR, however my current knowledge about creating libraries is not good enough for providing some constructive comments :/ I will have to research a little bit before proper PR evaluation :D

wberrier:
Hi,

Regarding the queue of rfcomm commands: I think the eventfd mechanism is ok as a notification of changes, much like a condition variable or semaphore would be used.

Originally this patch used a single string field for a command that was set in the ctl thread, and read by the rfcomm transport thread.  The issue was that sometimes the string would get set back to back and so the first command was never handled by the rfcomm thread.  Thus the queue so make sure no commands were lost.

No worries about #95.  It's a big change.  FWIW, they (the shared library and rfcomm mechanism) are working out great for my use case.  Let me know if I can help in any way.

wberrier:
Hello,

I see what you mean about the event mechanism now... just to also pass the command through the file descriptor.  Nicely done.  You'll have to forgive my comments about using a file descriptor as a condition variable :)

Your changes worked great.  I'll update #95 to rebase against the latest.

Thanks.

PR #117, opened 2018-06-07
fixed apt-x channel mode bits as per bluez

rxrbln:
Fixed apt-X channel mode bits as per Bluez and also to work with my Sennheiser Momentum 2.0 Wireless, as trying to use the incorrect JOINT_STEREO 0x1 was actually MONO as per hcidump and not configuring with my Sennheiser Headset.


Arkq:
Hi, thanks for this PR, however I've got some concerns.

First of all, how do you know that bit 0 (the lowest one) is for MONO? Do you have any specification for apt-X BT configuration? I'm not saying, that constants defined in the `a2dp-codecs.h` file are correct (this file was copied from bluez project), however wireshark reports these bits in the same way (but maybe wireshark also got this file from bluez project :D). The only thing I'm pretty sure is that bit 1 is for stereo :D, because I've got dump from headset (sony wh-1000xm2) which supports apt-X stereo and this headset set bit 1 only.

Second thing is, why have you removed dual_channel and joint_stereo? I've also got dump from a headset (sennheiser mm 400-x) which sets all 4 bits, so I'm pretty sure, that those bits exists in the specification and are responsible for something :)

Arkq:
OK, I know what has happened. When I was developing apt-X support I've copied channel mode definitions from wireshark (I thought that wireshark is right)... and apparently wireshark has messed up these bits.

https://archive.li/vxK1f
https://www.spinics.net/lists/linux-bluetooth/msg48539.html
https://github.com/edrosten/bluez/blob/master/profiles/audio/a2dp-codecs.h

rxrbln:
It did not work with my Sennheiser Momentum 2.0 when I tried to get it to work the other day: https://www.youtube.com/watch?v=mTxjfi2DPOA after debugging full circle thru all of this I spotted hcidump spitting out MONO, which made me wonder. The main bluez package only has the APTX_MONO and APTX_STEREO devices:  https://www.youtube.com/watch?v=tcV1xwbCcVk

Arkq:
Thanks again for debugging, PR applied!

PR #119, opened 2018-06-11
Fix memory leak in g_variant_builder

parthitce:
- Memory allocated using g_variant_builder_new needs to
be explicitly freed

Signed-off-by: Parthiban Nallathambi <pn@denx.de>


src/bluez.c
```diff
@@ -712,6 +713,8 @@ static int bluez_register_a2dp_endpoint(
 
 final:
 	g_free(path);
+	g_variant_builder_unref(properties);
```
Arkq: If code from the line [679](https://github.com/Arkq/bluez-alsa/pull/119/files#diff-8c15b7742dad8d123d1be2d1c3954505R679) will be executed, then we've got g_variant_builder_unref() called with an uninitialized variable. And this comment applies to all your changes.

Many thanks for pointing out that the memory allocated with the g_variant_builder_new() needs to be freed. However, I think, that in this case it will be better to call g_variant_builder_init() instead - it is simpler and avoids heap memory allocation.

parthitce: @Arkq Yes, stack variable should be fine. Shall I change it to g_variant_builder_init ?

Arkq: Yes, if you might :)

parthitce:
@Arkq I meshed it up, sorry. I will update it again. Can I open fresh PR without merge commits? or updating this is fine?

Arkq:
Force push your branch so this PR will be updated automatically. And one more thing. I think, that it is required to call g_variant_builder_end() upon retrieval of GVariant, otherwise memory allocated by the g_variant_builder_add() will not be freed :)

parthitce:
@Arkq Tried forced pushing, github auto closed it. Not sure why. Opened a new one https://github.com/Arkq/bluez-alsa/pull/120

PR #120, opened 2018-06-11
Use stack variable for g_variant building

parthitce:
- Use g_variant_builder_init instead of g_variant_builder_new to
use stack variable for variant building

Signed-off-by: Parthiban Nallathambi <pn@denx.de>

### DIFF
--- a/src/bluez.c
+++ b/src/bluez.c
@@ -253,13 +253,14 @@ static void bluez_endpoint_select_configuration(GDBusMethodInvocation *inv, void
 		goto final;
 	}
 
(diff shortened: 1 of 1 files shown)


Arkq:
PR applied.

PS.
Sometimes it is a good idea to actually run the code after changes has been made :)

parthitce:
@Arkq Sorry again, breaks something during runtime regarding the builder. Please revert it for clean tree, I will fix it and have a test again locally.

Arkq:
I think that revert is not required, I've already made required modifications. See the diff between your branch and master.

PR #133, opened 2018-07-25
Fix segmentation fault

joerg-krause:
Initialize `pcm` with NULL as this lines checks for `pcm`: https://github.com/Arkq/bluez-alsa/blob/master/utils/aplay.c#L384. Otherwise, this might result in a segmentation fault in the alsa library.


Arkq:
Many thanks for this PR! It was really an embarrassing oversight...

PR #134, opened 2018-07-26
Do not fail if recv returns less bytes than requested

joerg-krause:
`recv()` might return less bytes than requested, e.g. because it was interrupted by a system call. Therefore, do not fail and exit the main loop.

### DIFF
--- a/utils/aplay.c
+++ b/utils/aplay.c
@@ -617,6 +617,8 @@ usage:
 			continue;
 		if (ret != sizeof(event)) {
 			error("Couldn't read event: %s", strerror(ret == -1 ? errno : EBADMSG));
+			if (ret > 0)
+				continue;
 			goto fail;
 		}
 


utils/aplay.c
```diff
@@ -617,6 +617,8 @@ int main(int argc, char *argv[]) {
 			continue;
 		if (ret != sizeof(event)) {
 			error("Couldn't read event: %s", strerror(ret == -1 ? errno : EBADMSG));
+			if (ret > 0)
+				continue;
```
Arkq: Unfortunately, I'm not able to reproduce this issue, but...

Simple `continue` might not be good enough here. If the issue is that the `recv()` returns less bytes than requested (due to interruption), then special recovery will be required - reconstruction of the fragmented event structure. Simple `continue` might produce a bug (discarded event) which will be even more hard to spot.

joerg-krause: Thanks for looking into this!

It only happened for me on my `volume` branch. I guess you weren't able to reproduces this as your machine might not be slow enough :smile: So were you able to reproduce it now?

Arkq: Yes, I've looked into this some more and I found out that the problem is not EINTR but bad "design". I've seen strange behaviour in the past, but now I've got time and nerve to investigate further. But you're right, that I had to "slow down" my machine a little bit in order to see this issue :).

joerg-krause: I see! I hadn't the time to investigate this further, too. Was happy it just worked. So, thanks again for taking care of this issue. I do not have much time now, but I will tested it as soon as I can on my slow machine...

PR #140, opened 2018-09-10
Get bluez interface using NodeInfo introspection xml

parthitce:
Use GDBusNodeInfo to create static introspection xml
and extract the interface using g_dbus_node_info_new_for_xml.

This makes the code simpler and easily understandble from
introspection xml. introspection function order as in
bluez/doc/profile-api.txt and bluez/doc/media-api.txt

Signed-off-by: Parthiban Nallathambi <parthiban@linumiz.com>


Arkq:
What are the benefits of this (besides mentioned readability, which I'm kind of sceptical if such a verboseness is required)? Bluealsa implements interface defined by bluez, so if one wants to see how the interface should look like, one shall go to the bluez documentation (bluez/doc/...). Also, personally, I do not like incorporating XMLs in small C projects. For big clambered Java project it might be fine, though :D 

Whole g_dbus object registration is an internal wiring anyway, so (IMHO) if it is less visible the better. If one wants to "understand" the bluez interface one should look at bluez_endpoint_method_call() and bluez_profile_method_call() which are dispatching functions for all API calls.

Or maybe this PR is a part of some bigger picture. If so, please write some more, because it looks like some kind of XY problem :).

parthitce:
@Arkq There isn't any big benefits out of it except the readability and in place visibility of the actual bluez interface.

PR #150, opened 2018-10-11
Make transport protocol sampling rate a 32-bit field

juyrjola:
Otherwise playback fails with sampling rate of 96 kHz, because due to the integer overflow, ALSA thinks the actual sampling rate is around 30 kHz.

I've verified that this fixes the playback issue at least with LDAC (#104).

Thank you for the great work with `bluez-alsa`! I'm so happy being finally able to get rid of my headphone wires. :smile:

### DIFF
--- a/src/shared/ctl-proto.h
+++ b/src/shared/ctl-proto.h
@@ -152,7 +152,7 @@ struct __attribute__ ((packed)) ba_msg_transport {
 	/* number of audio channels */
 	uint8_t channels;
 	/* used sampling frequency */
-	uint16_t sampling;
+	uint32_t sampling;
 
 	/* Levels for channel 1 (left) and 2 (right). These fields are also
 	 * used for SCO. In such a case channel 1 and 2 is responsible for


Arkq:
WOW, many thanks for spotting that!

As a side note, I'm assuming you've got LDAC capable speaker/headphone :D Could you confirm that LDAC Source works for every sampling frequency and for every channel mode (you will have to manually "enable" particular configuration https://github.com/Arkq/bluez-alsa/blob/ldac/src/bluez-a2dp.c#L199)? If so, I will be able to finally merge LDAC support to master.

Also, if you are I willing to test some more, I'm trying to setup [LDAC ABR (Adaptive Bit Rate)](https://github.com/Arkq/bluez-alsa/commit/bc1dabc875e10c4934c53d9efbca79ae7f11e518). Right now it won't work because my assumption about the value returned by the ioctl(TIOCOUTQ) was not right. But when I make the code work (in theory), will you test it (probably next week)?

Issue #155, opened 2018-10-19
Couldn't get BlueALSA transport 


deadeert:
Hello, 

I just install everything this morning so I should have all the last versions. 
`$sudo bluealsa &  # starting manually `
`$alsamixer -D bluealsa # is working but saying that no command are available for this device. `
`$ aplay -D bluealsa:HCI=hci0,DEV=C0:28:HI:DD:EN:XX,PROFILE=a2dp /tmp/92645__cgeffex__organ-ending-tune.mp3`
`../../../src/asound/../shared/ctl-client.c:105: Connecting to socket: /var/run/bluealsa/hci0`
`../../../src/asound/../shared/ctl-client.c:219: Getting transport for C0:28:HI:DD:EN:XX type 1`
`bluealsa: ../../src/ctl.c:638: Received new connection: 10`
`bluealsa: ../../src/ctl.c:651: New client accepted: 10`
`bluealsa: ../../src/ctl.c:675: +-+-`
`bluealsa: ../../src/ctl.c:675: +-+-`
`ALSA lib ../../../src/asound/bluealsa-pcm.c:666:(_snd_pcm_bluealsa_open) Couldn't get BlueALSA transport: Aucun périphérique de ce type`
`aplay: main:828: erreur à l'ouverture audio: Aucun périphérique de ce type`
`bluealsa: ../../src/ctl.c:605: Client closed connection: 10`
`bluealsa: ../../src/ctl.c:675: +-+-`
`

I added the following in my dbus file  (dbus-1/system.d/bluetooth.conf):
` 
 <policy user='MYUSER'>
                <allow send_destination="org.bluez"/>
  </policy>
`

Regarding the same issues tickets I tried to switch to PROFILE to sco. Same problem..  

The request seems to be sent to bluealsa but it timeouts...

Any idea of what is going one ? 

Thanks.

beginner3456789:
I think I have a similar problem. I have used the bluez-alsa for about 6 months beginning with my build from sources and then from my voidlinux dist both without pulseaudio. I think the problem started after recently updating the alsalib from 1.1.6 to 1.1.7. The current 1.3.1_1 bluez-alsa voidlinux dist package and my latest github build from source connects to my bluetooth audio devices and then gives the following error shown with mplayer:
```

Opening audio decoder: [ffmpeg] FFmpeg/libavcodec audio decoders
[...]

PR #158, opened 2018-10-21
Move config files to /etc/alsa/conf.d

m-cz:
With the release of alsa-lib v1.1.7, the location for add-on configs
has changed to /etc/alsa/conf.d/.

See:
http://git.alsa-project.org/?p=alsa-lib.git;a=commit;h=93e03bdc2a3dcd5d12516f5de78e14d88a32ff2c

Also related to #155.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -56,6 +56,7 @@ PKG_CHECK_MODULES([GIO2], [gio-unix-2.0])
 PKG_CHECK_MODULES([SBC], [sbc >= 1.2])
 
 AM_CONDITIONAL([ALSA_1_1_2], [$PKG_CONFIG --atleast-version=1.1.2 alsa])
+AM_CONDITIONAL([ALSA_1_1_7], [$PKG_CONFIG --atleast-version=1.1.7 alsa])
 
 AC_ARG_ENABLE([aac],
 	[AS_HELP_STRING([--enable-aac], [enable AAC support])])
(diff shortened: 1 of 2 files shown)


Arkq:
Great, thanks for this PS. However, backward compatibility is required... To be honest, I'm not sure how to achieve that without keeping old and new version (`alsadatadir` and `alsaconfdir` variable). Otherwise, minimal requirement for `alsa` has to be set to 1.1.7. Any thoughts?

infirit:
For backwards compatibility, ask pkg-config for at least 1.1.7 and modify the default value of the ```--alsaconfdir``` option based on its answer.

Also it would probably be a good idea to change the name of the variable from ``asound_module_datadir`` as its only the path for the configuration file.

Arkq:
Yeah, I think you are right. So, we will use `--alsaconfdir`, which will fall-back to `$sysconfdir/alsa` or `$datadir/alsa` based on `AM_CONDITIONAL([ALSA_1_1_7], [$PKG_CONFIG --atleast-version=1.1.7 alsa])`. Also, `asound_module_data_DATA` -> `asound_module_sysconf_DATA` and `asound_module_datadir` -> `asound_module_sysconfdir` (for ALSA_1_1_7). @infirit, does it seem OK?

@m-cz will you update this PR and test whether it works for you? Then, I will test this with my setup (alsa-lib-1.1.2) and if it is OK, we are ready to go :D

m-cz:
Tested with alsa-lib-1.1.7. Any further suggestions are welcome.

Arkq:
With alsa-lib-1.1.2 works as well (after one typo fix :D). Squashed and merged.

PR #160, opened 2018-10-26
Add meson build system

infirit:
My attempt to port to meson from autotools. The project website is at http://mesonbuild.com/. The autotools suite is getting stale and is not keeping up with the times. It is slow, cumbersome to deal with and hard to debug when things go wrong.

 Typical workflow is,

* checkout source
* meson build -Daptx=true (see meson --help for optimisations, build types, warn levels etc)
* cd build
* ninja (add -v for more verbose output)
* ninja test

I can get the test to run for 3 but pcm and server fail which probably is because I miss the setup. If you can explain how this is supposed to work I can make sure they all pass.

Some things to note,
* Options got their enable/disable dropped as meson has better ways to deal with this.
* debug is a reserved word so had to make a new name for --enable-debug, feel free to suggest another.
* There are other ways to define options, it has lists with one or more choices. It may be good to explore them, see http://mesonbuild.com/Build-options.html

Let me know if you have question or thing you like changed.


infirit:
I don't use/know cmake so would not know how it is different to meson. But I can tell you why I like it.

* It has strong datatypes, integers, strings, arrays etc.
* It works for 99% of the things you will need. If you need something different its usually possible.
* Ninja is fast :rocket: 
* Its very python like in syntax and behaviour so very readable and object orientated.
* Its written in python which should be enough but sadly not for everyone :smile: 

Here is a write up from elementryos on their switch to meson away from cmake: https://medium.com/elementaryos/all-aboard-the-meson-future-hype-train-2b6c478b6b9e

>>The autotools suite is getting stale and is not keeping up with the times. It is slow, cumbersome to deal with and hard to debug when things go wrong.

>Very true :)

This is my primary reason, I was triggered by the recent alsa config dir change and how obnoxious it has to be dealt with in autotools. With meson its trivial to express.

>Simply put: https://www.rojtberg.net/1481/do-not-use-meson/

Following the logic why are we now programming in C while assembly is a perfectly fine standard. Why do we have different shells like bash or korn. All these competing standards and implementation are there because the existing ones are either not doing what is required or lacking in areas. The design perhaps makes certain things impossible to implement or are so obnoxious nobody is willing to deal with it anymore.

One could argue why does cmake exist there is already standard make which does something very similar and worked just fine for thousands of projects :wink:.

infirit:
Oh, and you can say no I don't want meson it's your project. I'll be perfectly contend as I had fun porting and learned from it, just don't make me learn cmake :smile_cat:

Arkq:
> Oh, and you can say no I don't want meson it's your project. I'll be perfectly contend as I had fun porting and learned from it, just don't make me learn cmake

Cool, I have no preferences for cmake over meson :D

I'm just thinking if such a change is required at all. I've just checked and I've got meson installed, so for me it's not an issue. I will learn new toll (cool). However, I'm concerned about others, who are not familiar with building tools, programming languages, etc. Meson adds new dependency - python. Since lots of users use this project on Raspbery, I'd like to know if this might be troublesome :/.

Quick comparison on my box:
- dev-util/meson pulled in by 4 packages (e.g. libinput, xorg-proto)
- dev-util/cmake pulled in by 56 packages
- sys-devel/autoconf pulled by ~280 packages

infirit:
Trusty comes with python 3.4 and meson wants 3.5 :disappointed:.  Travis has been working on xenial for years now and next year april trusty will be eol. I'll see if I can persuade travis to install python 3.5.

Arkq:
> I am more concerned about the failing test though

In the current autotools approach, there is an option `--enable-pcm-test`, the reason why it is there is that, that this test won't work on travis (by design). This rest requires bluez-alsa to be installed before running this test. Also, it requires patched alsa-lib version, because there are lots of bugs in the ioplug code (at leas in the 1.1.2 version). I've proposed patches on alsa mailing list, however they were not accepted (merged), nor reviewed. Recently, I've pulled master changes for alsa-lib and it seems that few of these bugs have been exterminated (I've seen conflicts with my patches, so I've dropped them). I will have to check why this test is still failing and maybe correct the test, or make changes to alsa-lib (ioplug PCM has not behaved like hw PCM from the alsa-lib API point of view, maybe it is now :D).

Other tests should pass, though.

> I'll update the gentoo live (9999) bluez-alsa ebuild to use meson. If you are using gentoo let me know and I'll post it in a gist.

Yes, I'm using gentoo. I'd be quite happy if you could share it somewhere :)

> Getting too frustrated with travis and remember why I moved to circleci. I don't think I'll be able to have travis get a recent version of python.

So CI change will be required too....

infirit:
I was able to drop the required meson version down to 0.21.0 which is in jessie. Also updated README build instruction and how to obtain meson. The gentoo ebuild is at https://gist.github.com/infirit/b303988d33415847f4db2ffec14eec77

infirit:
Using meson 0.21 was a disaster so target stretch and xenial instead. Updated README as well.

XECDesign:
I've recently evaluated the major build systems and ended up settling on Meson as well. As long as this works with the stretch version of Meson, there shouldn't be a problem in terms of Raspbian support.

As far ass cross-compiling goes, the support is no worse than any other build system's and probably much better than most.

Meson is used by systemd, mesa and a ton of other major projects (https://mesonbuild.com/Users.html). It seems like adoption is not slowing down either.

diizzyy:
I meant as far as various toolchains goes but at least keep it in parallel with gnu autotools or cmake.

infirit:
Its a small project so having to maintain two build systems is a bit much. What would be an option I think is to do the next release with both autotools and meson. Then drop autotools in master.

@Arkq I changed the debug handling a bit, moved a few things around and added some comments.

diizzyy:
This might be worth having in mind which also is my personal experience...
https://gitlab.com/gnutls/gnutls/issues/320#note_46321209

infirit:
Sure, cross compilations suck not just in autotools. Have you actually tried meson cross compilation?

Arkq:
Indeed it seems that few major projects are switching to meson (e.g. xorg-server), so if there is any trouble with cross compilations, I think that it will be resolved in the near future. Anyway, I will test this with Yocto/OpenEmbedded-based cross-compilation.

@infirit many thanks for this work. I haven't tried this new build system yet, however I think that in the next or the second next release I will merge this PR.

As for keeping autotools and meson alongside, hmm... nope :)

infirit:
I understand this is not wanted so closing.

Arkq:
It's not entirely true, that this is not wanted. It was parked for now, because of the objections that not every distribution can use meson. If something is not wanted or does not fit bluez-alsa architecture, I'm closing such PRs :) Anyway, your work will be still accessible via this PR (even though it's closed), so maybe one day I will dig it up :)

infirit:
Well you said below recently and you don't want to have both. So not sure how else to interpret it other than not wanted 😕 .

`There is a PR #160 for changing autotools to meson, but since there was a comment that not everyone can use it, I'm keeping autotools as it is.`

Arkq:
Yes, I don't want to have both, that's true. But in regard to this PR, maybe I should say "I'm keeping autotools as it is, *for now*." :D

PR #167, opened 2018-11-05
Prevent crashes when logging to syslog

kempniu:
As explained by the man page for `va_arg()`:

> If `ap` is passed to a function that uses `va_arg(ap,type)`, then the value of `ap` is undefined after the return of that function.

Meanwhile, when logging to syslog is enabled, `vlog()` reuses the same `va_list` variable for both the `vsyslog()` call and the subsequent `vfprintf()` call.  This causes crashes, try:

```sh
# touch /var/run/bluealsa/hci0
# bluealsa -S
```

This PR fixes the issue by passing `vsyslog()` a copy of the `va_list` supplied to `vlog()`.


Arkq:
Many thanks for spotting this! PR applied.

PR #172, opened 2018-11-22
Added support for HFP from Ofono

tbultel:
This brings HFP support over Ofono, using its DBus interface.

HFP-HF has been tested and both source and sink work. HFP-AG has not been tested yet.
The "transport" code has been slightly modified, in order to be able to have SCO without RFCOMM

The backend is robust to bluez-alsa restarts, in regard to Ofono, and is also robust to Ofono restarts.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -119,6 +119,14 @@ AC_ARG_ENABLE([test-pcm],
 	[AS_HELP_STRING([--enable-test-pcm], [enable PCM plugin test])])
 AM_CONDITIONAL([ENABLE_TEST_PCM], [test "x$enable_test_pcm" = "xyes"])
 
+# support for Ofono
+AC_ARG_ENABLE([ofono],
+	AS_HELP_STRING([--enable-ofono], [enable HFP over Ofono support]))
+AM_CONDITIONAL([ENABLE_OFONO], [test "x$enable_ofono" = "xyes"])
+AM_COND_IF([ENABLE_OFONO], [
+	AC_DEFINE([ENABLE_OFONO], [1], [Define to 1 if the support for Ofono is enabled.])
+])
+
 AC_ARG_WITH([alsaplugindir],
 	AS_HELP_STRING([--with-alsaplugindir=dir], [path where ALSA plugin files are stored]),
 	[alsaplugindir="$withval"], [alsaplugindir=$($PKG_CONFIG --variable=libdir alsa)/alsa-lib])
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -20,6 +20,12 @@ bluealsa_SOURCES = \
 	utils.c \
 	main.c
 
+if ENABLE_OFONO
+bluealsa_SOURCES += \
+	ofono.c \
+	ofono-iface.c
+endif
+
 AM_CFLAGS = \
 	@BLUEZ_CFLAGS@ \
 	@GLIB2_CFLAGS@ \
--- a/src/bluealsa.h
+++ b/src/bluealsa.h
@@ -48,6 +48,7 @@ struct ba_config {
 		bool a2dp_sink;
 		bool hsp_hs;
 		bool hsp_ag;
+		bool hfp_ofono;
 		bool hfp_hf;
 		bool hfp_ag;
 	} enable;
--- a/src/ctl.c
+++ b/src/ctl.c
@@ -88,7 +88,7 @@ static int _transport_lookup(GHashTable *devices, const bdaddr_t *addr,
 				}
 				break;
 			case BA_PCM_TYPE_SCO:
-				if ((*t)->type != TRANSPORT_TYPE_SCO)
+				if ((*t)->type != TRANSPORT_TYPE_SCO )
 					continue;
 				/* ignore SCO transport if codec is not selected yet */
 				if ((*t)->codec == HFP_CODEC_UNDEFINED)
@@ -285,8 +285,11 @@ static void ctl_thread_cmd_list_transports(const struct ba_request *req, int fd)
 		for (g_hash_table_iter_init(&iter_t, d->transports);
 				g_hash_table_iter_next(&iter_t, NULL, (gpointer)&t); ) {
 			/* ignore SCO transport if codec is not selected yet */
-			if (t->type == TRANSPORT_TYPE_SCO && t->codec == HFP_CODEC_UNDEFINED)
+			if (t->type == TRANSPORT_TYPE_SCO &&
+				t->codec == HFP_CODEC_UNDEFINED) {
 				continue;
+			}
+
 			_ctl_transport(t, &transport);
 			send(fd, &transport, sizeof(transport), MSG_NOSIGNAL);
 		}
@@ -666,7 +669,7 @@ static void *ctl_thread(void *arg) {
 			for (i = 0; i < BLUEALSA_MAX_CLIENTS; i++)
 				if (config.ctl.subs[i] & event.mask) {
 					const int client = config.ctl.pfds[i + __CTL_IDX_MAX].fd;
-					debug("Emitting notification: %B => %d", event.mask, client);
+					debug("Sending notification: %B => %d", event.mask, client);
 					send(client, &event, sizeof(event), MSG_NOSIGNAL);
 				}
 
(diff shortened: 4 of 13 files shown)


Arkq:
Many tanks for this PR. It indeed looks promising. I'll try to review it during this weekend, however right now I can say few things:

1. The usage of `__func__` in debug() calls. If the name of debugged function is desired, it should be added to the debug() macro definition, not to the call site. It generates a lot of boilerplate code.
2. Help message (in the main.c) lacks ofono version number (if it was desired, and I'm not saying that it should be), so the argument list is shifted.
3. The introduction of TRANSPORT_TYPE_OFONO_SCO. I do not know exactly why, but it seems too much for me (in almost every place it equals to TRANSPORT_TYPE_SCO) :D Maybe after thorough review it will make sense.

Extra thing to keep in mind:
I'd like the ofono code to be conditional (via preprocessor), however I will talk about this later, because I'm not sure yet whether ofono code or everything related to HFP/HSP.

tbultel:
Hi Arkq,
no problem, I wish it could be possibly get merged by next week, in 
order to simply bump the sha1 in our yocto meta.

While reviewing, you will notice that handsfree volume control is not 
supported yet.
For some reason, Ofono has a separate dbus interface to do so, that uses 
modems IDs instead of "cards" (phones),
and I have to figure out how to do the matching between them.
That is something that I could submit later in another pull request.

Cheers
Thierry

Arkq:
Hi,

I've took the liberty to make same changes by myself within this PR in order to speed up merging process (otherwise it would take many ping-pong-like conversations). In general this PR was OK, however there was some things to be fixed:

- general code cleanup (it has looked more like PoC, than merge-ready commit - badly formatted docstrings, mixed indentation, function names taken from other places without renaming, etc.)
- function reorganization within the ofono.c file in order to remove forward declarations
- I've placed all oFono DBus related strings and values in the iface.h file
- fixed many memory leakages (not freed glib error structs, not unrefed DBus replies, not initialized char* causing segfault, etc.)
- fixed SCO file descriptor leakage when there is no PCM client connected
- I've also prepared some generic framework for acquire/release handlers (in other commit), so this PR could be simplified

I hope you don't mind :)

Unfortunately, I was not able to test whether HFP-AG works... It seems, that it is required to configure modem within oFono in order to get AG profile. If you've got setup with AG configured, please test whether it works after my changes. Also, I've removed `connect_pending` flag, so right now there are warning messages that the operation is already in progress. I will think about some more elegant way of deciding whether to call "Connect" method, so for now I'm OK with this warning.

Once again may thanks for your work and making bluez-alsa better software!

tbultel:
Many thanks !
I do not mind you corrected it yourself, it was faster this way.
I had very little experience on dbus and glib before this work, I will 
pay more attention next time. I am happy to have succeeded in
doing something far more readable and simple than the old-style in pulse 
audio plugin.
I am just curious why you removed the ofono_unregister()

For information, I am currently working on having bluez-alsa build 
generate a bluealsa shared library.
All the functions in ctl_client.h are of interest for an application 
wanting to monitor the available transports (of both SCO and A2DP),
in order to start or stop PCM loops. I hope this kind of thing would 
also be of interest to you, the work is basically
most a matter of Makefile.am rather than C code, but I will have to 
change some include paths in order to have the packaging (make install)
work as expected. If you like the idea of that shared library, both the 
alsa plugins and bluealsa-aplay could rely on it in the future, instead of
being statically linked with the object files as they are today.

Cheers
Thierry

Arkq:
> I am just curious why you removed the ofono_unregister()

General idea is very simple. When DBus connection is terminated, every object registered in DBus is dropped as well (or should be dropped). That's exactly the case for `pkill ofonod` you've implemented. One cannot rely on others for cleanup procedures. This forces "us" (e.g. DBus, kernel itself, glibc) to implement cleanup actions upon e.g. connection lost or process termination (file descriptors are closed, heap memory is freed, mutexes are unlocked, etc.). So, most the time I'm exploiting this requirement by intentionally omitting cleanup upon main() return. One exception is filesystem, where cleanup is required (temporary files, named sockets are not unlinked). This eliminates code redundancy, hence saves some bytes.

> All the functions in ctl_client.h are of interest for an application wanting to monitor the available transports (of both SCO and A2DP), in order to start or stop PCM loops. I hope this kind of thing would also be of interest to you.

Yes, this kind of work is desired :) See this PR: #95 However, I'm not very eager to merge such PR right now, because providing library adds some extra responsibility. There will be some API and ABI which should be maintained then. I will merge such a feature when the ctl-proto.h and ctl-client.h will be mature enough. Also keep in mind that there is a possibility for #160 to be merged in the near future.

tbultel:
Hi again,

I am sorry but with the latest code on master, I am not able to get any 
SCO sound output.
What is your test case ?
My is as such:
Have a mobile phone always connected, launch bluealsa in command line

/usr/bin/bluealsa -p a2dp-sink -p hfp-ofono

In a separate shell,
bluealsa-aplay -v --profile-sco  00:00:00:00:00:00

Then start a phone call.

This works on my PR branch, I am investigating what is missing. I wonder 
if there is a chicken-and-egg or race condition around the "release" logic.
Namely, when the NewConnection event occurs, then it immediately 
releases the ofono fd, because
t->sco.spk_pcm.fd and  t->sco.spk_pcm.fd are not initialized yet, and it 
removes the transport, letting no chance to bluealsa-aplay to have it 
enumerated.

Thierry

tbultel:
Commenting out

/*    transport_send_signal(t, TRANSPORT_PCM_OPEN); */
at ofono.c line 326 fixes the issue.

Arkq:
My test approach is similar to yours. What architecture you are using? Because there indeed might be a problem related to memory barrier.As a quick test you might add volatile to bt_fd and remove compiler optimizations -O0 (still processor might rearrange instructions, however, it should suffice). I will try  to adopt transport mutex lock to this part of code, because I've omitted sco thread while I was adding lock for transport critical sections.

Arkq:
Yeap, however it is required :)

tbultel:
I do not understand why, if you wakeup the io thread it will immediately
close and forget the SCO transport. There is maybe something I am 
missing in the logic.

Do reply to your last question: I am running it on an UpBoard (x86_64)

Arkq:
> I do not understand why, if you wakeup the io thread it will immediately close and forget the SCO transport. There is maybe something I am missing in the logic.

The only place where the SCO file descriptor can be closed is `io_thread_sco()`. When the new transport is created in the `ofono_card_add()`, the `bt_fd` is initialized to `-1` and the thread is started (call to set_state(ACTIVE)). This "locks" thread_sco in the `poll()` until some event arrives (e.g. connection from the client). When `poll()` returns, there is a logic which adds new FDs to the poll (`pfds[].fd = t->bt_fd`) structure. When new SCO connection is initialized in `ofono_agent_new_connection()` it is required to wake thread_sco from the `poll()` call, so it could add SCO fd to the poll structure, which will allow proper SCO closing (prevents SCO fd leakage). The actual name of the signal might be indeed misleading (`TRANSPORT_PCM_OPEN`) - I will change that.

Sooo, finally the problem is, that when `poll()` wakes, thread_sco [doesn't know that the PCM speaker/microphone is opened](https://github.com/Arkq/bluez-alsa/blob/master/src/io.c#L1604) and closes SCO.

And now the question is, how bluealsa should behave when there is no client for reading SCO data, and why in your case, thread_sco sees `spk_pcm.fd == -1`?

Regarding the case where there is indeed PCM client, how bluealsa should act when new SCO connection arrives. I think that the most appropriate way is to let oFono know, that we are not able to handle audio, so it could release SCO connection - phone will not forward audio to the bluetooth. Or maybe it is OK to grab phone audio even though it will go straight to /dev/null so to speak. What do you think?

tbultel:
Many thanks,
this works now (I tested the speaker only) and I the code is far more 
readable.

killerkalamari:
Are ofono and phonesim required in order to play audio over hsp-ag, or was ofono added as an alternative to the built-in bluealsa support? I've been having playback trouble with hsp-ag and wasn't sure if that's just because I haven't installed ofono or whether I'm doing something else wrong.

xavi-b:
Hi,
It seems that @tbultel succeeded in making phone calls using bluealsa and ofono, do you have any information related to your ofono configuration, asoundrc configuration or anything else that could help me with my issue #385 ?

xavi-b:
@Arkq, do you happen to have any ofono/bluealsa/asound configuration files/commands that could help me solve my issue ?

dr-ni:
yes, some more info woul be nice. I could use both streams like this:

sudo su
export LIBASOUND_THREAD_SAFE=0
bluealsa -p hfp-ofono

and in another terminal:
arecord -D bluealsa | aplay
to receive stream

and in another terminal:
arecord | aplay -D bluealsa
to send stream

works, but huge delay  20s :-(

cat ~/.asoundrc
defaults.bluealsa.service "org.bluealsa"
defaults.bluealsa.device "XX:XX:XX:XX:XX:XX"
defaults.bluealsa.profile "sco"
defaults.bluealsa.delay 10000

PR #183, opened 2019-01-18
dbus: request a name on startup

tbultel:
when the service thread is ready, ask a name to dbus
The advantage of doing that is that client applications
can check dbus for bluez-alsa presence, without having
to implement any kinf of polling logic. Also, this
way, they can be notified on the service exit, to
perform any needed cleanup

Signed-off-by: Thierry Bultel <thierry.bultel@iot.bzh>


Arkq:
General idea is OK, but... there is one small problem in BlueALSA design...

Currently it is not possible to run bluealsa when there is no HCI available. The reason for this is that in my original concept (it hasn't changed) bluealsa shall be bound to a single HCI. It is possible that one has many BT devices. From the configuration point of view, system should allow every HCI to be configured differently. In fact, BlueZ allows that for A2DP with RegisterEndpoint, which can be called for every given BT device. Unfortunately, RegisterProfile is a global call, so once e.g. HFP profile is registered it is reported by every HCI. If I figure out how to force BlueZ to register HFP profile only for a given HCI I will change that.

So, considering design presented above, BlueALSA is not a single service, but rather a service _per_ HCI. Will it work for you, when the service name will be something like "org.bluez-alsa.hci0"? I'm not sure, whether using hci0, hci1 is a good idea, though. On my laptop, when I attach external BT dongle, my internal hci0 is visible as hci1, which is quite confusing :/

tbultel:
As suggested by Arkq, uses the hci interface name as a suffix in the dbus name

Arkq:
To be honest, I'm also feeling that this approach is some kind of XY
problem. Thierry, could you please explain why you need such signal? Also,
I was thinking for a while, and maybe the general solution to (your)
problem regarding ALSA and D-Bus signalling would be to implement "proxy"
between these two, e.g. also-dbus. It might catch signals from kernel about
added PCMs, and bluealsa might also integrate with this proxy. This will
provide universal solution for everyone who wants to integrate with  Linux
audio without PulseAudio. What do you think?

tbultel:
Hi,

The client application definitively needs a way to get the list of the 
available transports, both sco and a2dp.
I have done this by making a shared library of bluez-alsa, improving a 
little the existing pull request with
what was missing (basically a header file and pkgconfig file). I noticed 
that you do not want to merge
such a thing in the short term, so I am using my own branch for that for 
the moment.

But to open the bluez-alsa socket with bluealsa_open, I need the 
interface name.
In the earlier versions of the app, I had hardcoded 'hci0', but this was 
obviously not acceptable.

With the changes in my pull request, I can add a dbus match rule for 
NameOwnerChanged in the app,
and therefore get the list of interfaces to attach to or detach from. I 
made it work today and it
works fine.

I agree that this is a kind of hybrid solution, because one part of the 
information (the available
hcis) comes from dbus, and the other, from the library (that does not 
officially exists yet).
And still, in this model, the client app gets the hci name, to pass it 
back to the library,
and this is useless, in my use case. In fact, I do not see a use case 
where it could be ...

I do not mind using more dbus signals or methods to accomplish the same 
thing. It it not
a big deal, from the app point of view. It represents a little more work 
in bluez-alsa I guess.

I am not sure I fully understand what you mean about being notified from 
the kernel
about create PCMs, I suppose you are talking of something bigger than 
the bluez audio scope.

Anyway, a full dbus interface to bluez-alsa sounds cool, at least my 
commit goes this way.

Cheers
Thierry

tbultel:
Another topic, is that at startup, bluez-alsa does not get the list of 
available
a2dp streams from bluez, when there are already connected phones.

I have just started to read the code and it could be
related to bluez itself, still not sure.

Just performing a "connect" (on a already connected phone) in bluetoothctl
trigs the notification event from bluez to bluez-alsa, I wonder if
an 'on-demand' API exists.

Arkq:
What do you mean, that bluealsa does not get the list of available A2DP
streams? When there is no sink/source provider, bluez does not autoconnect
these profiles, because it is impossible. When bluealsa register itself in
bluez, there is no autoconnect procedure either - connection should
originate from the user action. So, even though HFP is connected, bluealsa
will not automatically connect other possible profiles. Hence, there is no
A2DP PCM listed in the amixer bluealsa.

Regarding my previous comment. What I was thinking of about kernel
notification about added PCMs is that ALSA core is a kernel module, so when
new audio card is added to the system (USB sound card), the list of
available PCMs is exposed by kernel. However, when new bluealsa PCM is
added (new BT device), such an event has to be read from the bluealsa ctl
socket. So, right now we've got a hybrid solution. In order to unify it
(the same thing which PulseAudio does, but it does it in a monolithic way)
there might be a place for another standalone component which will fill in
the gap between ALSA and other applications via D-Bus bridge. So,
afterwards instead of listening for bluealsa (and standard ALSA) events,
one could listen for D-Bus signals. You will receive PCM name which might
be used with snd_pcm_open(). In order to bypass bluealsa-pcm you will also
receive auxiliary data e.g. hci which might be used for direct bluealsa
integration.

tbultel:
On 01/22/2019 10:23 PM, Arkadiusz Bokowy wrote:
>
> OK, I see.
>
>     In the desktop/laptop world, the current active project is
>     PipeWire. I have met these guys already, and I have no doubt that
>     most linux distributions will have it in the coming years, even if
>     there is still a long way to go.
>
> Nice project. I've never heard of it. Getting rid of libasound might 
> be a good idea, though. The code is indeed one bit entangled mess.
>
I agree, but at the end, one must not forget that it just -works-.

By the way, I omitted to tell you something quite important (how did I 
forget that ...), for bluetooth audio, the PipeWire guys are thinking
about taking back the code from the pulseaudio plugins. This is not 
really nice code either.
I told them about bluez-alsa, they were not aware that it was still living.

>     As a conclusion, I agree that having a d-bus interface to
>     bluez-alsa is a good thing, but I am not convinced that delegating
>     it the dbus support to a standalone service is the best idea.
>
> Hmmm... Maaaybe, bluealsa should provide D-Bus interface instead of 
> dedicated library. I thought about such an approach before, however, I 
> was not sure, whether requiring D-Bus library (for clients) is a good 
> idea. So, I've chosen dedicated thin client library (and at that time 
> I was not aware about file descriptors passing via unix sockets). 
> Using D-Bus will offload connection and error handling, so maybe it's 
> the way to go. What do you think about using D-Bus instead of 
> clt-proto.c ?
>
+1 for this.
Is this something you could implement ?

Arkq:
> I told them about bluez-alsa, they were not aware that it was still
living.

This project is not a successor of old bluez-alsa (other bluez-alsa seems
to be dead). The same project name is a mare coincidence :)

> +1 for this.
> Is this something you could implement ?

Yes, I think, I can implement this. My only concern is performance,
however, D-Bus will be used only for signalling, so maybe the downside is
negligible. Bluez and oFono are using it anyway...

PR #186, opened 2019-02-16
Improvements for applications that use poll() with snd_pcm_sw_params_set_start_threshold

borine:
Alsa applications do not need to call snd_pcm_start(), many instead call snd_pcm_sw_params_set_start_threshold () to request that alsa-lib should automatically start the pcm when sufficient data is in the ring buffer.

Such applications risk being blocked forever in a call to poll() with the current bluealsa code. This PR attempts to generate poll() events at those points where such a block may occur.

Note this PR is a superset of #184, duplicating that change and adding some additional event triggers.

### DIFF
--- a/src/asound/bluealsa-pcm.c
+++ b/src/asound/bluealsa-pcm.c
@@ -182,7 +182,6 @@ static void *io_thread(void *arg) {
 
 		}
 		else {
-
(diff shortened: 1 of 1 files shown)


Arkq:
Thanks, applied.

PR #188, opened 2019-03-03
Make --a2dp-volume also work for sinks

loblik:
Make possible to disable software volume control inside bluealsa
also for sink profiles.

### DIFF
--- a/src/io.c
+++ b/src/io.c
@@ -339,7 +339,9 @@ void *io_thread_a2dp_sink_sbc(void *arg) {
 			rtp_payload_len -= len;
 
 			const size_t samples = decoded / sizeof(int16_t);
-			io_thread_scale_pcm(t, pcm.data, samples, channels);
+			if (!config.a2dp.volume)
+				io_thread_scale_pcm(t, pcm.data, samples, channels);
+
 			if (io_thread_write_pcm(&t->a2dp.pcm, pcm.data, samples) == -1)
 				error("FIFO write error: %s", strerror(errno));
 
@@ -760,7 +762,8 @@ void *io_thread_a2dp_sink_aac(void *arg) {
 			error("Couldn't get AAC stream info");
 		else {
 			const size_t samples = aacinf->frameSize * aacinf->numChannels;
-			io_thread_scale_pcm(t, pcm.data, samples, channels);
+			if (!config.a2dp.volume)
+				io_thread_scale_pcm(t, pcm.data, samples, channels);
 			if (io_thread_write_pcm(&t->a2dp.pcm, pcm.data, samples) == -1)
 				error("FIFO write error: %s", strerror(errno));
 			ffb_rewind(&latm);


loblik:
Maybe this option could have some better name. Something like --no-softvol? I was really confused what it does until I opened the code.

If I've read the code right it just disables the volume scaling which is set either by bluez (avrcp) or alsa control.

Arkq:
> Maybe this option could have some better name. Something like --no-softvol? I was really confused what it does until I opened the code.

I know... this project lacks documentation...

But regarding the issue covered by the `--a2dp-volume` option. The problem is, that the AVRCP does not work reliably (at least it wasn't when I was developing this feature, which was like 2 year ago), so, this feature is optional. The problem with sink is that there is no alternative (as far as I know). So, bluealsa can provide only software volume scaling.

You might see this fork: https://github.com/joerg-krause/bluez-alsa/tree/volume2

Of course, I'm open to propositions, even conceptual ones. How volume scaling should work in general? When bluealsa is working as a source it might be desirable to send audio without scaling and let the hardware (e.g. BT speaker) do the scaling - this case is pretty simple (that's why there is `--a2dp-volume` option). But what if bluealsa is working as a sink? Bluealsa does not play the audio, it only exposes it as a PCM so it can be forwarded to the speaker (e.g. via blualsa-aplay). Then the scaling can be done by the hardware. But there are two actors in here: application which connects to the bluealsa PCM, and the BT device which connects to the host with bluealsa installed.

And there are two questions:
1) Who should do the scaling from the BT device point of view?
2) Who should do the scaling from the application connected to the PCM point of view?

From the BT device point of view the case is simple. It is the same situation as with bluealsa source case. The scaling should be done by the "sink hardware". In the case of bluealsa + bluealsa-aplay, scaling should be done by the hardware PCM which is used by the bluealsa-aplay (joerg-krause's fork).

From the application point of view, the scaling should be done by the "hardware" which exposes "capture PCM". Which in our case is.... connected BT device.

So IMHO, the sink scaling strongly depends on the use case. And I'm not sure how this issue should be addressed, so it just hags in the abyss :)

loblik:
> From the application point of view, the scaling should be done by the "hardware" which exposes "capture PCM". Which in our case is.... connected BT device.
I'm not sure which sink usecase is this. Like if you would like to pass the samples to some application instead of feeding it to the sound card?
I don't know much about alsa but maybe there could be a capture device which level could be adjusted using avrcp commands from device and also from application using alsalib?

Yes, joerg-krause's branch is doing exactly what I need. It even adds an option to directly control alsa volume control from bluealsa using avrcp commands sent from BT device.

Any chance this would get merged? I think this very useful for sinks with hardware volume control. In these cases you want to do the volume scaling at the last stage of whole chain (in the soundcard). For the best audio performance.

Arkq:
> I'm not sure which sink usecase is this. Like if you would like to pass the samples to some application instead of feeding it to the sound card?

It is a "capture" use case - recording sound from the microphone (in our case remote BT device). And in the hardware sound card, there is a (hardware) gauge for microphone volume.

> Any chance this would get merged? I think this very useful for sinks with hardware volume control. In these cases you want to do the volume scaling at the last stage of whole chain (in the soundcard). For the best audio performance.

Right now, I think, no. Maybe when I'll finish D-Bus interface for BlueALSA, this feature might be less "entangled" so I'll try to merge it :)

PR #189, opened 2019-03-03
Add socket activation support (systemd) to bluealsa

loblik:
This adds basic socket activation support to bluealsa. It allows bluealsa to be started on demand by bluealsa-aplay in systemd environment.

I did this to solve the racing problem between bluealsa and bluealsa-aplay during boot (socket not present yet when bluealsa-aplay is started).

### DIFF
--- a/src/ctl.c
+++ b/src/ctl.c
@@ -673,23 +673,28 @@ struct ba_ctl *bluealsa_ctl_init(const char *hci) {
 		goto fail;
 	}
 
-	if ((pfd.fd = socket(PF_UNIX, SOCK_SEQPACKET, 0)) == -1) {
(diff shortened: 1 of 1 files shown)


Arkq:
Thanks for this PS, however, right now I'm in the middle of a BIG refactoring of BlueALSA internals, which finally should result in providing D-Bus interface instead of UNIX socket :)

The concept is simple:
Start bluealsa service during system boot and forget about it.

Arkq:
Ctl interface is now gone. Now BlueALSA uses D-Bus which should separate bluealsa server and bluealsa-aplay. You can use D-Bus signals for starting bluealsa-aplay.

PR #192, opened 2019-03-11
make bluealsa_poll_revents() behave like hardware devices

borine:
Modify bluealsa_poll_revents() so that behaviour is closer to that of the hw plugin. After poll() has been called by the application while the pcm is not in a running state, set the POLLERR flag and reload the event trigger in the cases where the hw plugin does.


Arkq:
Thx again! Applied.

PR #218, opened 2019-06-02
utils/aplay.c: use correct type for pfd count

borine:
original code assumed sizeof(nfds_t) == sizeof(size_t); that is not true on all architectures. This commit corrects that and so removes the compiler warning on such architectures.


Arkq:
Thx :)

PR #231, opened 2019-07-08
Improve bluealsa_delay heuristics

jc-kynesim:
The current method can be quite uneven when reporting buffer fullness.
VLC is confused by this and keeps dropping in and out of avsync
recovery. This patch provides a more even response which VLC likes
better.


Arkq:
Many thanks for this PR, I will try to review and test it in the next week or two.

spl237:
Is there any chance this could get merged into the released code at some point? It's particularly important for Raspberry Pi, as VLC used with Bluetooth is basically broken without it.

Many thanks.

PR #237, opened 2019-08-09
Adding dependency packages via apt-get install in Readme

mdasari823:


Arkq:
Thx for this PR, but I'm not sure whether the wording is right. There are many Linux distributions, and not all of them is using `apt` for managing packages. If I type these commands on my distro, I will get `command not found` :). For me more familiar names are `zypper` and `emerge`. So, the question is whether such specific commands should be displayed just next to the dependency list?

I'm not against putting such info on the readme (from time to time, there is an issue related to "project is not compiling because of missing dependency"), but maybe such info shall be below the dependency list with the hint that these commands are for Ubuntu based distributions?

infirit:
This is what wikis are for https://github.com/Arkq/bluez-alsa/wiki which you can link to in the readme.

You can also add things like ``apt-get install build-essential`` :wink: .

Arkq:
@mdasari823 I've just created the stub for ["installing dependencies"](https://github.com/Arkq/bluez-alsa/wiki/Installing-dependencies). Could you extend it (correct) for your distribution? I'm assuming that you are using Ubuntu. Then, we can add link to the readme as @infirit suggested (thx for that) :).

mdasari823:
@Arkq Thanks for checking this out. The problem is that people are lazy enough to just follow step-by-step instructions on the front page of the repo. So, I was thinking giving these hints for at least one (Ubuntu distribution) even before the configure and make instructions in the Readme is useful. I think  depending on the distribution, they figure out what works for them instead of using apt-get, for e.g., brew on Mac.

Arkq:
> The problem is that people are lazy enough to just follow step-by-step instructions on the front page of the repo.

The problem is, that if someone lazy is trying to setup bluez-alsa there are more obstacles than dependency installation. So there is a great chance that he/she will not cut it through even though dependencies are right. Alternatively, there should be a lot more info on the front page, however, the maintenance of such page might be an issue. I agree with @infirit that the place of an installation manual is a wiki page.

I've updated the installation guide: https://github.com/Arkq/bluez-alsa/wiki/Installation-from-source Fell free to update it if something is not clear enough.

PR #239, opened 2019-08-20
src/ba-transport.c: use correct mtu for SCO links

borine:
For SCO audio links, we *must* send data in chunks of the link mtu size to ensure correct timing. To use a constant value of 48 or 24 bytes is just wrong.

Obtain the link config *after* successful connection, because before then the mtu has not been negotiated with the remote device

### DIFF
--- a/src/ba-transport.c
+++ b/src/ba-transport.c
@@ -815,26 +815,19 @@ static int transport_acquire_bt_sco(struct ba_transport *t) {
 	if (t->bt_fd != -1)
 		return t->bt_fd;
 
-	if (hci_devinfo(t->d->a->hci_dev_id, &di) == -1) {
-		error("Couldn't get HCI device info: %s", strerror(errno));
+	if ((t->bt_fd = hci_open_sco(t->d->a->hci_dev_id, &t->d->addr, t->type.codec != HFP_CODEC_CVSD)) == -1) {
+		error("Couldn't open SCO link: %s", strerror(errno));
 		return -1;
 	}
 
-	if ((t->bt_fd = hci_open_sco(di.dev_id, &t->d->addr, t->type.codec != HFP_CODEC_CVSD)) == -1) {
-		error("Couldn't open SCO link: %s", strerror(errno));
+	if (hci_devinfo(t->d->a->hci_dev_id, &di) == -1) {
+		error("Couldn't get HCI device info: %s", strerror(errno));
+		transport_release_bt_sco(t);
 		return -1;
 	}
 
 	t->mtu_read = di.sco_mtu;
 	t->mtu_write = di.sco_mtu;
-
-	/* XXX: It seems, that the MTU values returned by the HCI interface
-	 *      are incorrect (or our interpretation of them is incorrect). */
-	t->mtu_read = 48;
-	t->mtu_write = 48;
-
-	if (t->type.codec == HFP_CODEC_MSBC)
-		t->mtu_read = t->mtu_write = 24;
 
 	debug("New SCO link: %d (MTU: R:%zu W:%zu)", t->bt_fd, t->mtu_read, t->mtu_write);
 


Arkq:
WOW, so the correct solution for the MTU issue was in the plain sight
all the time. @borine many thx for this PR! I will test it and merge
as soon as possible :)

Arkq:
@borine please look at this:
https://github.com/pulseaudio/pulseaudio/commit/82e7fe413e86e9e3098d9676649e1c66e5996899
What do you think?

Unfortunately, I've got limited possibilities for testing; however,
I'll try to use all my Bluetooth devices which support HFP and see if
something will pop up. Also, I don't know whether such issue is BT
dongle specific or remote BT device specific or Kernel specific...

borine:
Yes I've seen that link; not sure what to make of it. I have no idea where the fixed MTU of 48 bytes for SCO came from either.
There are a number of differences in the pulseaudio code from bluealsa, and it may be some other issue that is breaking HSP for them - for example, pulse never sets the BT_VOICE property on the socket.
The only way to know if this commit improves HSP/HFP performance is to test with a selection of headsets and bt chips, some which previously worked well, and some that previously failed.
My suspicion is that the critical factor is the host controller - specifically that the driver or the firmware defaults vary between models, and neither bluealsa nor pulseaudio explicitly sets every possible property/option.
For example, I note that bluealsa does not explicity set BT_VOICE unless the connection is "transparent". Can we really be certain that the default is always BT_VOICE_CVSD_16BIT, or should we force that value for CVSD connections? I find the lack of documentation and source comments in the Bluez project to be extremely frustrating - even worse than the alsa project, which I had not thought possible!

Arkq:
I've tested it with my laptop (samsung np530) and unfortunately it does not work. The `hci_devinfo()` reports MTU to be 64 bytes for both read and write. With such settings, it seems that the BT chip hangs and kernel needs to reload driver (see `dmesg` output below). :(

```
[192261.948819] usb 2-1.5: USB disconnect, device number 6
[192261.950329] Bluetooth: hci0 sending frame failed (-19)
[192261.965325] Bluetooth: hci0 setting interface failed (19)
[192262.211936] usb 2-1.5: new full-speed USB device number 7 using ehci-pci
[192262.366581] usb 2-1.5: New USB device found, idVendor=8087, idProduct=07da
[192262.366583] usb 2-1.5: New USB device strings: Mfr=0, Product=0, SerialNumber=0
```

PS.
This happens when bluez-alsa is configured without mSBC support (you need updated master branch because there was a bug which has prevented proper HFP connection). When bluez-alsa is configured with mSBC and BT headset supports mSBC codec, the transparent voice is used which does not result in crash. However, there is no sound.... mSBC supports is not working correctly (at least for me).

PS2.
Build in USB BT dongle:
```
Bus 002 Device 007: ID 8087:07da Intel Corp. 
Device Descriptor:
  bLength                18
  bDescriptorType         1
  bcdUSB               2.00
  bDeviceClass          224 Wireless
  bDeviceSubClass         1 Radio Frequency
  bDeviceProtocol         1 Bluetooth
  bMaxPacketSize0        64
  idVendor           0x8087 Intel Corp.
  idProduct          0x07da 
  bcdDevice           78.69
  iManufacturer           0 
  iProduct                0 
  iSerial                 0 
  bNumConfigurations      1
```

borine:
I need to do a lot more reading, the controller lockup is very surprising to me so there must be something fundamental I have not understood. Please leave this alone for now, I will get back if I find anything, but that is unlikely to be for a few days (or even weeks) now.

Arkq:
I've made some changes in the code in the area of HCI info. Right now it will be fetched during adapter creation (part of some "coming soon" change). I've pushed it right now, so it might bring some light in the area of MTU.
I've checked the values obtained by the `hci_devinfo()` during the bluealsa startup and after the SCO connection acquisition. And guess what... for me, these values are the same....

You can use this simple dump function:
```
debug("\n Dev: %s (%s), flags: %#x, type: %#x"
     "\n Features: %x %x %x %x %x %x %x %x"
     "\n PKT type: %#x, link policy: %#x, link mode: %#x"
     "\n ACL MTU: %d (pkts: %d), SCO MTU: %d (pkts: %d)",
     a->hci.name, batostr_(&a->hci.bdaddr), a->hci.flags, a->hci.type,
     a->hci.features[0],a->hci.features[1],a->hci.features[2],a->hci.features[3],
     a->hci.features[4],a->hci.features[5],a->hci.features[6],a->hci.features[7],
     a->hci.pkt_type, a->hci.link_policy, a->hci.link_mode,
     a->hci.acl_mtu, a->hci.acl_pkts, a->hci.sco_mtu, a->hci.sco_pkts);
```

PS.
There was one minor conflict between current master and this PR. I've resolved the conflict and created merge commit (new feature of GitHub?) and it seems thet this commit was committed in your branch... from which this PR was created o_O What a great way of trolling :D

borine:
I realized that my first commit got the mtu from the wrong socket - HCI instead of SCO; not sure if that alters the value returned, but needs fixing anyway. Unfortunately my local repo is out of sync with github after your merge so I can't upload any more commits atm. I'll look into sorting that out soon.
Meantime, I'm trying to make sense of the MTU values I'm seeing in the context of the packet format specifications and not having much joy at present. Will continue with this in the background over the next couple of weeks.

Arkq:
> Unfortunately my local repo is out of sync with github after your merge so I can't upload any more commits atm.

Try something like this :) :
```
git remote add arkq https://github.com/Arkq/bluez-alsa.git
git pull --rebase arkq master
git push --force
```

borine:
After a lot of research, I've realised that the original premise of this PR was based on a misunderstanding of how the SCO MTU is agreed between BT adapters. So, I'm closing it here. 

Having better understood the issues around HSP/HFP with Bluez, I have some ideas on what bluealsa can do to improve its support. I will open an issue ticket soon as a place to discuss them, and if we can agree on an approach there I will then raise a new PR.

PR #246, opened 2019-09-14
Start stop issues

borine:
Looking into problems with sox play and mpg123, I've found an issue caused by the way these applications close and then immediately re-open a bluealsa device, and also with mpg123 in particular that an application can cause the client to call bluealsa_stop() twice.

This PR attempts to fix these issues. See comments against each of the two commits for further details.

### DIFF
--- a/src/asound/bluealsa-pcm.c
+++ b/src/asound/bluealsa-pcm.c
@@ -279,6 +279,15 @@ static int bluealsa_stop(snd_pcm_ioplug_t *io) {
 	struct bluealsa_pcm *pcm = io->private_data;
 	debug2("Stopping");
 
+       /* It is possible for this function to be called from an invalid state,
+        * so we need to make it a no-op in those cases. */
+       switch (io->state) {
+               case SND_PCM_STATE_OPEN:
+               case SND_PCM_STATE_SETUP:
+               case SND_PCM_STATE_DISCONNECTED:
+                       return 0;
+       }
+
 	if (pcm->io_started) {
 		pcm->io_started = false;
 		pthread_cancel(pcm->io_thread);
--- a/src/ba-transport.c
+++ b/src/ba-transport.c
@@ -685,11 +685,18 @@ static int transport_acquire_bt_a2dp(struct ba_transport *t) {
 	GUnixFDList *fd_list;
 	GError *err = NULL;
 
-	/* Check whether transport is already acquired - keep-alive mode. */
+	/* Check whether transport is already acquired - keep-alive mode.
+	 * We must ensure that transport release is not in progress before
+	 * testing the BT fd. Otherwise we might have the IO thread close it
+	 * after we have decided to reuse it! */
+	ba_transport_pthread_cleanup_lock(t);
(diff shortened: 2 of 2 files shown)


src/ba-transport.c
```diff
@@ -685,11 +685,18 @@ static int transport_acquire_bt_a2dp(struct ba_transport *t) {
 	GUnixFDList *fd_list;
 	GError *err = NULL;
 
-	/* Check whether transport is already acquired - keep-alive mode. */
+	/* Check whether transport is already acquired - keep-alive mode.
+	 * We must ensure that transport release is not in progress before
+	 * testing the BT fd. Otherwise we might have the IO thread close it
+	 * after we have decided to reuse it! */
+	ba_transport_pthread_cleanup_lock(t);
```
Arkq: Thanks for nailing down the cause of race condition during PCM open. However, I think, that the lock should be done in the [bluealsa_pcm_open()](https://github.com/Arkq/bluez-alsa/blob/4c32fc30068a4d6faca462d8e7fbc64e3dec4c0b/src/bluealsa-dbus.c#L253), just before PCM fd is checked.
The D-Bus messages coming to the bluez-alsa are handled by a single thread, so there cannot be any race. So if I understand correctly, the PCM FIFO has been closes which caused IO thread to terminate itself. Before termination, the pcm.fd is checked (the lock in the IO thread), so the lock should be acquired also just before pcm.fd is checked in the D-Bus handler. Otherwise, during pcm open call someone might get busy error code which in fact is not true, because PCM is about to be freed.

src/asound/bluealsa-pcm.c
```diff
@@ -279,6 +279,15 @@ static int bluealsa_stop(snd_pcm_ioplug_t *io) {
 	struct bluealsa_pcm *pcm = io->private_data;
 	debug2("Stopping");
 
+       /* It is possible for this function to be called from an invalid state,
```
Arkq: I'm not sure whether it resolves problem or just covers it :). If state is not STATE_RUNNING (or similar), then IO thread should not be stared as well, so the `pthread_cancel()` will not be called (the `if` below). So, the only function which causes trouble is `bluealsa_dbus_pcm_ctrl_send_drop()`, am I correct? If so, then the problem should be fixed inside the bluez-alsa code. The "grand design" of bluez-alsa is that bluealsa server handles everything correctly and the ALSA plugin is just a wrapper for ALSA API compatibility :D.

Arkq:
@borine I've made the modification that I've mention in the commend above and pushed amended commit to master. I hope you don't mind. Many thanks for spotting this bug (and resolving issues with mpg123)!

I've pushed only the commit with IO thread synchronization, though. I think that the other one is not necessary. If I'm wrong, please provide some example which fails without this "no-op stop" commit.

borine:
Thanks, I can confirm your commit fixes the issues that this PR is intended to address. Sox play now works correctly with all my test files on both Intel and raspberry pi.
mpg123 is not yet fully fixed, but that's a different story and I'll raise an new issue to discuss that.
Closing this PR as complete.

PR #259, opened 2019-10-15
Allow to set ALSA PCM format

joerg-krause:
Currently the audio output is limited to signed 16-bit (S16_LE). This PR allows to use different PCM audio format in the future. This is necessary if a library decodes samples into a different audio format, e.g. S24_3LE (libopenaptx does it).

### DIFF
--- a/src/ba-transport.c
+++ b/src/ba-transport.c
@@ -546,6 +546,10 @@ unsigned int ba_transport_get_sampling(const struct ba_transport *t) {
 	return 0;
 }
 
+uint16_t ba_transport_get_format(const struct ba_transport *t) {
+	return g_dbus_transport_get_pcm_format(t->type);
+}
+
 uint16_t ba_transport_get_delay(const struct ba_transport *t) {
 	if (t->type.profile & BA_TRANSPORT_PROFILE_MASK_A2DP)
 		return t->delay + t->a2dp.delay;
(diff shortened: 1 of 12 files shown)


Arkq:
Hi :) It looks promising! I will pull this PR in parts, though. In a day or two I will release version 2.0.0, so before this release I will add D-Bus format property without internal changes. Later I will merge other commits.

One small thing. I'm not sure whether format should be based on the ALSA enumeration values (SND_PCM_FORMAT_S16_LE). Right now bluealsa server does not rely on ALSA library (and other API-derived values) so maybe it should stay that way. What do you think?

joerg-krause:
Thanks for merging already some parts of the PR!

I considered about not relying on the ALSA enumerations as well. But then I thought, bluealsa is an ALSA-only (it cannot be used with any other sound library) program, so why not rely on it?

Arkq:
It's not entirely true, that bluealsa server is ALSA only. Bluez-alsa
project is more or less ALSA only due to PCM plug-in. But server can be
used everywhere. See other forks :), e.g. integration with pipewire.

joerg-krause:
I see! So, I am okay when the audio format is mapped between the server and bluealsa-aplay.

Arkq:
I'm closing this PR sice all commits except "Allow to pass the length of the PCM buffer length" have been merged into master. And the not merged commit needs improvement anyway (e.g. patching the code for reading from PCM pipe).

PR #277, opened 2019-12-16
fix Makefile.am to include bluetooth headers

tassig:
The current Makefile.am was not including bluetooth headers path in AM_CFLAGS


Arkq:
Thx, patch applied :)

PR #290, opened 2020-02-02
Fix 'Invalid argument' when setting buffer_time

graham8:
When running bluealsa-aplay on a AML 905X box I get
`bluealsa-aplay: W: Couldn't open PCM: Set HW params: Set buffer time: Invalid argument: 500000
` 
Setting period_time first (like alsa's aplay does) fixes it.


Arkq:
Thx :)

PR #297, opened 2020-02-17
configure.ac: default to /etc for sysconfdir

borine:
without options, bluealsa configure by default places the alsa config file in /usr/etc/alsa/conf.d for alsa-lib >= 1.1.7, and so libasound does not load it into its configuration.

This PR uses a hard-coded /etc/alsa/conf.d as the default, so that configure should "just work" without options on a default alsa 1.1.7+ install. I've tried to maintain backwards compatibility with the --sysconfdir option

Alsa hard-codes this path here:
https://github.com/alsa-project/alsa-lib/blob/master/src/conf/alsa.conf
there is no substitution by autoconf variables so for the great majority of installations this PR should give the correct default outcome.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -171,6 +171,14 @@ AC_ARG_WITH([alsaplugindir],
 	AS_HELP_STRING([--with-alsaplugindir=dir], [path where ALSA plugin files are stored]),
 	[alsaplugindir="$withval"],
 	[alsaplugindir=$($PKG_CONFIG --variable=libdir alsa)/alsa-lib])
+# unfortunately, for ALSA >= 1.1.7 the directory for add-on configuration files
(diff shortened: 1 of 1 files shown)


borine:
apologies - this is not working for alsalib <= 1.1.6 - I'll look at it again later, please leave it for now

borine:
hmm, works for me with either alsa 1.1.3 (ubuntu 18.04 lts) or alsa 1.1.9 (raspbian buster). not sure what caused the test failure. Will come back to it.

Arkq:
@borine, do not worry about travis failed build. This pcm test fails a lot because it is not deterministic :p ALSA io plug code adds a lot of randomness to this test. I'll merge that when I'll be at home. Thanks!

PR #317, opened 2020-04-10
do not ignore sco acquisition failure

### LINKED ISSUE #315: sco acquistion failure not handled
Sometimes, when a client opens a sco pcm for playback on an AG instance, transport acquisition can fail. This failure is ignored by the sco io thread, so the client becomes blocked writing to a full pipe that the server never reads from. At the very least, the client pcm should be closed when no transport can be acquired (the a2dp code does this).

I can trigger this by first starting an a2dp playback client, then starting a sco playback client to the same device (using `bluealsa -p a2dp-source -p hfp-ag`). Once this has happened with my device, all future sco attempts fail, even after the a2dp transport has been released. Disconnecting the device then reconnecting restores normal function.

borine:
firmware or driver bug (for example) may cause sco transport acquisition failure. If we ignore this, then the client is left blocked indefinitely because the daemon never reads or writes to the pcm pipe.

This PR proposes to close the pcm pipe(s) so that an alsa client will get an ENODEV error.

Fixes #315

### DIFF
--- a/src/sco.c
+++ b/src/sco.c
@@ -325,7 +325,14 @@ void *sco_thread(struct ba_transport *t) {
 				 * For a headset mode we will wait for an incoming connection from
 				 * some remote Audio Gateway. */
 				if (t->type.profile & BA_TRANSPORT_PROFILE_MASK_AG)
-					t->acquire(t);
+					if (t->acquire(t) == -1) {
+						/* without an acquired transport, it is impossible
+						 * to send or receive audio from the bt device.
+						 * So we close the client pcm to prevent clients
+						 * becoming blocked indefinitely */
+						ba_transport_release_pcm(&t->sco.mic_pcm);
+						ba_transport_release_pcm(&t->sco.spk_pcm);
+					}
 				/* fall-through */
 			case BA_TRANSPORT_SIGNAL_PCM_RESUME:
 				asrs.frames = 0;


src/sco.c
```diff
@@ -325,7 +325,14 @@ void *sco_thread(struct ba_transport *t) {
 				 * For a headset mode we will wait for an incoming connection from
 				 * some remote Audio Gateway. */
 				if (t->type.profile & BA_TRANSPORT_PROFILE_MASK_AG)
-					t->acquire(t);
+					if (t->acquire(t) == -1) {
```
Arkq, line 328: What will happen if you try to reacquire transport instead of releasing PCMs? E.g.:
```c
if (t->acquire(t) == -1 && some_max_counter++ < 10)
  ba_transport_send_signal(t, BA_TRANSPORT_SIGNAL_PCM_OPEN);
```

borine: I'm unable to reproduce the underlying error since I rebooted, so I an unable to test that. When the connect() call at hci.c:101 failed, errno was ECONNABORTED, which is the result of HCI error 0x16 "Connection Terminated by Local Host". At that point, hcitool showed that the SCO link was lost but the ACL link was still alive. I was unable to re-establish a SCO llnk until the device was completely disconnected and re-connected using bluetoothctl.
So in the particular case I saw, repeating the acquision attempt would, I think, just fail again. But I am unable to actually test that.

My thinking with this PR was just that in the rare event that acquire() should fail, we should not just leave the client blocked. Perhaps a retry would help in some cases, but just not in the specific case that I saw. So perhaps retry is a better, more general solution. Would a timed delay ever be needed rather than an immediate retry? Without any information about the cause of the error or the state of the host and controller stacks its impossible to tell.

Arkq: I'm not saying that the solution for this issue should be "retry". I think that returning error and disconnecting PCMs is the right decision. However, I was wondering what might have caused connection abort in the first place...

On my host I often see that first SCO acquisition is OK, but next attempt sometimes produces a LOT of error messages in syslog, e.g.:
```
kernel: Bluetooth: hci0: SCO packet for unknown connection handle 35
```

borine: > I was wondering what might have caused connection abort in the first place...

me too :)

There is nothing unusual in my logs. I will continue to try different scenarios to reproduce the failure

Arkq:
After closer look, I've simplified SCO acquisition. Now, A2DP and SCO are acquired in the same way. Also, I've found one peculiar bug which has caused PCM Open DoS :D

Issue #125, opened 2018-06-27
Feasibility to implement a command line parameter to disable volume control.


badbat75:
When bluealsa-aplay is started volume of reproduction device go to 100%.
There's no way to control it by remote device and volume control can't be disabled.

badbat75:
It is possible to implement a parameter in blue-alsa command line to disable volume control when used as a sink device?

Arkq:
What do you mean by "disable volume control"? Could you specify exact usecase?

Also, I'm not able to test it properly, because my smarphone does not communicate well with bluez and volume control works fine on the smartphone site (AVRCP volume command is not sent to PC). Anyway, AVRCP communication is beyond the scope of bluez-alsa.

Take a look at this fork: https://github.com/hmbedded/bluez-alsa

badbat75:
I'm sorry, I understand I was not so clear.
The problem is that in the BT sink host, independently of how the master volume is set, when blualsa-aplay is started, the master volume of the default alsa device is set to 100%. So if I want to hear at soft volume a bluetooth streaming I have to reduce the bluetooth volume on the source device but the quality of reproduction is very bad because in every device I'm using the blootooth volume is a software volume.
I supposed the behaviour in bluealsa is because the host can manage the volume parameter sent from client devices, but I really don't know.
Is it a bug that volume go to 100% when bluealsa-aplay start, or is there a reason for this behaviour?
If it is the right behaviour, is there a way to disable it keeping the volume at its own value without changing it when bluealsa-aplay is started?

joerg-krause:
I have implemented a volume control based on @hmbedded 's fork: https://github.com/joerg-krause/bluez-alsa/tree/volume.

The volume is set at the beginning of a transport session to the current volume of the Mixer element passed via the `--volume-control` option, e.g. "Master":

```
bluealsa -p a2dp-sink --a2dp-volume &
bluealsa-aplay --profile-a2dp --volume-a2dp --volume-control Master 00:00:00:00:00:00
```
With volume control, if the transport volume is changed, e.g. on the iPhone, the ALSA mixers control volume is changed. If the volume is changed on the device, e.g. by hardware buttons or via amixer, the transport value is updated, too.

badbat75:
Hi Joerg,
I tested the volume branch with parameter you suggested but:

In this way:
`
/usr/bin/bluealsa -p a2dp-source -p a2dp-sink
[...]

PR #321, opened 2020-04-14
Add ALSA mixer volume control support to aplay

joerg-krause:
Add a new command line option `-M` or `--mixer=NAME` to allow using an ALSA mixer for volume control. For using the native volume control the bluealsa server needs to be started with the `--a2dp-volume` command line option, e.g.:

```
$ bluealsa -p a2dp-sink --a2dp-volume
$ bluealsa-aplay -D default -M Master --profile-a2dp XX:XX:XX:XX:XX:XX
```

Setting the volume works in both ways.

Note, that if the bluealsa server is started with `--a2dp-volume` and no mixer control name is passed to bluealsa-aplay, the volume will not change when the source changes the volume.

Closes: #125

### DIFF
--- a/utils/aplay/aplay.c
+++ b/utils/aplay/aplay.c
@@ -14,6 +14,7 @@
 
 #include <errno.h>
 #include <getopt.h>
+#include <math.h>
 #include <poll.h>
 #include <pthread.h>
 #include <signal.h>
@@ -34,6 +35,7 @@
 #include "shared/log.h"
 #include "alsa-pcm.h"
 #include "dbus.h"
+#include "volume_mapping.h"
 
 struct pcm_worker {
 	pthread_t thread;
@@ -45,8 +47,14 @@ struct pcm_worker {
 	int ba_pcm_ctrl_fd;
 	/* opened playback PCM device */
 	snd_pcm_t *pcm;
+	/* mixer pointer */
+	snd_mixer_elem_t *melem;
 	/* if true, playback is active */
 	bool active;
+	/* if true, an ALSA mixer event occured */
+	bool mixer_event;
+	/* if true, a transport volume event occured */
+	bool transport_volume_event;
 	/* human-readable BT address */
 	char addr[18];
 };
@@ -73,6 +81,9 @@ static pthread_rwlock_t workers_lock = PTHREAD_RWLOCK_INITIALIZER;
 static struct pcm_worker *workers = NULL;
 static size_t workers_count = 0;
 static size_t workers_size = 0;
+
+static snd_mixer_t *mixer = NULL;
+static const char *mixer_device = NULL;
 
 static bool main_loop_on = true;
 static void main_loop_stop(int sig) {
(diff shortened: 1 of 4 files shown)


borine:
I think this implementation repeats an error that exists in the MPD alsa mixer event handling. It assumes that *all* alsa mixer events are caused by a volume change. That is not true. An event is also generated when a mixer control is unplugged (eg if aplay output is to a bluetooth or USB device that is disconnected). 

The SND_CTL_EVENT_MASK_REMOVE event must not be ignored, as it indicates that the snd_mixer_elem_t* is no longer valid, and any attempt to dereference it will result in a segfault (if you are lucky) or undefined behaviour (if not so lucky).

For this code to be safe for use with removable alsa mixer controls, it will have to implement appropriate behavior when a the chosen control is removed. See the alsamixer source for how that handles mixer control events: https://github.com/alsa-project/alsa-utils/blob/10e5cef85a7e32828a73fea88edf7c6c5d4259bc/alsamixer/mixer_widget.c#L61

utils/aplay/aplay.c
```diff
@@ -326,20 +463,43 @@ static void *pcm_worker_routine(struct pcm_worker *w) {
 	size_t pause_counter = 0;
 	size_t pause_bytes = 0;
 
-	struct pollfd pfds[] = {{ w->ba_pcm_fd, POLLIN, 0 }};
+	int npfds = 0;
+	struct pollfd *pfds;
 	int timeout = -1;
 
+	if (mixer_device)
+		w->melem = init_mixer_control(&mixer, pcm_device, mixer_device);
```
borine, line 471: No, you cannot pass a PCM device id to init_mixer_control(). "default" is a special case, but try "hw:0,0" or "plughw:0,0" they give "Invalid CTL" error. You have to use a CTL device id, eg "hw:0" (assuming card 0 has a mixer), "bluealsa", etc.

In general, to achieve volume control with alsa you need 2 parameters, a control name (sometimes also referred to as a "mixer" name) and a control element name (eg "Master", etc). So to allow the user to select an arbitrary volume control, bluealsa-aplay would need 3 alsa params: PCM name, CTL name, and element name.

Arkq:
@joerg-krause some time ago I've started revitalization of Howard Mitchell (original code for mixer support) work as well. You can find it in [aplay](https://github.com/Arkq/bluez-alsa/commit/51cc153eca48b64605a06a38cfe8a7f769c18a0a) branch - look at last commit only. However, I've faced some design dilemma... I will describe it in more details later - this feature is not so simple to develop... :)

Arkq:
There are two things with mixer support.

Firstly, I'm not sure whether volume_mapping should be there. It doesn't seem right.... but maybe there is no other way of doing it. Similar problem is with amixer bluealsa ctl plugin. The volume control via `alsamixer -D bluealsa` is not linear, because alsa mixer maps linear gauge input into "real" update values. Bluetooth devices (I think) expose volume as a "linear" (from the human ear perception point of view) gauge, though. So, after applying alsamixer normalization, the effect is terrible - only top 50% of the volume gauge is usable. I think that as a first draft I would go for this feature without volume_mapping.

Secondly (which is more important), mixer support should take into account multiple devices playing simultaneously. There have to be some rules, though. Who should control the volume? Maybe any device (?), then the volume change caused by one device should emit volume update action in other devices, and vice versa. Maybe only the first device (that was my original idea) should have access to mixer, but other devices should be notified about volume change... Maybe first device should be master (it controls mixer) but other devices can control volume in respect to the master device... And I don't know which answer is the right one :D

Also, mixer can not be opened (and attached) before PCM. The PCM device might be not available during the bluealsa-aplay start (even during the pcm_worker_routine() startup), e.g. one will use USB sound card which is not connected yet, or some other bluealsa device (which is also not connected yet). Hence, mixer shall be opened after output PCM device for consistency.

borine:
> I'm not sure whether volume_mapping should be there

I think it should, otherwise the source device will not display the same value for volume as alsamixer (although I am not able to test this at the moment) - see next paragraph

> Who should control the volume?

It seems to me that this PR does not use a Master/Slave pattern, rather it uses an Implementation/Interface pattern. The alsa CTL provides the volume control implementation, the source device(s) provide interface(s) to the volume control. So all source devices display the same volume level, as obtained from the CTL, and their volume switches all send change requests to the CTL which applies them in the order recieved. We also want the source devices to display the same value as displayed by alsamixer on the host, and that is why the alsamixer volume_mapping algorithm must be used. I disagree with the way this code sets the alsa ctl volume when a source device connects - it should be the other way round, ie the initial source device volume should be set to match the ctl volume. That way all source devices can be treated equally.

> mixer can not be opened (and attached) before PCM

The mixer and specific control element within it may be available without the PCM, and vice-versa; it depends on the actual implementation. Therefore we should not assume that it is available. We should attempt to open and attach only when a device is connected *and* we have a valid PCM handle. We should release the mixer resources when the PCM is closed (which happens when the last source device disconnects, or when the output device is disconnected).

utils/aplay/aplay.c
```diff
@@ -116,6 +127,132 @@ static snd_pcm_format_t bluealsa_get_snd_pcm_format(const struct ba_pcm *pcm) {
...
+	if ((err = snd_mixer_load(*handle)) < 0) {
+		error("Mixer %s load error: %s", device, snd_strerror(err));
+		snd_mixer_close(*handle);
+		goto fail;
+	}
+	for (elem = snd_mixer_first_elem(*handle); elem; elem = snd_mixer_elem_next(elem)) {
+		if (!strcmp(snd_mixer_selem_get_name(elem), name))
+			break;
+	}
+	snd_mixer_elem_set_callback(elem, melem_event);
```
borine, line 248: elem could be NULL here if no element called _name_ exists. so snd_mixer_elem_set_callback() needs to be guarded.

utils/aplay/aplay.c
joerg-krause: Thanks for the review! I will update the code to use two parameters for the mixer control element. Or three with an additional index.

utils/aplay/aplay.c
joerg-krause: Thanks! I will improve the mixer handling.

joerg-krause:
> Firstly, I'm not sure whether volume_mapping should be there. It doesn't seem right.... but maybe there is no other way of doing it.

You need at least some kind of mapping, as the ALSA device might have different min and max values, e.g `min_dB: -6, max_dB: 20`.

> The volume control via `alsamixer -D bluealsa` is not linear...

I am using the ALSA softvol PCM plugin, which is also not linear. Hardware DACs will also have a non-linear volume curve.

> Secondly (which is more important), mixer support should take into account multiple devices playing simultaneously.

Forgive my ignorance, but how does that work (how does multiple devices play simultaneously)? I mean, there is only one ALSA audio output device. And I must admit, that I haven't tried to connect more than one Bluetooth device.

Arkq:
> You need at least some kind of mapping, as the ALSA device might have different min and max values, e.g min_dB: -6, max_dB: 20.

Yeah... that why I think that the mapping problem is a separate feature/issue (definitely separate commit). For now I do not care that volume gauges for ALSA and BlueALSA will not match.

> how does multiple devices play simultaneously

If you select ALSA PCM which supports mixing (i.e. hardware-based or dmix-based), you can open PCM for writing multiple number of times. For every new BT connection, bluealsa-aplay starts new worker thread which tries to play to the selected ALSA PCM. Additionally, there is a `--single-audio` option, which tries to stop BT playback for all except first BT deceive by sending Stop command to Player interface. Sometimes it works sometimes not, but playback is blocked in `bluealsa-aply` anyway, however, remote BT device might not be aware of that in case where Player interface does not work.

joerg-krause:
> If you select ALSA PCM which supports mixing (i.e. hardware-based or dmix-based), you can open PCM for writing multiple number of times. For every new BT connection, bluealsa-aplay starts new worker thread which tries to play to the selected ALSA PCM. [...]

I see. As the ALSA volume control is mapped to the transport volume, a change of the ALSA volume changes the transport volume and therefore should be propagated to all Bluetooth devices being connected to the PCM device (which uses the ALSA mixer control interface).

Many thanks for the feedback so far. I will revise the PR, but i cannot yet say exactly when.

hifiberry:
Any progress here?

Arkq:
I've updated (rebased on current master) PR #390. Also, I've revised my old work with @borine comments regarding master/slave vs implementation/interface and yeap, that seems to be true that BT device should be synced with ALSA playback mixer, not the other way around. So in PR #390, upon playback start, the volume indicator of BT device will be synced with mixer. Currently, in #390 there is no code for forwarding ALSA mixer volume back to BT device - it can be added later though. I'm still testing it, but I think that it will be ready for merge with master in few days.

hifiberry:
@Arkq That's great news!

Arkq:
@hifiberry  if you are using such feature it would be great if you could test PR #390, whether it suits you. Please note, that BlueALSA PCM can not use software volume, so either start bluealsa with `--a2dp-volume` or use `bluealsa-cli soft-volume <pcm> n` to disable software volume.

hifiberry:
I'll have a look. Right now I'm using a mix of your old patch with a newer aplay which doesn't work perfectly well ;-)
Might still take some time as I'm quite busy today.

hifiberry:
@Arkq Just trying to compile it, but it looks like ba-transport.c always imports bsd/sys/time.h that is not available on my system.
It probably should be handled the same way as it is handled in rt.h.
I see if I can add a hack to get it compiled.

Arkq:
Simply remove this include, or merge this PR branch with current master.

hifiberry:
Just did it.

hifiberry:
Ok, initial tests are looking good :-)

Arkq:
This functionality has been merged with 7050c9a (PR #390)

flatsiedatsie:
Whoa, aplay has volume control now? That's fantastic! I'm currently literally generating 100 different wav-files for each volume level, and playing the correct one based on the user's selected volume. Can't wait to see this arrive on the Raspberry Pi.

PR #327, opened 2020-04-25
improve alsa pcm poll event generation

borine:
for playback, implement support for avail_min sw parameter, and generate internal events during the time synchronization sleep, so that applications can continue to write while the io thread sleeps.

this change enables using bluealsa with alsaloop, for example, and possibly other applications that attempt to synchronize 2 pcms with managed latency.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -1,9 +1,20 @@
 # BlueALSA - Makefile.am
-# Copyright (c) 2016-2018 Arkadiusz Bokowy
+# Copyright (c) 2016-2020 Arkadiusz Bokowy
 
 ACLOCAL_AMFLAGS = -I m4
 SUBDIRS = src utils
 
 if ENABLE_TEST
 SUBDIRS += test
+endif
+
+if WITH_COVERAGE
+cov:
+	$(MAKE) $(AM_MAKEFLAGS) check CFLAGS="$(CFLAGS) -O0 --coverage"
+	$(LCOV) --capture -d src -d test --exclude '/usr/*' --exclude "*/test/*" -o cov.info
+	$(GENHTML) -o coverage -t $(PACKAGE) cov.info
+clean-local:
+	find $(top_builddir) -name "*.gcno" -delete
+	find $(top_builddir) -name "*.gcda" -delete
+	rm -rf coverage cov.info
 endif
--- a/configure.ac
+++ b/configure.ac
@@ -44,6 +44,19 @@ AM_CONDITIONAL([ENABLE_DEBUG_TIME], [test "x$enable_debug_time" = "xyes"])
 AM_COND_IF([ENABLE_DEBUG_TIME], [
 	AC_DEFINE([DEBUG_TIME], [1], [Define to 1 if the debug timing is enabled.])
 ])
+
+# embedded test coverage
+AC_ARG_WITH([coverage],
+	AS_HELP_STRING([--with-coverage], [use lcov for test coverage reporting]))
+AM_CONDITIONAL([WITH_COVERAGE], [test "x$with_coverage" = "xyes"])
+AM_COND_IF([WITH_COVERAGE], [
+	AC_PATH_PROG(LCOV, lcov)
+	AC_PATH_PROG(GENHTML, genhtml)
+	AC_SUBST([LCOV])
+	AC_SUBST([GENHTML])
+])
+
+# in-place call-stack unwinding
 AC_ARG_WITH([libunwind],
 	AS_HELP_STRING([--with-libunwind], [use libunwind for call-stack unwinding]))
 AM_CONDITIONAL([WITH_LIBUNWIND], [test "x$with_libunwind" = "xyes"])
--- a/src/asound/bluealsa-pcm.c
+++ b/src/asound/bluealsa-pcm.c
@@ -55,7 +55,7 @@ struct bluealsa_pcm {
 	int event_fd;
 
 	/* virtual hardware - ring buffer */
-	snd_pcm_uframes_t io_ptr;
+	snd_pcm_sframes_t io_ptr;
 	pthread_t io_thread;
 	bool io_started;
 
@@ -134,6 +134,26 @@ static void io_thread_cleanup(struct bluealsa_pcm *pcm) {
 }
 
 /**
+ * Some playback applications may start the PCM before they have written the
+ * first full period of audio data. To match the behavior of the ALSA hw
+ * plug-in we do not report an underrun in this case. Instead we pause the IO
+ * thread for the estimated time it should take for a real-time application to
+ * write the balance of the period. */
+static void io_thread_wait_first_period(snd_pcm_ioplug_t *io) {
+	struct bluealsa_pcm *pcm = io->private_data;
+	while (io->appl_ptr < pcm->io_avail_min && (
+				io->state == SND_PCM_STATE_RUNNING ||
+				io->state == SND_PCM_STATE_PREPARED)) {
+		uint64_t nsec = (pcm->io_avail_min - io->appl_ptr) * 1000000000 / io->rate;
+		struct timespec ts = {
+			.tv_sec = nsec / 1000000000,
+			.tv_nsec = nsec % 1000000000,
+		};
+		nanosleep(&ts, NULL);
+	}
+}
+
+/**
  * IO thread, which facilitates ring buffer. */
 static void *io_thread(snd_pcm_ioplug_t *io) {
 
@@ -156,7 +176,10 @@ static void *io_thread(snd_pcm_ioplug_t *io) {
 	}
 
 	/* Block I/O until a full period of frames is available. */
-	bool started = false;
+	if (io->stream == SND_PCM_STREAM_PLAYBACK) {
+		debug2("Waiting for first period of frames");
+		io_thread_wait_first_period(io);
+	}
 
 	struct asrsync asrs;
 	asrsync_init(&asrs, io->rate);
@@ -164,42 +187,27 @@ static void *io_thread(snd_pcm_ioplug_t *io) {
 	debug2("Starting IO loop: %d", pcm->ba_pcm_fd);
 	for (;;) {
 
-		int tmp;
-		switch (io->state) {
-		case SND_PCM_STATE_RUNNING:
-			/* Some playback applications may start the PCM before they have
-			 * written the first full period of audio data. To match the
-			 * behavior of the ALSA hw plug-in we do not report an underrun in
-			 * this case. Instead we pause the IO thread for the estimated time
-			 * it should take for a real-time application to write the balance
-			 * of the period. */
-			if (!started && io->stream == SND_PCM_STREAM_PLAYBACK) {
-				if (pcm->io_avail_min > io->appl_ptr) {
-					struct timespec ts = {
-						.tv_nsec = (pcm->io_avail_min - io->appl_ptr) * 1000000000 / io->rate };
-					debug2("IO thread started with insufficient frames - pausing for %ld ms", ts.tv_nsec / 1000000);
-					sigtimedwait(&sigset, NULL, &ts);
-					asrsync_init(&asrs, io->rate);
-					continue;
-				}
-				started = true;
-			}
-			break;
-		case SND_PCM_STATE_DRAINING:
-			break;
-		case SND_PCM_STATE_DISCONNECTED:
-			goto fail;
-		case SND_PCM_STATE_XRUN:
-			started = false;
-			/* fall-through */
-		default:
+		if (io->state == SND_PCM_STATE_PAUSED ||
+				pcm->io_ptr == -1) {
+			int tmp;
 			debug2("IO thread paused: %d", io->state);
 			sigwait(&sigset, &tmp);
+			if (pcm->io_ptr == -1)
+				continue;
+			if (pcm->ba_pcm_fd == -1)
+				goto fail;
+			if (io->stream == SND_PCM_STREAM_PLAYBACK) {
+				debug2("Waiting for first period of frames");
+				io_thread_wait_first_period(io);
+			}
 			asrsync_init(&asrs, io->rate);
 			debug2("IO thread resumed: %d", io->state);
 		}
 
-		snd_pcm_uframes_t io_ptr = pcm->io_ptr;
+		if (io->state == SND_PCM_STATE_DISCONNECTED)
+			goto fail;
+
+		snd_pcm_sframes_t io_ptr = pcm->io_ptr;
 		snd_pcm_uframes_t io_buffer_size = io->buffer_size;
 		snd_pcm_uframes_t io_hw_ptr = pcm->io_hw_ptr;
 		snd_pcm_uframes_t io_hw_boundary = pcm->io_hw_boundary;
@@ -221,7 +229,7 @@ static void *io_thread(snd_pcm_ioplug_t *io) {
 		len = frames * pcm->frame_size;
 
 		io_ptr += frames;
-		if (io_ptr >= io_buffer_size)
+		if (io_ptr >= (snd_pcm_sframes_t)io_buffer_size)
 			io_ptr -= io_buffer_size;
 
 		io_hw_ptr += frames;
@@ -251,7 +259,6 @@ static void *io_thread(snd_pcm_ioplug_t *io) {
 
 			/* check for under-run and act accordingly */
 			if (io_hw_ptr > io->appl_ptr) {
-				io->state = SND_PCM_STATE_XRUN;
 				io_ptr = -1;
 				goto sync;
 			}
@@ -309,7 +316,6 @@ static int bluealsa_start(snd_pcm_ioplug_t *io) {
 	 * we might end up with a bunch of IO threads reading or writing to the
 	 * same FIFO simultaneously. Instead, just send resume signal. */
 	if (pcm->io_started) {
-		io->state = SND_PCM_STATE_RUNNING;
 		pthread_kill(pcm->io_thread, SIGIO);
 		return 0;
 	}
@@ -322,19 +328,11 @@ static int bluealsa_start(snd_pcm_ioplug_t *io) {
 		return -errno;
 	}
 
-	/* State has to be updated before the IO thread is created - if the state
-	 * does not indicate "running", the IO thread will be suspended until the
-	 * "resume" signal is delivered. This requirement is only (?) theoretical,
-	 * anyhow forewarned is forearmed. */
-	snd_pcm_state_t prev_state = io->state;
-	io->state = SND_PCM_STATE_RUNNING;
-
 	pcm->io_started = true;
 	if ((errno = pthread_create(&pcm->io_thread, NULL,
 					PTHREAD_ROUTINE(io_thread), io)) != 0) {
 		debug2("Couldn't create IO thread: %s", strerror(errno));
 		pcm->io_started = false;
-		io->state = prev_state;
 		return -errno;
 	}
 
@@ -487,7 +485,6 @@ static int bluealsa_pause(snd_pcm_ioplug_t *io, int enable) {
 		return -errno;
 
 	if (enable == 0) {
-		io->state = SND_PCM_STATE_RUNNING;
 		pthread_kill(pcm->io_thread, SIGIO);
 	}
 
--- a/src/asound/bluealsa-pcm.c
+++ b/src/asound/bluealsa-pcm.c
@@ -356,6 +356,10 @@ static int bluealsa_stop(snd_pcm_ioplug_t *io) {
 	}
 	pcm->delay_valid = false;
 
+	/* Bug in ioplug - if pcm->io_hw_ptr == -1 then it reports state
+	 * SND_PCM_STATE_XRUN instead of SND_PCM_STATE_SETUP after pcm is stopped */
+	pcm->io_hw_ptr = 0; 
+
 	if (!bluealsa_dbus_pcm_ctrl_send_drop(pcm->ba_pcm_ctrl_fd, NULL))
 		return -errno;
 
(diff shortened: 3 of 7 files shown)


borine:
I've now found a few issues with this, I'll push updated code later in the week. Also I can't get capture to work at all with alsaloop,.

Arkq:
OK, no problem. Right now I've got not much time for checking github activity, though. But I will review this code in a week or two :)

Regarding the PCM IO plug-in. Some time ago (maybe year or two) I though about restoring old bluealsa-pcm approach (the one used by all other plugins) - using `transfer()` callback. This (I hope) would simplify things because IO thread could be removed. Unfortunately, I do not remember why I switched to hardware imitation with IO thread :/ I think that the reason was the problem with synchronization caused by named PIPE opening. Right now PIPE file descriptor is passed via socket so there is no sync issue. I know that with `transfer()` approach capture PCM was very simple and has worked more or less. The problem was with playback (mostly). However, I'm afraid that a lot of knowledge, which is already in the bluealsa-pcm, will be lost, so old bugs may reemerge. But maybe (for now) only port capture PCM type.... @borine what do you think? Do you have any experience with other IO plugins (e.g. from PulseAudio)?

borine:
update on this PR

Investigating the interaction between bluealsa and alsaloop has found 2 issues.

First, which the original push here was intended to address, is that to achieve low latency while synchronizing 2 pcm devices, alsaloop relies on the use of the avail_min sw parameter to control the frequency of poll() ready events. Currently bluealsa ignores the avail_min setting which makes it impossible for alsaloop to achieve low latency. After initial promising results, it has now become clear that attempting to invoke poll events in the application thread from the i/o thread cannot work. So please do not waste any time reading the code currently on this branch. I now have a different approach based on using a timertfd to trigger poll() in the application, which I have working for playback, but not capture.

Second, alsaloop uses a drop/prepare/start sequence to reset the capture pcm when it falls too far behind the playback pcm. Bluealsa is far too slow to perform this sequence with a latency much less than 500 ms. My target latency for this work is 20ms. To achieve this will require some major surgery to the i/o thread, but I believe it is possible and am now starting on an implementation. It will take some time to complete this.

So, I can :
- close this PR and raise a new one when/if I get a working solution, or
- leave this PR and eventually push the new code over it when ready, or
- tidy up the playback fix as it now stands as work-in-progress and push that, possibly next week, so others can see it.

@Arkq Unless you have some other preference, I'll choose the second option and just leave this here. Also, a warning that the capture fix involves modifying the i/o loop to use pol() to monitor both a timerfd to signal period timings and the FIFO fd for i/o readiness. This is a big change that will necessitate a lot of regression testing. So I will understand if you decide it is too risky and will not be offended if you choose to reject it in the end.

Arkq:
> leave this PR and eventually push the new code over it when ready

Yes, keep this work here.

> Also, a warning that the capture fix involves modifying the i/o loop to use pol() to monitor both a timerfd to signal period timings and the FIFO fd for i/o readiness. This is a big change that will necessitate a lot of regression testing.

Adding new file descriptor to the `poll()` shall not have bit impact, I think. I'm OK with that.

borine:
> Adding new file descriptor to the poll() shall not have bit impact, I think. I'm OK with that.

Actually the timerfd replaces the eventfd, so the number of fds monitored remains the same. Results for the playback fix (timerfd in application thread only, io thread timing remains the same) are good using  quad core intel laptop, but I want to confirm the design with a single core Rpi Zero W before publishing the new code. I hope to do that sometime in the coming week.

borine:
I've force pushed a completely new attempt. Having experimented with, and abandoned, the use of a timerfd, I've been forced to accept that it is not possible, nor necessary, to force accurate period timing with an ioplug plugin. This revision goes back to the eventfd approach, but with a simplified io thread loop and improved revents function that respects the avail_min sw param setting.

I have tested on an intel laptop and rpi zero W, using aplay, arecord, mpg123, mpd (both playback and capture) and also vlc on the intel laptop only. I have also tested all the above through alsaloop with very good results using a latency setting of 50 ms.

Arkq:
> This revision goes back to the eventfd approach, but with a simplified io thread loop and improved revents function that respects the avail_min sw param setting.

Great, I will apply this PR on top of my local work and will test it as well. At first glance it looks OK. I've got one question, though. Could you please create separate commit for `SND_LIB_VERSION <= 0x010103`? It is rather a separate change and can go into the master right away :)

borine:
> Could you please create separate commit for SND_LIB_VERSION <= 0x010103?

Will do.

> At first glance it looks OK

Great. This obviously needs a ***lot*** of regression testing, but hopefully it is now close to being fit for purpose. In the hope that this will eventually get merged, I've started drafting a wiki page on using dmix with bluealsa. It still needs a lot of work, but is visible here:
https://github.com/borine/bluez-alsa/wiki/Using-bluealsa-with-dmix

Since the last push, I've noticed a couple more issues that need to be addressed:

- compiling with `make CFLAGS=-Wall` shows up a few typos that do not affect the function of the plugin but I would like to fix. I'll do them when I separate out the SND_LIB_VERSION commit.

- pause function is clearly not right. In practice, the IO thread gets blocked in the write() call because the FIFO is full. But the design says it should block in the sigwait() call. I haven't noticed any adverse effects of this for applications, but my ability to test pause is limited. I will continue to work on this as time allows.

borine:
I've broken down this PR into a set of distinct commits, the first is the SND_LIB_VERSION change.
Fix for pause function is still outstanding.

Arkq:
Thank you very much @borine for breaking this PR into separate commits. It has helped me a lot during reviewing changes.

I've made a preliminary review with minor modifications (on my own, mostly codestyle, variable naming, and few readability improvements, and fix-commit squashes), and I've pushed it to branch [borine-alsa-pcm](https://github.com/Arkq/bluez-alsa/tree/borine-alsa-pcm). Please, check on your side whether I did not break anything (the "alsa pcm plugin: fix pause funcion" is not yet there).

However, there are two things left before merging into master. I've tested this code in various scenarios, and it seems that the behaviour is almost the same (in my usage cases I did no see any difference), but I have noticed that sometimes VLC is "triggered" to restart PCM. I'm not using VLC on regular basis so maybe commit https://github.com/Arkq/bluez-alsa/commit/51bd19bddfe6daea79003b0591565fc984387a6b also triggers VLC from time to time.... Also PCM test does not pass...

I've modified a little changes in the `test-alsa-pcm` test in order to allow running it with user defined PCM device. You can run it as follows: `test-alsa-pcm --pcm=hw:0 playback`. On my hardware (e.g. hw:0, hdmi) test-alsa-pcm passes, but with this PR test with bluealsa plug-in fails due to two things: delay is not reported unless PCM is in the running state (where my hardware reports delay), and `snd_pcm_state(pcm)` does not report XRUN after long delay in writing samples. Maybe there are more issues, but I did not check as thoroughly. So the question is, whether this test shall be changed (it was written by me based on empirical tests on my hardware), or maybe fixes in this PR should be addressed in a slightly different way... @borine what do you think?

The usage of test-alsa-pcm:
- `test-alsa-pcm` - run all tests with bluealsa (you have to be in the `build/test` directory)
- `test-alsa-pcm playback` - run only playback test
- `test-alsa-pcm --pcm=hw:0 playback` - run playback test on custom PCM
- `test-alsa-pcm --pcm=hw:0 termination-ref` - run playback termination reference test on custom PCM

borine:
Thank you for taking time with this. I will review your branch as soon as possible.

Getting delay reporting right is, I think, essential. So I need to read up some more to understand what is meant by "delay" when a pcm is not running, and to make sure that by trying to satisfy alsaloop I have not broken other, more important, applications.

Leave it with me.

Arkq:
> Getting delay reporting right is, I think, essential. So I need to read up some more to understand what is meant by "delay" when a pcm is not running, and to make sure that by trying to satisfy alsaloop I have not broken other, more important, applications.

I think that this PR does not affect any "real" application, but maybe I'm wrong. Anyway, it's a pitty that alsa-lib does not provide any test cases for PCM devices (since it is possible to write software PCM). It could be even a good tool to unify behavior of hw, plug, ioplug internals, etc. So, maybe the first step to "fix" this issue once and for all, would be to prepare reliable PCM test application? (or maybe there is such app already?) Then it might be used as a guide to fix alsa-lib. I'm not sure whether all issues related to ioplug has been fixed within alsa-lib or not, because after no response for ~6 months from alsa-lib devs after sending patches related to "device disconnection based on USB sound card" (not plain bug reports) for ioplug I simply gave up on them :D And then we could "fix" bluealsa ioplug plug-in :)

I think that test-alsa-pcm might be a good stub for such test app, but is should be more like unit-test. So the question is, which behavior we shall test. For sure, writing, pausing, underrun, overrun, PCM prepare, and so on. But what about more sophisticated use cases which are found in various well-known applications? The good case is VLC which relies on "smooth" delay reporting, and any rapid change to delay value triggers PCM restart. And for sure there are more "clever" implementations :D

borine:
Here is a patch against your borine-alsa-pcm branch for the XRUN test failure. Unfortunately it requires the io thread to directly modify the ioplug state, which was something I had intended to avoid with this PR. But I can see no way to force ioplug to read the hw pointer when the application calls `snd_pcm_state()`
```
diff --git a/src/asound/bluealsa-pcm.c b/src/asound/bluealsa-pcm.c
index 225bf45..25255ee 100644
--- a/src/asound/bluealsa-pcm.c
+++ b/src/asound/bluealsa-pcm.c
@@ -226,10 +226,12 @@ static void *io_thread(snd_pcm_ioplug_t *io) {
                        frames = hw_avail;
 
                /* There are 2 reasons why the number of available frames may be zero:
-                * xrun or drained final samples - ioplug will determine which, we just
-                * need to set the HW pointer to -1. */
+                * xrun or drained final samples; we set the HW pointer to -1 to
+                * indicate we have no work to do. */
                if (frames == 0) {
                        io_hw_ptr = -1;
+                       if (io->state != SND_PCM_STATE_DRAINING)
+                               io->state = SND_PCM_STATE_XRUN;
                        goto sync;
                }
```

I will look into VLC issues later.

I agree that an application designed to test all possible pcm states and state changes would be extremely valuable, but I don't know of any, and I think it would not be a trivial undertaking to create one :smile:

borine:
and here is a patch for branch borine-alsa-pcm that fixes VLC synchronization and also appears to work with alsaloop.
```
diff --git a/src/asound/bluealsa-pcm.c b/src/asound/bluealsa-pcm.c
index 225bf45..adb43ca 100644
--- a/src/asound/bluealsa-pcm.c
+++ b/src/asound/bluealsa-pcm.c
@@ -521,10 +521,9 @@ static int bluealsa_delay(snd_pcm_ioplug_t *io, snd_pcm_sframes_t *delayp) {
 		return -EPIPE;
 	}
 
-	/* It is not clearly defined what the delay of a PCM that is not running
-	 * should be, and alsaloop fails with the method below. So in this release
-	 * we simply report zero. */
-	if (io->state != SND_PCM_STATE_RUNNING) {
+	/* For capture PCMs only, the delay must be reported as zero when an overrun
+	 * occurs. */
+	if (io->stream == SND_PCM_STREAM_CAPTURE && io->state == SND_PCM_STATE_XRUN) {
 		*delayp = 0;
 		return 0;
 	}
@@ -539,7 +538,7 @@ static int bluealsa_delay(snd_pcm_ioplug_t *io, snd_pcm_sframes_t *delayp) {
 
 	if (!pcm->delay_valid) {
 		/* If we have never transmitted anything, the delay is just based
-		 * on the current buffer length (transfered frames or free frames
+		 * on the current buffer length (transfered frames or unread frames
 		 * respectively for playback or capture). */
 		delay = snd_pcm_ioplug_hw_avail(io, io->hw_ptr, io->appl_ptr);
 		if (io->stream == SND_PCM_STREAM_CAPTURE)
```

Arkq:
> Here is a patch against your borine-alsa-pcm branch for the XRUN test failure. Unfortunately it requires the io thread to directly modify the ioplug state, which was something I had intended to avoid with this PR. But I can see no way to force ioplug to read the hw pointer when the application calls snd_pcm_state()

Yes, I've seen that your intention was not to mess around with internal ioplug state, which I agree is a bad design. And I thought that maybe it will work that way now - I was struggling with the same design dilemma ~2 years ago, and I thought that maybe someone has already fixed alsa-lib, but it seems that not yet :D 

I will apply your patches to the borine-alsa-pcm branch.

borine:
You may have already spotted this, but in the VLC patch above, I clumsily left in these lines in bluealsa_delay(), which clearly need to be deleted:
```
-       if (io->state == SND_PCM_STATE_XRUN) {
-               *delayp = 0;
-               return -EPIPE;
-       }
```
my apologies

Arkq:
Branch updated, ready for testing :)

borine:
I've noticed that using:
```
test-alsa-pcm --pcm=hw:Loopback,0,0 playback
```
also fails the XRUN test. So perhaps, the test is not really vaild as it stands? If we ever get issues from modifying pcm state without locking, then I suggest we can revert my "fix" for this test and change the test program to call snd_pcm_avail() or some such before snd_pcm_state().

Arkq:
I'm currently rewriting pcm test so it will be more modular (it will be easier to add more tests). Also, I'm adding new test scenarios (with verbose comments) and not all of them pass with current master bluealsa ioplug plug-in. I will check later with your changes (borine-alsa-pcm branch), but now I'm verifying these test cases with `hw:0`. I will push changes maybe today evening so you could check whether my test-cases make any sense. Then, it will be more like test-driven development :)

Arkq:
@borine I've rebased the `borine-alsa-pcm` branch on top of [redesigned PCM test](https://github.com/Arkq/bluez-alsa/blob/a8c413dd0662c439d2ee8ff6db60cf80391df62c/test/test-alsa-pcm.c). Unfortunately, test does not pass even with your changes, but it's better than with current master - capture test passes :) Please, see whether new "unit" tests make any sense. Since you've got loopback device configured, please see whether pcm test passes with it.

I've checked new test framework with my `hw:0` like follows (for over 2 minutes):
```sh
while ./test-alsa-pcm -D hw:0 capture playback; do true; done
```

> On further reading of the alsa api docs, it seems that an application needs to call snd_pcm_status() to guarantee that the user space view of the hardware state is updated. I think snd_pcm_state() is intended for use when it is known that the application state is still fresh, eg immediately after snd_pcm_readi(), etc.

I've replaced all `snd_pcm_state()` calls with `snd_pcm_status()` with a wrapper function (it will be easy to revert it to `snd_pcm_state()` and see whether it indeed causes issues).

borine:
From test/test-alsa-pcm.c:385
```
	/* check if buffer fullness is at the next-to-last period */
	ck_assert_int_eq(snd_pcm_avail(pcm), period_size);
```
That code does not allow for the fact that alsa-lib does not necessarily allocate the buffer as an exact multiple of periods. In fact AFAICT the ioplug rarely does. In the test log:
```
D: ../../../src/asound/bluealsa-pcm.c:423: /org/bluealsa/hci0/dev_12_34_56_78_9A_BC/a2dpsrc/sink: Selected HW buffer: 8 periods x 4408 bytes <= 35280 bytes
```
8 periods x 4408 bytes = 35264 bytes = 8816 frames
but buffer size 35280 bytes = 8820 frames
```
../../test/test-alsa-pcm.c:386:F:../../test/test-alsa-pcm.c:test_playback_start:0: Assertion 'snd_pcm_avail(pcm) == period_size' failed: snd_pcm_avail(pcm) == 1106, period_size == 1102
```
so we have one period plus 4 frames available in the buffer. That is correct, and expected.

Arkq:
> That code does not allow for the fact that alsa-lib does not necessarily allocate the buffer as an exact multiple of periods.

Yes, I'm aware of that (that's why in the bluealsa-pcm.c logs tere is an explicit `<=` sign in such case), but to be honest I'm not sure whether it's not a ioplug bug, I see no reason why buffer is not evenly partitioned.... Anyway I will change this check to `ck_assert_int_eq(snd_pcm_avail(pcm), buffer_size - (i - 1) * period_size)` so this check will pass.

borine:
bluealsa-pcm.c:561
```
		unsigned int tsamples =
			(diff.tv_sec * 1000 + diff.tv_nsec / 1000000) * io->rate / 1000;
		/* Do not allow us to have a negative number of samples queued! */
		if ((delay = delay - tsamples) < 0)
			delay = 0;
```
That is the cause of the pause failure. The tsamples calculation needs to be adjusted to account for pcm in paused state.

borine:
> Since you've got loopback device configured, please see whether pcm test passes with it.

I can confirm both playback and capture tests pass with Loopback card

Arkq:
> I can confirm both playback and capture tests pass with Loopback card

That's great to hear. In the meantime I've made some more modifications to the alsa-pcm test (changed playback start test to account for extra frames in the buffer; reverted [XRUN patch](https://github.com/Arkq/bluez-alsa/pull/327#issuecomment-662352469) and `snd_pcm_state()` -> `snd_pcm_status()`) and force pushed to borine-alsa-pcm :).

> That is the cause of the pause failure. The tsamples calculation needs to be adjusted to account for pcm in paused state.

Yeah, it seems so...

borine:
I've just synced my branch improve_alsa_poll with your borine-alsa-pcm, and added a new commit to address the pause delay test failure. This branch now passes all tests.

Arkq:
I've added test coverage reporting to the makefile (for convenience), and I've found out that `bluealsa_drain()` callback was not testes. I've added new simple snd_pcm_drain test (code pushed to borine-alsa-pcm) and voila, bug have emerged :)

After call to `snd_pcm_drain()`, PCM shall be stopped and in the SETUP state (according to alsa-lib source code, I did not found any documentation to backup this statement). Hardware PCM passes test, but bluealsa PCM fails, because we end up in the XRUN. I think that the problem is again with `io_thread()`. I think that `bluealsa_drain()` should also notify IO thread that we are going to drain PCM (and finally stop it), so IO thread could write out all data (in case of playback) and stop. @borine what do you think? (after so much research within alsa-lib I regard you as an expert in ALSA :))

borine:
hmm - yes this is a bug :disappointed:

It seems ioplug expects bluealsa_drain() to block until the drain is complete - whether or not the pcm is set to non-blocking; I also consider that a bug in ioplug logic which may confuse some applications! It is not documented, but is implicit in the code here:

https://github.com/alsa-project/alsa-lib/blob/a3ca4803cb8db73d01231c69620e2d18573ffba9/src/pcm/pcm_ioplug.c#L526

The code where ioplug determines drained or xrun is only used when the plugin does not implement its own drain() function - I had not realised that previously. 

This will require some synchronization between iothead and application thread.

borine:
I've done some debugging with extra logging, and found that bluealsa_drain() is being called with `io->state == SND_PCM_STATE_RUNNING`. It turns out this is a bug in ioplug that was fixed in release 1.1.6 https://github.com/alsa-project/alsa-lib/commit/07a17bd5a50289e2fdb2714a4e39f38f41811558

I'll prepare a new commit that implements blocking behavior and forces `io->state == SND_PCM_STATE_PAUSING`  in bluealsa_pause() where necessary.

borine:
last commit addresses drain issues, and passes tests for me with alsa-lib 1.1.3 on ubuntu 18.04LTS. I'll test with alsalib 1.1.8 on raspberry pi later.

borine:
commit 5057484 deals with poll events for nonblocking drain, that just leaves applications that call snd_pcm_status() immediately after snd_pcm_drain() with non-block flag set - with alsa-lib < 1.1.6 they may still see state SND_PCM_STATE_XRUN instead of SND_PCM_STATE_SETUP

Maybe we do need to set io->state = XRUN explicitly in the io thread after all for alsa-lib < 1.1.6. I'll take some time to think that through.

borine:
didn't much time for thinking after all! It turns out that before 1.1.6, ioplug ignored the nonblock flag, and ignored the return value from bluealsa_drain(), and always called snd_pcm_ioplug_drop() after bluealsa_drain(). There was no way for snd_pcm_drain() to return -EAGAIN when using ioplug. So I think bluealsa can also ignore the nonblock flag for those alsa-lib releases. I'll redo the last commit with this idea see how that goes.

borine:
commit f75f040 has drain working for me with alsa-lib 1.1.3 and 1.1.8

Arkq:
> actually, bluelasa_drain() does not always need to block - if io->nonblock is true then we can return -EAGAIN

Hmm... for non-blocking drain operation `bluealsa_dbus_pcm_ctrl_send_drain()` will have to be modified. Currently, this function waits for server response (for "OK" message). So, when `bluealsa_dbus_pcm_ctrl_send_drain()` returns, drain is already completed.

The problem with non-blocking `dbus_pcm_ctrl_send` operation is that then there will be no way to be notified that the drain is complete other than the D-Bus signal. I think that for now non-blocking draining might not be supported.... Or if non-blocking drain means, that we send "drain" signal and forget, then such behavior might be supported. I did not found any info in the ALSA documentation what non-blocking drain means :/

borine:
OK, I was over-thinking the issue. If we are happy to always block until drain() completes, then the fix is much simpler. Revised commit coming up ...

Arkq:
> OK, I was over-thinking the issue.

Maybe also my fault, because I have said abut IO thread and app thread synchronization. That, wast my first thought, and (mea culpa) I did not check my assumption...

I can confirm that with your last commit, drain tests passes :) I think, that we are ready to go with merge. I was testing this code for the past two weeks, and now we've got enhanced test coverage. So, there should be no big issues with this new improved PCM plug-in.

However, before merge I will have to "fix" PCM test a little bit, because sometimes mock-server does not quit (hangs somewhere), and PCM test fails with timeout. As a workaround I've added [SIGQUIT](https://github.com/Arkq/bluez-alsa/blob/borine-alsa-pcm/test/test-alsa-pcm.c#L164) but it will have to be changed to SIGTERM, because otherwise test coverage is not generated :)

Arkq:
@borine please, sync your branch with borine-alsa-pcm branch, so this PR will be automatically (after pushing changes to master) marked as merged by github :)

PR #365, opened 2020-08-26
add man pages for bluealsa and bluealsa-aplay

borine:
just for fun, I created some man pages. They are generated from markdown sources, so if you prefer I can easily post them to the wiki instead.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -4,6 +4,10 @@
 ACLOCAL_AMFLAGS = -I m4
 SUBDIRS = src utils
 
+if ENABLE_MANPAGES
+SUBDIRS += doc
+endif
+
 if ENABLE_TEST
 SUBDIRS += test
 endif
--- a/README.md
+++ b/README.md
@@ -80,6 +80,8 @@ run as a root during system startup. It will register `org.bluealsa` service in 
 bus, which can be used for accessing configured audio devices. In general, BlueALSA acts as a
 proxy between BlueZ and ALSA.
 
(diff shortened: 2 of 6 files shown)


Arkq:
Nice! :)

Please use '--enable-manpages' as a configuration option, since this will be a new feature. 'With' is mostly used when feature is already there, but one wants to use some others dependency/backend.

Also, I think, that you can use 3rd argument of AC_PATH_PROG to call msg error.

Other than that, it's fine. I will check the result of this pandoc conversion today evening :)

graham8:
> I can easily post them to the wiki instead

I hope this information won't be _only_ in manpages.  At OSMC we deliberately don't bundle man to save space.

borine:
> I hope this information won't be only in manpages. At OSMC we deliberately don't bundle man to save space.

Its basically generated by running `bluealsa --help` and `bluealsa-aplay --help`. I then added a bit of text taken from the wiki. As the source is markdown, it could be linked from the README perhaps?

borine:
> Please use '--enable-manpages' 

done.

> Also, I think, that you can use 3rd argument of AC_PATH_PROG to call msg error.

I couldn't get that to work, I haven't found any examples of such use online.

Arkq:
One side note. I think that by convention, the extension of man file is the manual section for which man is written, which in this case should be '.1'.

borine:
> the extension of man file is the manual section for which man is written

Yes, that is taken care of by automake in the `make install` target.

PR #366, opened 2020-08-28
Redact man pages according to man-pages convention

Arkq:


Arkq:
@borine I've updated man pages according to `man-pages(7)` convention and I've added some new bits of information. I know that your redacting skills are much better than mine, so could you please review my wording :) - you can add new commits to this PR. Also, maybe there is something more which should be added to these manuals? I've tried to cover all recent github talks, but for sure there are some more shady things about options/usage, maybe I could try to explain them...

I didn't know that writing man files with markdown might be so easy :D I was always afraid of writing manual because of this nasty groff format....

borine:
I'll look at the wording, possibly tomorrow. 

The rendering on Github is very disappointing. I'm actually wondering whether its such a good idea to link from the README (and wiki) after all. I originally did not use the markdown "definition" syntax because github does not understand it. However, my attempts to approximate it for `pandoc` were not too successful either.

Arkq:
> The rendering on Github is very disappointing. 

Yeap, that's unfortunate... However, since the purpose of these files is to produce man pages, the man page layout is the priority, IMHO. One way to overcome the messy github rendering might be to structure man pages a little bit differently. The main issue is the double indentation of multi paragraph option descriptions (single paragraph description looks OK). So, maybe write only one  paragraph and state that more information is in the section below, which then could be written in a multi-paragraph manner...

borine:
> the man page layout is the priority

I agree. I've looked at using rst instead of markdown, as Github renders that too, but it still suffers the same fatal limitation - no support for definition lists in Github. So there appears to be no source format that is suitable for both `man` and Github. Had I realized that earlier I probably would not have started this in the first place!

borine:
A suggestion for putting these man pages online.

With a recent enough version of pandoc (>=2.4), it is possible to generate github markdown (gfm) from man (troff) sources. So rather than linking the source markdown, perhaps it would be better to create gfm pages and post them to the wiki, then link to those instead. It would mean extra maintenance overhead, and the project would lose some level of control over them. But at least they would be readable !!

Arkq:
> usage massage

Nice catch! :D

Regarding github rendering for bluealsa.man, maybe indeed it would be a good idea to move (at leas some parts) detailed description of e.g. profiles to separate section below options... What do you think?

Arkq:
> It would mean extra maintenance overhead, and the project would lose some level of control over them. But at least they would be readable !!

Hmmm... If nothing else will work, that that might be a solution, but I would prefer not to add extra (unnecessary) maintenance tasks :/

borine:
> move (at leas some parts) detailed description

I prefer to have complete man page available for when no internet access is available. 

I've found its possible to create gfm from directly from standard markdown, and posted a sample generated by

 `pandoc -s -f markdown -t gfm -o example-man-page.md doc/bluealsa-manpage.md`

to https://github.com/borine/bluez-alsa/wiki/example-man-page,

Arkq:
Regarding "move detailed description", I was thinking about something like this: https://github.com/Arkq/bluez-alsa/commit/456876f94a633c9306fd392bd3273260579aed94

borine:
>  I was thinking about something like this: 456876f

... ah, ok I see now. Yes, that is *much* better. I think the man page is also better this way too, so that's a win-win then !

Arkq:
> yes, that is much better. I think the man page is also better this way too

Yes, it seems to be more readable. However, I've got no idea how to "transform" codecs quality lists. Maybe some form of table in the PROFILES section? Or maybe only say which number is high quality and which is low, and what is the default - without any info about bit rates?

borine:
> without any info about bit rates

It would be a shame to lose that from the man page, since it is useful information and the presentation by `man` is fine. And IMHO in general tables do not work very well in man pages. Given that the priority is `man` I would suggest leaving it as it stands with a44a26e - the markdown is at least readable.

borine (approved):

PR #372, opened 2020-09-11
Alsa plugin refinements

borine:
The 2 commits here are actually independent of each other and can be cherry-picked. They are intended to improve the performance and application compatibility of the ALSA pcm plugin.


borine:
After some long-running tests, I've found that 1a85f72 causes instability with alsaloop. So I'm withdrawing this PR and will revisit it after release 3.0.0

PR #374, opened 2020-09-13
32-bit PCM for LDAC; 24-bit PCM for apt-X HD

Arkq:


borine (approved):
I've read through this code and all seems well, and I can confirm it compiles with CFLAGS=-Wall without any warnings or errors. I don't have any LDAC or APTX-HD equipment with which to test though.

PR #375, opened 2020-09-13
Alsa plugin refinements

borine:
These 3 commits are independent of each other and can easily be cherry-picked. They help bring pcm plugin behaviour closer to that of the hw plugin.


Arkq:
Thanks! Commits applied!

I was worried about Bluetooth responsiveness, that's why I've set constraint on the lower number of periods. But maybe you are right, that app should be responsible for that, not plugin. After all, app might use 2 long periods, so there should be no underruns (from the point of view of some remote BT device) :)

PR #378, opened 2020-09-14
Do not restrict minimal length for PCM buffer

Arkq:
This commit is a continuation to 3b474ae.


Arkq:
Number of periods is already limited
https://github.com/Arkq/bluez-alsa/blob/125e322c72c1a2861739af76cb10d0fcbea677b1/src/asound/bluealsa-pcm.c#L780-L782
so I've simply removed buffer size constraint.

borine (approved):

PR #381, opened 2020-09-29
Alsa pcm updates

borine:
This completes the work I've been doing to provide better support for "difficult" applications such as alsaloop that require the bluealsa plugin to behave more like the hw plugin. alsaloop is now able to capture from bluealsa devices reliably, and even to capture from one bluealsa device and playback to another. As a side-effect, VLC appears to have reduced latency.

### DIFF
--- a/src/asound/bluealsa-pcm.c
+++ b/src/asound/bluealsa-pcm.c
@@ -79,10 +79,15 @@ struct bluealsa_pcm {
 	struct timespec delay_ts;
 	snd_pcm_uframes_t delay_hw_ptr;
 	unsigned int delay_pcm_nread;
-	bool delay_valid;
+	/* In capture mode, delay_running indicates that frames are being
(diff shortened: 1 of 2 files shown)


borine:
Testing on another machine has shown a race condition in commit 20f8775 - it is possible for an application to invoke bluealsa_pointer() before the io thread has set up its signal handlers, with the result that the application is killed by SIGIO. My preferred solution is to add a condition variable and have the application thread wait in bluealsa_start() until the io thread signals the cond var that it is ready - the existing delay_mutex can be safely used to protect the cond var.
@Arkq - If you don't want to use a cond var here I can re-submit with 20f8775 removed - I am testing that now but will need several days before I can be sure alsaloop remains stable. If you are happy to use cond var I will re-work with that instead.

Arkq:
Thanks! I'll try to review this PR during this weekend (sorry for not responding earlier, but I've got rather busy week...).

borine:
rebased against current master

PR #388, opened 2020-10-08
Fix PCM pause synchronization and revent for capture

Arkq:


Arkq:
@borine, I've pulled one commit from #381 and one commit from your previous work (which I've intentionally omitted month ago) regarding pause synchronization. The synchronization commit is reworked a little bit by me in order to fix two race conditions: possible deadlock on condition variable when IO thread is terminated prematurely (without triggering condition variable), and possible "double" pause in case where IO thread is already paused when calling `bluealsa_pause()`. Also, while fixing test cases I've found that there was a race condition during `close_transport()` call - fixed by mutex addition.

Could you verify with your setup that the "Fix synchronization for PCM pause functionality" commit works for you?

Also, today I've found out that using capture PCM with audiacity does not work, and I remember that it was working ~year ago. I will have to investigate it further....

borine:
> Also, today I've found out that using capture PCM with audiacity does not work,

I've just tried recording with audacity with PR #381 (commit 3caa9ae) and it appears to be working fine. I'll try with master and with this PR later. This is with audacity v2.2.1 on Ubuntu 18.04 LTS. What symptoms do you see?

Arkq:
Same PC setup (ubuntu 18 and audacity 2.2.1). Audacity tests devices (opens
them), and during bluealsa capture device after opening it hangs :/ I've
added logging to all bluealasa ioplug callbacks to see what is going on,
but apparently nothing is called, maybe it waits on poll() forever :/. I
will check with PR #381 fully applied.

borine:
> during bluealsa capture device after opening it hangs

I needed to use LIBASOUND_THREAD_SAFE=0 - this is because audacity opens playback device first, which enables the alsa-lib thread locking function, so the bluealsa plugin is unable to disable it.

Arkq:
Yes, that was the case... I forgot that ubuntu uses alsa-lib affected by
deadlock... Sorry for the confusion...

borine (approved):
Yes, this works well - VLC in particular restarts from pause much more smoothly. I've also tried cherry-picking commit 6b2743b into PR #381 with excellent results.

PR #390, opened 2020-10-18
Forward BlueALSA PCM volume to user defined mixer

Arkq:
This commit allows to forward PCM volume to ALSA mixer element, but
only if BlueALSA PCM does not use software volume scaling. In case of
hardware PCM for playback, audio scaling might also be performed by a
hardware component which should be more efficient.


Arkq:
Merged in 7050c9a

PR #393, opened 2020-11-02
calculate scaling factor only when volume changes

borine:
The A2DP volume scaling factor is currently recalculated on every io event, event though its value is almost always the same. This PR does the calculation only when the volume is changed. The downside is that memory for 2 doubles is required for each PCM (whether SCO or A2DP).

### DIFF
--- a/src/a2dp-audio.c
+++ b/src/a2dp-audio.c
@@ -107,9 +107,9 @@ static void ba_transport_pcm_scale(
 
 	/* scaling based on the decibel formula pow(10, dB / 20) */
 	if (!pcm->volume[0].muted)
-		ch1_scale = pow(10, (0.01 * pcm->volume[0].level) / 20);
+		ch1_scale = pcm->volume[0].scaling_factor;
 	if (!pcm->volume[1].muted)
-		ch2_scale = pow(10, (0.01 * pcm->volume[1].level) / 20);
+		ch2_scale = pcm->volume[1].scaling_factor;
 
 	switch (pcm->format) {
 	case BA_TRANSPORT_PCM_FORMAT_S16_2LE:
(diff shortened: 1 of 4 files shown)


Arkq:
Thx for this PR :) In general I'm OK with not recalculating all the time the same thing, but maybe also it would be possible to clean things a little bit as well (after all it's an enhancement). I though about some function (ba-transport API) which will set channel value (and this function could also pre-calculate scaling factor). However, I did not came up with a "pleasant" (if that's the right word) API for such task... It would be easier with C++, though.

Firstly, I though about getter/setter for volume and mute, but then there is a channels count thingy.... Sometimes it's one channel sometimes two, so such function would have to be flexible (preferably setting two channels at the same time; I don't like the idea of ALSA API where there is a function which takes the channel index, and you can end up with only one channel muted, context switch, only one channel of PCM scaled, context switch, second channel muted, etc.). But how to make a getter which will return two channels... and returning value via arguments does not look nice... :D In the and of the struggle I've left the code as it is right now (for future code maintenance improvements).

And as a final note, I think, that the code in the `ba_transport_pcm_volume_update` should be called in the setter body. But then there is a `volume = xxx` code which does not call `ba_transport_pcm_volume_update`, e.g. in the ba-rfcomm code... Tough nut.... :D

If you've got some time, I'd be happy to see some propositions. However, currently I've got maybe 1-2 hours a week for open-source contributions, so please, do not expect prompt response from my side. I hope I will have some more spare time during the second half of November.

borine:
It looks like I'll be unable to do any programming for several weeks, possibly months, in the new year, so I want to concentrate on other projects in the time before then. In particular, I have been working on multi-client support for the bluealsa daemon, and want to get that to a stage where i can at least share it so others can experiment with it, and possibly finish it, while I'm away. This volume scaling change was part of that work.

I will look again at volume control and think through your ideas if I get time before Christmas, but for now lets just park this here and come back to it when we both have more free time.

Issue #371, opened 2020-09-10
Device or resource busy after reopening PCM


hrc-amilcar:
Hi, I'm running into an issue where `snd_pcm_hw_params` reports `"Device or resource busy"` if a PCM is closed and re-opened very quickly. The issue is alleviated if I insert a ~500ms sleep between `snd_pcm_close` and the subsequent `snd_pcm_open`. I'm on commit 7ff14a0.

I will try to modify aplay to open -> close -> open to see if I can reproduce it outside of my application. I can't get it to happen on the command line with a while loop around aplay. I'm sure enough time passes between the process ending and restarting.

Here is the output without the sleep:
```
D: src/asound/bluealsa-pcm.c:761: /org/bluealsa/hci0/dev_00_1B_41_B2_86_9A/hfpag/source: Setting constraints
D: src/asound/bluealsa-pcm.c:397: /org/bluealsa/hci0/dev_00_1B_41_B2_86_9A/hfpag/source: Initializing HW
./bluealsa: D: src/dbus.c:57: Called: org.bluealsa.PCM1.Open() on /org/bluealsa/hci0/dev_00_1B_41_B2_86_9A/hfpag/source
./bluealsa: D: src/ba-transport.c:964: New SCO link: 00:1B:41:B2:86:9A: 22
./bluealsa: D: src/hci.c:132: SCO link socket MTU: 22: 180
D: src/asound/bluealsa-pcm.c:426: /org/bluealsa/hci0/dev_00_1B_41_B2_86_9A/hfpag/source: Selected HW buffer: 20 periods x 160 bytes == 3200 bytes
D: src/asound/bluealsa-pcm.c:441: /org/bluealsa/hci0/dev_00_1B_41_B2_86_9A/hfpag/source: Initializing SW
D: src/asound/bluealsa-pcm.c:441: /org/bluealsa/hci0/dev_00_1B_41_B2_86_9A/hfpag/source: Initializing SW
D: src/asound/bluealsa-pcm.c:480: /org/bluealsa/hci0/dev_00_1B_41_B2_86_9A/hfpag/source: Prepared
D: src/asound/bluealsa-pcm.c:441: /org/bluealsa/hci0/dev_00_1B_41_B2_86_9A/hfpag/source: Initializing SW
D: src/asound/bluealsa-pcm.c:480: /org/bluealsa/hci0/dev_00_1B_41_B2_86_9A/hfpag/source: Prepared
[...]

PR #406, opened 2020-12-24
Implement SCO state model

borine:
Attempts to address the problems reported in issue #371

### DIFF
--- a/src/ba-transport.c
+++ b/src/ba-transport.c
@@ -16,6 +16,7 @@
 #include <string.h>
 #include <sys/ioctl.h>
 #include <sys/socket.h>
+#include <sys/timerfd.h>
 #include <unistd.h>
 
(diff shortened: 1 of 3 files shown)


borine:
looks like I broke a2dp with this. more work required in the new year.

borine:
Fixed it - simple clumsy error !!

Arkq:
Thanks for this PR! It seems promising :)

Anyway, I will wait a little bit before merging it, because firstly I have to move to the master some not yet open-sourced commits. For some proprietary A2DP codecs I've got two threads, one for main audio stream and other for back-channel (e.g. microphone). And I'm not sure whether it will be feasible to also split SCO thread into two threads. It will help with PCM draining, but might complicate your state model for delay after SCO closing. I will have to determine which approach will be more maintainable.

PR #407, opened 2020-12-29
A proposal for multi client support

borine:
Implement PCM "mix" and "snoop" in the server, permitting multiple clients to connect to the same bluetooth PCM. This work is incomplete, but I am raising it as a PR here as I do not have resources to develop it any further. Hopefully others can contribute.

I have used it successfully on little-endian hardware (x86_84 and arm) using A2DP/SBC and SCO/CVSD. Chromium browser with multiple tabs seems to work very well in this scenario; also tested successfully with aplay/arecord, VLC, and mpg123.

mpd appears to work well on an intel laptop, but mixing with mpd fails on a RPi 3 (ARM v7, 32-bit OS)

Not tested and therefore probably not working:
- big-endian hardware
- codecs other than SBC and CVSD

Not implemented and therefore definitely not working:
- 24-bit codecs.


borine:
I have a new branch with the mix & snoop code rebased onto your fd-locking branch, which works better than this one, so I am closing this PR and will raise a new one when the refactoring work is merged. I think there is little to be gained from reviewing this now when significant changes will be required (and improvements gained) when fd-locking is merged.

Arkq:
OK, sure. But please be aware, that there is some non-zero chance, that the fd-locking will not be merged in it's current form. I want to kill two birds with one stone, so to speak, but I've faced some obstacles. My idea was to separate decoder and encoder, which will improve maintenance (for bidirectional profiles/codecs, e.g. SCO, FastStream) but now I'm stuck with the problem of file descriptor closing in case of BT socket unexpected termination. Also, current locking is based on simple mutex, but I think it will have to be changed to rwlocks, where read()/write() operations will use read lock, and close() will require write lock.

PR #425, opened 2021-03-16
Transport IO thread model refactoring

Arkq:


src/sco.c
```diff
@@ -316,7 +320,11 @@ void *sco_thread(struct ba_transport_thread *th) {
 		if (pfds[0].revents & POLLIN) {
 			/* dispatch incoming event */
 			switch (ba_transport_thread_recv_signal(th)) {
-			case BA_TRANSPORT_SIGNAL_PING:
```
borine, line 319: github won't let me add a comment to unchanged line, so this is for the case 0: statement on line 307 above. Since poll_timeout is now always == -1, poll() should never time out. So this case, if it should occur, must be an error?

Arkq: Yes, you are right. Error handling code for `poll()` is a boilerplate from `a2dp-audio.c` file. I'm not updating this, because finally I would like to separate encoder and decoder to dedicated threads (actually, I've already done that as a PoC, and right now I'm trying to introduce all required changes in the thread model).

borine (approved):
This is working well for me, also I have the sco state model and multi-client branches rebased to it and working well.

Arkq:
> This is working well for me, also I have the sco state model and multi-client branches rebased to it and working well.

Thanks for testing it. I'm also trying to test that extensively before merging it into master. There is a lot of changes with breaking potential :D I think that the shape of these changes will be more or less like it is right now. I will add few more commits, though.

Arkq:
@borine, I've seen that you have been rebasing your milti client branch on top of fd-locking. For the past few days I've been doing lots of force pushes (due to some fixups in previous commits), I hope that that this did not make you trouble (e.g. lots of conflicts). I think that right now the fd-locking PR is mostly ready. I'll be testing it for the next few days. It would be great if you too could run bluealsa based on fd-locking branch to check whether your use cases works as expected.

As I've said earlier, I've also got a code (not pushed anywhere yet) which separates sco_thread into encoder and decoder threads. HOWEVER, it seems that such approach will not be possible for CVSD codec.... (at lest for Intel chip connected via USB; I will check with UART later this week). The mSBC codes works mostly fine (with latest patches in Kernel). The problem is (I think) that when using two separate threads I've got no synchronization for read/write to SCO link (I've got no time yet to check whether introducing such sync will resolve this issue). I've only tested it with my headset so far, but two-thread based CVSD produces a lot of crackling noises in my headset (data corrections in headset due to missing frames). Anyway, I will try to somehow separate sco_thread (maybe in some other PR), so after that you could try to apply your SCO state model on top of that. Also, I've added a simple scrip (it is available in this PR) for testing open->close->open case, e.g.:
```
./test/integration/test-pcm-open.sh /org/bluealsa/hci0/dev_1C_48_F9_9D_81_5C/hfpag/sink
```

borine:
> I hope that that this did not make you trouble

No trouble - I haven't tried the latest update yet, but hope to do so in next couple of days.

> you could check CVSD with UART chips? 

I rarely use SCO now, but do have some raspberry pis that I might be able to use for testing. It will take me quite some time to get them set up though.

Arkq:
> I rarely use SCO now, but do have some raspberry pis that I might be able to use for testing. It will take me quite some time to get them set up though.

OK, so I will not bother you with testing my work in progress :). I thought that you are using UART connected BT chip on a daily basis (i.e. while developing and testing).

borine:
> I thought that you are using UART connected BT chip on a daily basis

I was last year, but not so much now. I can set up a couple of RPis for SCO testing, just need to obtain power supplies and make the effort to setup sd cards for them. Perhaps in a week or so I'll be ready.

Arkq:
After a long battle, I've finally merged all these changes into the master. When testing it few days ago, I've discovered that it would be nice (and desired) to have keep-alive feature not only for A2DP but also for SCO. So, I've rewritten few previous commits from scratches (sorry for rebase conflicts yet again). I thing that right now encoder/decoder threads management should me simpler and more robust - now there is a dedicated manager thread for managing IO threads :D Also, I've introduced small backward incompatibility. I've renamed command line option for keep-alive timeout (if that's a potential problem I can add hidden option for backward compatibility).

PR #435, opened 2021-04-19
Use last connected PCM device as default

borine:
Introduces new PCM property "Sequence" and uses this to identify the most recently connected device. Clients can then use 00:00:00:00:00:00 to select that device.

### DIFF
--- a/src/bluealsa-dbus.c
+++ b/src/bluealsa-dbus.c
@@ -117,6 +117,10 @@ static GVariant *ba_variant_new_pcm_volume(const struct ba_transport_pcm *pcm) {
 	return g_variant_new_uint16((ch1 << 8) | (pcm->channels == 1 ? 0 : ch2));
 }
 
+static GVariant *ba_variant_new_pcm_sequence(const struct ba_transport_pcm *pcm) {
+	return g_variant_new_uint32(pcm->t->d->seq);
+}
+
 static void ba_variant_populate_pcm(GVariantBuilder *props, const struct ba_transport_pcm *pcm) {
 	g_variant_builder_init(props, G_VARIANT_TYPE("a{sv}"));
 	g_variant_builder_add(props, "{sv}", "Device", ba_variant_new_device_path(pcm->t->d));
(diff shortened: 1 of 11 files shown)


src/bluealsa.h
```diff
@@ -49,6 +49,10 @@ struct ba_config {
 	 * during profile registration. Leave it empty to use any adapter. */
 	GArray *hci_filter;
 
+	/* Device connection sequence number */
+	pthread_mutex_t seq_mutex;
+	uint32_t seq;
```
Arkq, line 54: Some time ago I thought about using atomics in this project. It would require C11 as a minimum compiler, though. However, I'm not sure whether bluez-alsa can be built with pure C99, so maybe the requirement for C11 is already there. Currently, there is no explicit atomic usage, because there is no real use case (ref counts for ba_adapter, ba_device and ba_transport will need mutex anyway). But this counter seems like a legit atomic :) What do you think?

src/shared/dbus-client.c
```diff
@@ -327,22 +327,36 @@ dbus_bool_t bluealsa_dbus_get_pcm(
 	dbus_bool_t rv = TRUE;
 	size_t length = 0;
 	size_t i;
+	dbus_bool_t get_last = bacmp(addr, BDADDR_ANY) == 0;
```
Arkq, line 330: Use real `bool` type. The `dbus_bool_t` above is required because `dbus_bool_t != bool` and I wanted `bluealsa_dbus_` API to be compatible with libdbus-1.

Also, maybe it will be a good idea to add 00:00:00:00:00:00 as a default MAC in the `20-bluealsa.conf` file?

src/bluealsa-dbus.c
```diff
@@ -129,6 +133,7 @@ static void ba_variant_populate_pcm(GVariantBuilder *props, const struct ba_tran
 	g_variant_builder_add(props, "{sv}", "Delay", ba_variant_new_pcm_delay(pcm));
 	g_variant_builder_add(props, "{sv}", "SoftVolume", ba_variant_new_pcm_soft_volume(pcm));
 	g_variant_builder_add(props, "{sv}", "Volume", ba_variant_new_pcm_volume(pcm));
+	g_variant_builder_add(props, "{sv}", "Sequence", ba_variant_new_pcm_sequence(pcm));
```
Arkq, line 136: Put this new property after the `Device` - I think it's more related to device than transport. Please, make the same change in the API txt file and in the bluealsa-iface.c.

src/bluealsa.h
borine: OK, I've now used atomics for this.

src/shared/dbus-client.c
borine: done

src/bluealsa-dbus.c
borine: done

Arkq:
Patch applied with minor fixes (e.g. `__atomic_fetch_add` is a GCC extension not a C11 standard function :) ). It works very nicely!

PR #437, opened 2021-04-29
Service properties

borine:
This code was written as an exercise when I was learning about the dbus iface code. I occasionally find it useful so I decided to share it. There is a lot of code here for possibly little benefit, so if you feel it is not worth the maintenance burden then please just close it - that's OK.

### DIFF
--- a/doc/bluealsa-api.txt
+++ b/doc/bluealsa-api.txt
@@ -21,6 +21,176 @@ Signals         void PCMAdded(object path, dict props)
                         object path is only for a reference - it has been
                         already removed.
 
+Properties      string Version [readonly]
+
+                        Release version of BlueALSA service.
+
+                array{string} Profiles [readonly]
+
+                        Registered Bluetooth Profiles.
+
+                        Possible values: "A2DP-sink", "A2DP-source",
+                                         "HFP-ofono", "HFP-AG", "HFP-HF",
+                                         "HSP-AG" or "HSP-HS"
+
+                array{string} Adapters [readonly]
+
+                        Used HCI adapters. The device names ("hci0", etc) of the
+                        bluetooth adapters that the bluealsa service is using.
+
+                array{string} AdapterFilter [readonly]
+
+                        Configured HCI Adapters. A list of device names or local
+                        adapter MAC addresses that the bluealsa service is
+                        configured to use. If the list is empty, then the
+                        service will use all available adapters.
+
+                dict HFP [readonly]
+
+                        HFP Profile configuration.
+
+                        Possible values:
+
+                                array{string} FeaturesSDPHF
+
+                                        A list of features supported for HFP-HF,
+                                        advertised by Service Discovery.
+
+                                        Possible Values: "CLI", "ECNR", "TWC",
+                                                         "VOLUME", "VREC",
+                                                         "WBAND"
+
+                                array{string} FeaturesSDPAG
+
+                                        A list of features supported for HFP-AG,
+                                        advertised by Service Discovery.
+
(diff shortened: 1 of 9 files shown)


Arkq:
> There is a lot of code here for possibly little benefit, so if you feel it is not worth the maintenance burden then please just close it - that's OK.

No, I think it will be fine to have this, there shouldn't be a lot of burden with maintenance. However, I will pull this not as a single commit. Firstly I will have to review it, because there are some things which will have to be changed. The most important thing is the modification of `config.enable` flags in case of profile register failure. The problem with that is that if there is some issue with BlueZ registration might fail, but later user might restart BlueZ service, and then registration might succeed. With your change, it will not be possible. Silent modification of command line arguments is not OK, IMHO (at least in that case).

borine:
> Silent modification of command line arguments is not OK

Yes, that makes sense. These "properties" are a mix of "configuration" (invariant, build-time or command-line selections) and "state" (variable runtime values). I should have been more careful to keep the two separate. Profile is an example of both, so I should perhaps have "ConfiguredProfiles" and "AvailableProfiles" as separate properties. I'll have to introduce extra global state to do that - will look into it soon. Perhaps a more consistent naming convention too might help the user distinguish the two types of Property.

doc/bluealsa-api.txt
```diff
@@ -21,6 +21,176 @@ Signals         void PCMAdded(object path, dict props)
...
+
+                array{string} Profiles [readonly]
+
+                        Registered Bluetooth Profiles.
+
+                        Possible values: "A2DP-sink", "A2DP-source",
+                                         "HFP-ofono", "HFP-AG", "HFP-HF",
+                                         "HSP-AG" or "HSP-HS"
+
+                array{string} Adapters [readonly]
```
Arkq, line 36: As a first commit, which will introduce properties, I will keep `Version` and `Adapters`. `Version` is a nice to have property. Where, used `Adapters` is something which is not known unless BlueALSA service is running.

`AdapterFilter` is defined via command line, so the purpose of exposing it via D-Bus API would be only as a debug information. But, IMHO, it is better not to clutter API with debug-only properties - stderr/syslog is for debug output.

As for other properties I will have to think a little bit how to structure them. Because currently there is some inconsistency: HFP has sub-properties, where A2DP has not. Also, I'm not sure whether they should be there at all...

`FeaturesSDPHF`, `FeaturesSDPAG`, etc. is more like a configuration. However, I like the verbose output of `FeaturesRFCOMM`, so maybe it will be a good idea to extend RFCOMM properties.

Arkq:
FYI: I've just pulled into the master the commit which adds "list-services" command.

borine:
I've raised  a new PR with just Version and Adapters. I've left this original branch here for reference - if you would like me to rework any of the properties implemented here I will add them as new commits to PR  #444

> it is better not to clutter API with debug-only properties

I was thinking of this more as "troubleshooting" than "debugging". An issue is raised "X isn't working any more" and sometimes you have to play 20 questions before it becomes clear that X is not enabled in the service. I was thinking it would be nice to be able to say "please post output of `bluealsa-cli status` to get all this info in one shot.

graham8:
This proposal looks excellent.  I've been meaning to get my head round a set of troubleshooting steps we can go through with users that report 'X isn't working' without subjecting them to arcane d-bus commands. If this makes that easier, then it gets my vote.

Arkq:
> I was thinking of this more as "troubleshooting" than "debugging". 

Hmm... if troubleshooting is the rationale for this feature, then OK. I will look at this from a different angle :)

PR #444, opened 2021-05-17
Expose service properties to D-Bus API (revised)

borine:
Introduces only "Version" and "Adapters" as Properties for org.bluealsa.Manager1 DBus interface

### DIFF
--- a/doc/bluealsa-api.txt
+++ b/doc/bluealsa-api.txt
@@ -21,6 +21,16 @@ Signals         void PCMAdded(object path, dict props)
                         object path is only for a reference - it has been
                         already removed.
 
+Properties      string Version [readonly]
+
+                        Release version of BlueALSA service.
+
+                array{string} Adapters [readonly]
+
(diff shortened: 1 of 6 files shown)


Arkq:
I've applied the patch with some minor and two major changes. In order to speed up merge process, I've updated the PR by myself (I hope you don't mind). Fixed "issues" with original PR:

1. It is not right to access `ba_adapter` without referencing it with `ba_adapter_ref()` or locking all adapters with mutex. It is possible that "if" which checks for available adapter will pass, and in the same time in some other thread this adapter will be freed, so accessing adapter fields will lead to invalid memory access.
2. I've changed `parse_dbus_string_array` function into more generic `bluealsa_dbus_message_iter_array_get_strings` which work in a similarly to `g_variant_get_strv` but without malloc usage. I did not like this part: `&array[count * strlen]` -accessing elements of `char[][]` via `char[index * element_size]`.

Other that that I've also added "Service" to the listing. So in case of multi bluealsa setup, it will be easier to spot which listing is for which service.

Issue #453, opened 2021-06-01
Daemon is running, headphones are connected, yet alsa is not sending audio to them


czajnik2137:
I have all required services turned on and running correctly, my headphones are connected to bluetooth but I can't get them to play audio. It is directed to my laptop speakers instead. I can tell that bluealsa is functioning as I tried this command

> aplay -D bluealsa Ensoniq-ZR-76-03-Dope-85.wav

which successfully plays file in my headphones.

my .asoundrc
> \# Bluetooth headset
> defaults.bluealsa {
>     interface "hci0"            # host Bluetooth adapter
>     device "20:1B:88:97:1B:9C"  # Bluetooth headset MAC address
>     profile "a2dp"
> }

czajnik2137:
I added a band-aid solution based on some .asoundrc found on the internet, but it isn't perfect either. This script makes my bluetooth headphones the default sound device that receives output, but the default one is ignored when I unplug my headphones. Besides, when I reconnect my headphones they are cranked up to 100% which is... inconvenient to say the least.

> pcm.headphones {
>    type plug
>    slave.pcm {
>        type bluealsa
>        device "20:1B:88:97:1B:9C"
>        profile "a2dp"
>    }
>    hint {
>        show on
>        description "Headphones"
>    }
> }
>
> pcm.!default {
>    type plug
>    slave.pcm "headphones"
> }

borine:
I've written up my thoughts on using BlueALSA as the system default device, as a wiki page here: https://github.com/Arkq/bluez-alsa/wiki/Using-bluealsa-as-default-ALSA-PCM

The examples in that article require the latest bluez-alsa source code, which introduces a new utility called `bluealsa-cli` and also introduces a new feature that uses the most recently connected bluetooth device when the application does not specify which bluealsa device to use.

As you can see from that article, bluetooth audio is not suitable as default for every system - you should consider carefully your own requirements to decide if it is OK for your system.

borine:
> when I reconnect my headphones they are cranked up to 100%

Latest bluez-alsa code now has a run-time option on the `bluealsa` daemon (`--initial-volume=N`) to set the initial volume to less than 100% when a device connects. This may help in your case.

borine:
@czajnik2137 has the referenced wiki article and the new `--initial-volume` option answered your questions here?

borine:
[...]

PR #456, opened 2021-06-02
option to set initial volume level

borine:
A simple change, if the concept is acceptable then hopefully the implementation can be reviewed in time for the coming release.

Addresses a comment that has been made a number of times in issues, most recently here https://github.com/Arkq/bluez-alsa/issues/453#issuecomment-851735570

### DIFF
--- a/doc/bluealsa.8.rst
+++ b/doc/bluealsa.8.rst
@@ -51,6 +51,11 @@ OPTIONS
     Without this option, **bluealsa** enables **a2dp-source**, **hfp-ag** and **hsp-ag**.
     For the list of supported profiles see the PROFILES_ section below.
 
+--init-volume=NB
+    Set the initial volume to *NB*% when the device is connected.
+    *NB* must be an integer in the range from **0** to **100**.
+    The default value is **100** (full volume).
+
 --a2dp-force-mono
     Force monophonic sound for A2DP profile.
 
--- a/src/ba-transport.c
+++ b/src/ba-transport.c
@@ -225,7 +225,6 @@ struct ba_transport *ba_transport_new_a2dp(
 		const char *dbus_path,
 		const struct a2dp_codec *codec,
 		const void *configuration) {
-
 	const bool is_sink = type.profile & BA_TRANSPORT_PROFILE_A2DP_SINK;
 	struct ba_transport *t;
 
@@ -243,12 +242,14 @@ struct ba_transport *ba_transport_new_a2dp(
 			is_sink ? BA_TRANSPORT_PCM_MODE_SOURCE : BA_TRANSPORT_PCM_MODE_SINK);
 	t->a2dp.pcm.soft_volume = !config.a2dp.volume;
 	t->a2dp.pcm.max_bt_volume = 127;
+	t->a2dp.pcm.volume[0].level = t->a2dp.pcm.volume[1].level = config.volume_init_level;
 
 	transport_pcm_init(&t->a2dp.pcm_bc,
 			is_sink ? &t->thread_enc : &t->thread_dec,
 			is_sink ?  BA_TRANSPORT_PCM_MODE_SINK : BA_TRANSPORT_PCM_MODE_SOURCE);
 	t->a2dp.pcm_bc.soft_volume = !config.a2dp.volume;
 	t->a2dp.pcm_bc.max_bt_volume = 127;
+	t->a2dp.pcm_bc.volume[0].level = t->a2dp.pcm_bc.volume[1].level = config.volume_init_level;
 
 	t->acquire = transport_acquire_bt_a2dp;
 	t->release = transport_release_bt_a2dp;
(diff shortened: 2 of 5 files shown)


Arkq:
I think that the idea is OK. Once I though about persistent volume control (like in ALSA), but I come to a conclusion that it should be done in BlueZ, not BlueALSA. BlueZ already has a cache for connected/remembered devices. For instance it remembers lastly used codec for A2DP (I've not checked whether it will persist BlueZ daemon restart, though). So it would be reasonable to keep there a volume level, if sink device does not report/set initial volume.

One thing that I would change (cosmetic change): rename `a2dp.initial_volume` -> `a2dp.volume_init_level` and place it right after `a2dp.volume`. I often use "namespaced" variables, so they are sorted in the auto-completion list in most IDEs/editors.

As for command like option name I also thought about something along `--a2dp-volume-initial-level`, but this will result in too long option.. :/ So, I think that `--a2dp-initial-volume` is OK.

I'm not sure, whether it is right to remove the "Skip volume level initialization in case of A2DP Source profile and software volume control" `if`. Because it's there for a reason. When software volume is enabled, BlueALSA scales volume internally. So, if you set initial "hardware" volume you will have double volume scaling - by BlueALSA and later by sink BT device. Let say that you will set initial volume to 20%, this will result in 20% attenuation by BlueALSA, and later 20% (of this already attenuated sound) by sink device. As a result you will hear nothing.

borine:
> remove the "Skip volume level initialization in case of A2DP Source profile and software volume control" if

OK - that was completely wrong. I've force pushed a new version that hopefully get the initialization right. This sets initial softvol level for both source and sink - I'm not sure if that is right, and I have no way of testing what happens when bluealsa is the sink. Is this correct, or should this option  only be applied where bluealsa is the source?

Arkq:
>  I've force pushed a new version that hopefully get the initialization right. 

Yes, right now the initialization is done in a right way.

I've tested it and I've got a second thought. I think that this feature should be more generic (I'm not sure how much, though). Right now it's only for software volume control, only for A2DP. I think that it should be for SCO as well and not only for software volume. The initialization part is OK, but please remove the `if (!config.a2dp.volume)` - this will resolve the "not only for software volume".

**Rationale:**
Without this patch, the initial value is 0 dB for soft and non-soft volume. With this patch it should be a value specified by a user, so later it could be overwritten by BT device if BT device supports volume notification (either A2DP or HFP). Otherwise (or in case of soft volume), it will stay as a default until user will not change it via BlueALSA D-Bus API (e.g. `amixer -D bluealsa`).

Since I would like it to be also for SCO, now we have to face two things. Firstly, SCO and A2DP uses two different ranges for volume (0-127 vs 0-15). Secondly, SCO (HFP) has a backchannel stream (microphone) with a separate volume control. So, the command line option(s) should allow to set volume for speaker and for mic. With A2DP it is a little bit simpler at first glance (but in reality it's more complicated). A2DP specification does not mention anything about volume control for backchannel, because backchannel is not part of standard spec. However, from the user point of view it would be nice to be able to set the volume for microphone as well (e.g. in case of FastStream codec), even though there is not dedicated volume control in AVRCP spec (maybe it will be in the future). So, the problem is, how to design command line interface for this. As a result we've got a lot of configurations which does not necessarily fit into the command line. So my proposition is to use a single option which will set initial volume (loudness) for both A2DP and SCO for main stream and backchannel, e.g. `--init-volume=X` where X is in range 0-100. What do you think?

borine:
> What do you think?

I'm thinking there's more to this than I first imagined. So I recommend that the 3.1.0 release is done first, and then the ideas in here can be developed with time to experiment with different options.

Having persistent settings per device sounds like an excellent idea. I can't find the Bluez API for this though.

graham8:
I don't think I've come across a BT device that doesn't have automatic volume control on the mic.  As such, I doubt mic volume would get much use.

But my experience is very limited :)

Arkq:
> Having persistent settings per device sounds like an excellent idea. I can't find the Bluez API for this though.

I doubt that there is such public D-Bus API. Device cache (mainly SEPs) is an internal BlueZ thingy.

> As such, I doubt mic volume would get much use.

It might be of no use for many, but might irritate few. I think that for the correctness sake, setting initial volume (and volume scaling) should be for speaker and microphone (in general for main stream and backchannel stream). It's not a technical challenge (this PR already does that) but rather a presentation layer challenge - how to expose these settings to end-user.

My further idea was to add "special" values for `--init-volume` option. In general it would accept integer value, but one could specify e.g. `--init-volume=a2dp:80 --init-volume=sco:50` which will set A2DP initial volume to 80% and SCO initial volume to 50%. Further distinction might be a2dp sink, a2dp source, a2dp sink main channel, a2dp sink backchannel, sco audio gateway speaker, sco audio gateway mic, etc. But the number of possibilities very easy gets out of hands, so maybe that's not a good idea after all :D

Arkq (approved):

Arkq:
Feature applied. I've changed the option name to `--initial-volume`, though. Since there is no `a2dp` prefix, the verbose option name is not too long, so I think it's better to keep it that way.

Issue #156, opened 2018-10-19
 increasing delay as a sink


vano353:
Hello,

I using AAC sink, with blueasla-aplay on Raspberry Pi Zero W. I tried to play with settings  in /src/transport.c 

https://github.com/Arkq/bluez-alsa/blob/dbfa1dc3fe339276955cb9163848d3d4ee49996b/src/transport.c#L867-L870

and flags of bluealsa-aplay 
`--pcm-buffer-time=20000
 --pcm-period-time=10000`

I achieved low delay and sound become almost synch in videos, but the longer I watch the longer the delay. Аlmost 0.5 sec of delay every 2-3 min of watching. 

I don't clearly understand how it happends because I didn't notice any gaps and sound seems to play at normal speed.
So am I doing it right with parameters and how to stop increasing of the sound delay?

Thanks for this tool, using it makes my day happier.

alejandro-n-rivera:
So I'm using my Raspberry Pi 3 Model B as an A2DP sink over Bluetooth. More specifically, I want to use my Samsung phone's dual-audio over Bluetooth feature, which allows it to pair to two Bluetooth speakers at the same time.

I've been trying to get my UE Megaboom speaker to play in-sync with the non-Bluetooth speaker connected to my RPi's audio jack. I also stumbled onto those bluealsa-aplay flags you mentioned: `pcm-buffer-time` and `pcm-period-time`.

I looked more into those flags, and found this on the [aplay man page](https://linux.die.net/man/1/aplay):

> -F, --period-time=#
> Distance between interrupts is # microseconds. If no period time and no period size is given then a quarter of the buffer time is set.
>
> -B, --buffer-time=#
> Buffer duration is # microseconds If no buffer time and no buffer size is given then the maximal allowed buffer time but not more than 500ms is set.
>
> --period-size=#
> Distance between interrupts is # frames If no period size and no period time is given then a quarter of the buffer size is set.
>
> --buffer-size=#
> Buffer duration is # frames If no buffer time and no buffer size is given then the maximal allowed buffer time but not more than 500ms is set.

So it seems bluealsa-aplay only allows us to adjust aplay's `period-time` and `buffer-time`. If we don't set these values, then bluealsa-aplay defaults to a `buffer-time` of 500000us (500ms) and a `period-time` of 100000us (100ms). 

However, I noticed something interesting about the output of the command: \
`bluealsa-aplay -vv 00:00:00:00:00:00`

Output:
```
Selected configuration:
[...]

Issue #451, opened 2021-05-31
Audio clock synchronization issue


mb-arturia:
Hi !
I use bluez-alsa on an embedded ARM system and it works really well with a2dp-sink mode and bluealsa-aplay when it can use the real samplerate clock of the DAC, even at low latency (period_time <2ms, 3 periods).

But when I use it with a sampleclock that is not exact, but some 100 Hz lower, there is a bufferization that grows until getting dropped RTP packets (e.g message "Missing RTP packet: 53662 != 53657" after some time, and then every ~2 seconds)

I have this case when I use a dmix pcm device (forcing 48kHz samplerate and low latency), and the alsa samplerate plug to do the resampling from 44100 Hz to 48kHz. In this case the alsa resampling resulting clock is not exactly 44100 HZ: It is rather below 44000 Hz.

My question:
How can the clocks and the data flow be synchronized ?
Is there a min sampleclock precision necessary ?

Thank you in advance
Markus

borine:
bluealsa-aplay makes no attempt to adjust for clock drift. If the clock on your soundcard is really so far different from the bluetooth controller clock then either constant underruns or dropped a2dp codec frames are inevitable. What I don't understand is how the dmix plugin and/or the rate plugin can cause such rate errors, since both are driven by the hardware clock of the underlying soundcard. If the soundcard is consuming exactly 480 frames per 10ms from its ring buffer, then the rate plugin will consume exactly 441 frames per 10ms from its buffer. There will of course be some increase in jitter (as seen by the application, but not the soundcard), but that will not affect the rate that frames are consumed over any period greater than 10ms.

Also, given that A2DP latency is at least 100ms (or 40ms for so-called "low latency" codecs), and sometimes greater than 300ms, what is the expected benefit of running the soundcard with such a low period time? If you increase the period time to 20ms or 100ms does that reduce the frequency of dropped bluetooth packets?

mb-arturia:
Thank you for your fast response.
One of the main developers of ALSA explained me that "alsa-lib rate plugin tries to align the rate with the period size (in some level), so 44.1kHz to 48kHz isn't exactly from 44.1kHz".
The error grows with smaller period and buffer sizes. (I use 3 periods of 64 samples (1.33 ms) at 48 kHz). I need the low latency setup for my second application running on the system.
[...]

PR #459, opened 2021-06-16
Use Ring Buffer for Aplay

niko132:
As #156 points out when using aplay for speaker functionality, the delay increases while playing audio.
I noticed around 1s of delay every 1h of continuous audio.

From my analysis, this delay comes from the file descriptor filling up because ALSA blocks for too long while executing `snd_pcm_writei()`. But I still don't know why this happens.

However using a ring buffer to constantly flush the file descriptor and only store the most recent ms of audio data (a.k.a. dropping excessive frames) solved the problem for me.

### DIFF
--- a/utils/aplay/Makefile.am
+++ b/utils/aplay/Makefile.am
@@ -7,8 +7,8 @@ bin_PROGRAMS = bluealsa-aplay
 
 bluealsa_aplay_SOURCES = \
 	../../src/shared/dbus-client.c \
-	../../src/shared/ffb.c \
 	../../src/shared/log.c \
+	ring-buffer.c \
 	alsa-pcm.c \
 	dbus.c \
 	aplay.c
--- a/utils/aplay/aplay.c
+++ b/utils/aplay/aplay.c
@@ -312,6 +312,7 @@ static void *pcm_worker_routine(struct pcm_worker *w) {
 	size_t pcm_1s_samples = w->ba_pcm.sampling * w->ba_pcm.channels;
 
 	/* store 50 ms of PCM data in the ring buffer */
+	/* will be adjusted to one period size when pcm is openend */
 	size_t ring_buff_bytes = pcm_1s_samples * pcm_format_size * 50 / 1000;
 	ring_buff_t buffer = { 0 };
 	unsigned char *read_buff = malloc(ring_buff_bytes);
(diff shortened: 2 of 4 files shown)


Arkq:
I will review it in the next week (I've got no access to the PC for whole
this week). However, I think that the solution should mitigate
underclocking (e.g. duplicate samples) and overclocking (e.g. skip
samples).

borine:
> ALSA blocks for too long while executing snd_pcm_writei()

That is because the clock used by the ALSA sink is the playback device clock (ie the soundcard clock), whereas the clock used by the bluetooth source is the bluetooth controller clock (synchronized by the bluetooth protocol to the source device). As no two independent clocks ever "tick" at *exactly* the same rate (at least not for consumer products) the there will always be some drift and jitter when capturing from one device and playing back to another. ALSA can make this worse, because the *rate* plugin introduces rounding error, which when used with small period sizes can be very significant - this is evidenced by #451

An alternative solution, rather than dropping samples or inserting duplicates, is to resample "on the fly" using a variable output sample rate calculated to keep the delay within a narrow range. By this means the "pitch" of the resulting sound is not modified, and the "harsh" audio artifacts caused by discontinuities in the sample stream are avoided. See "loopback" applications such as `alsaloop` or JACK's `zita-j2a` for examples. Unfortunately both those examples are GPL, so we cannot just import their code. After looking at #451 I have thought about proposing a `--sync=` option to `bluealsa-aplay` to enable sync adjustment, but I have not done any work on that idea - I suspect it would not be trivial and may be beyond my capabilities. Perhaps the approach proposed in the PR could be `--sync=simple` whereby the stream is padded with duplicates or frames are dropped to achieve the required playback rate. The existing (default) would be `--sync=none`, and some future enhancement might be `--sync=resample`.

Arkq:
> An alternative solution, rather than dropping samples or inserting
duplicates, is to resample "on the fly" using a variable output sample rate
calculated to keep the delay within a narrow range.

> After looking at #451 *I* have thought about proposing a --sync= option
to bluealsa-aplay to enable sync adjustment. Perhaps the approach proposed
in the PR could be --sync=simple whereby the stream is padded with
duplicates or frames are dropped to achieve the required playback rate. The
existing (default) would be --sync=none, and some future enhancement might
be --sync=resample.

Yes, resamplig would be a better choice, than simple sample
insertion/deletion. I think that there is a libresample library which would
do the job required for that (I've never used it so I'm not sure whether
I'm right).

Asikbct:
Hi, 
I am facing the same audio latency issue. After pausing the audio, it plays for some time (noticeable small delay). 

I did check with this ring buffer patch `Use Ring Buffer for Aplay #459` also, It did not make any changes in my audio latency issue. 

For this issue, I have tried `increasing delay as a sink #156` method, it reduces the latency. But randomly I get audio break (underrun) issue. Is it acceptable or still we can improve the audio latency along with avoid random audio breaks?

is there any other way to get a solution?

Here is our bluealsa and bluetoothd version,

# bluealsa --version
4.1.1
# bluealsa-aplay --version
v4.1.1
# /usr/libexec/bluetooth/bluetoothd --version
5.72

arkq:
Resolved by https://github.com/arkq/bluez-alsa/commit/f0f9427da7d747cfe48b5c83b3076841f48d8703

PR #462, opened 2021-06-23
ALSA CTL fixes

borine:


Arkq:
Wow, many thanks. I knew that something was not right with ctl plugin when disconnecting BT device, but I've got no time to investigate it. And it has turned out that I've broke it :D It seems that one day I will have to add better test coverage for ctl.

Arkq (approved):

Arkq:
Merged into master fa43ce2c89.

PR #469, opened 2021-07-02
Add support for FastStream A2DP source and sink

Arkq:


Arkq:
If anyone is lurking here and has got a headset with FastStream support, I would like to ask for a favor. Currently I've got no possibility to properly test this PR - I've got no real (i.e. from retail market) device with FastStream codec. If that's not much trouble I'd like to know (e.g. via review comment) whether this PR is working or not. Thanks in advance :)

PR #472, opened 2021-07-09
Delay between close/connect calls for SCO links

Arkq:


Arkq:
@borine this PR is a fix for SCO close/connect issue. It's based on your research in PR #406, but due to recent changes in bluealsa thread model, it is greatly simplified - I've simply added a delay before the connect() call. So, there is no need for you to rebase the sco-state-model branch.

borine:
> from where you've got a delay of 600 ms?

There is no scientific basis to that value. That work was done 8 months ago, and I haven't kept my notes from then. I would have started with the 500ms suggested by @hrc-amilcar and so presumably I got (or thought I got) at least one failure at that value. 

I suggest you stay with the lowest value that works for you in the initial merge, and invite all users to try it. After all, any value should better results than the effective 0 minimum time used now.

Arkq:
> There is no scientific basis to that value. That work was done 8 months ago, and I haven't kept my notes from then. I would have started with the 500ms suggested by @hrc-amilcar and so presumably I got (or thought I got) at least one failure at that value.

OK, thanks. So, I'm merging these changes to master.

PR #473, opened 2021-07-18
Integration with systemd - install service files

### LINKED ISSUE #420: Add systemd files to repo?
Can you add the example systemd and config files as part of the repo as opposed to in the wiki? One way is to do it like Mike Brady does with shairport-sync:
https://github.com/mikebrady/shairport-sync where he has a build flag. Another is just to have these in a folder for repurposing when the package is built.

Arkq:
Fixes #420

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -199,6 +199,13 @@ AM_COND_IF([ENABLE_OFONO], [
 	AC_DEFINE([ENABLE_OFONO], [1], [Define to 1 if oFono is enabled.])
 ])
 
+AC_ARG_ENABLE([systemd],
+	AS_HELP_STRING([--enable-systemd], [enable systemd integration]))
+AM_CONDITIONAL([ENABLE_SYSTEMD], [test "x$enable_systemd" = "xyes"])
+AM_COND_IF([ENABLE_SYSTEMD], [
+	PKG_CHECK_MODULES([SYSTEMD], [systemd >= 200])
+])
+
 AC_ARG_ENABLE([upower],
 	AS_HELP_STRING([--enable-upower], [enable UPower integration]))
 AM_CONDITIONAL([ENABLE_UPOWER], [test "x$enable_upower" = "xyes"])
@@ -245,6 +252,13 @@ AM_COND_IF([ENABLE_MANPAGES], [
 	AS_IF([test "x$RST2MAN" = "x"], [AC_MSG_ERROR([[--enable-manpages requires rst2man]])])
 ])
 
+AC_ARG_ENABLE([test],
+	[AS_HELP_STRING([--enable-test], [enable unit test])])
+AM_CONDITIONAL([ENABLE_TEST], [test "x$enable_test" = "xyes"])
+AM_COND_IF([ENABLE_TEST], [
+	PKG_CHECK_MODULES([CHECK], [check >= 0.9.10])
+])
+
 AC_ARG_WITH([bash-completion],
 	AS_HELP_STRING([--with-bash-completion@<:@=DIR@:>@], [install bash completion
 		script in DIR (or the bash completions default directory if not specified)]))
@@ -255,22 +269,13 @@ AS_IF([test "x$with_bash_completion" = "xyes"], [
 		[bashcompdir=$($PKG_CONFIG --variable=completionsdir bash-completion)],
 		[bashcompdir="$datadir/bash-completion/completions"])
 	], [bashcompdir="$with_bash_completion"])
+AC_SUBST([BASH_COMPLETION_DIR], [$bashcompdir])
 
-AC_ARG_ENABLE([test],
-	[AS_HELP_STRING([--enable-test], [enable unit test])])
-AM_CONDITIONAL([ENABLE_TEST], [test "x$enable_test" = "xyes"])
-AM_COND_IF([ENABLE_TEST], [
-	PKG_CHECK_MODULES([CHECK], [check >= 0.9.10])
-])
-
-AC_ARG_WITH([dbusconfdir],
-	AS_HELP_STRING([--with-dbusconfdir=dir], [path to D-Bus system bus configuration files]),
-	[dbusconfdir="${withval}"],
-	[dbusconfdir=$($PKG_CONFIG --variable=sysconfdir dbus-1)/dbus-1/system.d])
 AC_ARG_WITH([alsaplugindir],
-	AS_HELP_STRING([--with-alsaplugindir=dir], [path where ALSA plugin files are stored]),
+	AS_HELP_STRING([--with-alsaplugindir=DIR], [path where ALSA plugin files are stored]),
 	[alsaplugindir="$withval"],
 	[alsaplugindir=$($PKG_CONFIG --variable=libdir alsa)/alsa-lib])
+AC_SUBST([ALSA_PLUGIN_DIR], [$alsaplugindir])
 
 # Unfortunately, for ALSA >= 1.1.7 the directory for add-on configuration files
 # is hard-coded as /etc/alsa/conf.d (unless the distribution has patched the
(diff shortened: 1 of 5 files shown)


borine:
I would like to suggest that the use of `EnvironmentFile` is dropped completely - just delete that line from the two *.service files (and `Environment=` from the aplay service file). Distributions can add their own command line parameters by including systemd override files as:
```
/etc/systemd/system/bluealsa.service.d/10-{distribution-name}.conf
/etc/systemd/system/bluealsa-aplay.service.d/10-{distribution-name}.conf
```
Options can then be changed by local admins with (for example)
```
systemctl edit bluealsa.service
```
An example override file would be:
```
[Service]
Environment=OPTIONS="-p a2dp-sink -p a2dp-source"
```
This seems more in the "pure" systemd spirit, and is completely distribution agnostic.

Arkq:
What about removing `EnvironmentFile` and using executable without any option placeholders in `ExecStart`? If I understand correctly, systemd unit file can be copied to `/etc` and edited by user. So, I think, that there is no need for ".conf" files what so ever.

borine:
> systemd unit file can be copied to /etc and edited by user

Yes, that is often done. In #420 @gearhead says that is how Rune Audio does things. Using override files is an alternative method which others use (if I understand @graham8 correctly, that is how OSMC do it). So a simple `ExecStart=/usr/bin/bluealsa` should be good enough in the upstream file. There is no one "right way" so it is impossible to satisfy all distributions. So choose whichever you feel is easiest to document and will generate the least issues raised.

The "gotcha" with requiring "overriders" to provide an `ExecStart=` line is that directive can be used multiple times. So a simple override file:
```
[Service]
ExecStart=/usr/bin/bluealsa -p a2dp-sink -p a2dp-source
```
actually tells systemd to start a second instance of bluealsa. That is not allowed for services of type "dbus" (nor "simple"), so results in a cryptic error such as :
```
Failed to start bluealsa.service: Unit bluealsa.service has a bad unit file setting.
```
The override needs to be:
```
[Service]
ExecStart=
ExecStart=bluealsa -p a2dp-sink -p a2dp-source
```
This stuff has nothing to do with bluez-alsa, but I fear will inevitably lead to issues raised here by confused users.

Of course, my method of using `Environment="OPTIONS=-p a2dp-sink -p a2dp-source"` may also confuse other users, so in the end clear documentation is the only "right way".

Arkq:
> The "gotcha" with requiring "overriders" to provide an ExecStart= line is that directive can be used multiple times.

@borine from where you've got the info that there might be a problem with overwriting unit files. According to the systemd manual (https://www.freedesktop.org/software/systemd/man/systemd.service.html):

> Unless Type= is oneshot, exactly one command must be given. When Type=oneshot is used, zero or more commands may be specified.

Also, I've tested current unit files and it seems that overwrite is possible and works correctly. I've tested on Raspbina 10. There is one strange thing though (for a non-systemd-oriented user like me). I've copied the `bluealsa.service` to `/etc/systemd/system` and when enabling bluealsa service (`systemctl enable bluealsa`) daemon has created a symlink in `/etc/systemd/system` like this: `dbus-org.bluealsa.service -> /etc/systemd/system/bluealsa.service`. So, after that I've got two bluealsa-related files in the `/etc/stsremd/system` directory. Anyway when issuing `systemctl start bluealsa`, the bluealsa service has started correctly (with added command line options).

borine:
I think I have not explained that very well. Let me go into a little more detail about some basics mainly so that my terminology is better defined. 

First I need to point out that I am not a systemd "expert", although I have used it since it was adopted by Ubuntu and have read (and mostly understood, I hope :smiley:) the official documentation.

Systemd will search for a unit file in a number of locations, in a priority order, and will use only the first one it finds. The two locations of particular interest here are:

- /lib/systemd - this is where upstream "original" unit files shopuld go (e.g the proposed files from the bluez-alsa project)

- /etc/systemd - this is where locally edited unit files go - some distributions put patched files here, some others put their patched files in /lib/systemd and do not distribute the upstream original.

Under each of these are directories called "system" - where system-wide services are defined, and "user" where per-user services are defined.

If there is a file `/etc/systemd/system/bluealsa.service` then it will be used; in that case systemd will not even look at `/lib/systemd/system/bluealsa.service`. 

The problem for local admins then is that if they edit the file in /etc/systemd then their changes may be lost if the distribution later updates it.

So systemd has an alternative method for modifying unit files, which is called "override". An override file is a "snippet" - only containing those lines that should be changed from the original. Override files are placed in a sub-directory named after the unit they apply to. So for bluealsa that would be (for example) /lib/systemd/system/bluealsa.system.d/

Now a distribution can make its own changes to an upstream unit file by placing an override in /lib/systemd/system/bluealsa.service.d/, and a local admin can make further changes by placing another override file in /etc/systemd/system/bluealsa.service.d/ 

So the original upstream file can remain intact, and both distribution and local admin can make "tweaks" without stepping on each other's toes. Override files are therefore the recommended way to maintain variations from upstream maintained originals.

My comments about the "gotcha" apply to these override "snippets", not to copied files in /etc/systemd.

> symlink in /etc/systemd/system like this: dbus-org.bluealsa.service -> /etc/systemd/system/bluealsa.service

That's because the bluealsa service file defines an alias "dbus-org.bluealsa.service". I cannot imagine why that alias is included. The purpose of an alias is that a service can be started under a different name, typically by a different sub-system. Names "dbus-org.*" are for the dbus auto-launch mechanism. Their purpose is to make it clear whether a service was started directly by systemd (via "enable", "start", "requires", etc) or by by dbus when a client requests a service that is not currently running; but it is also necessary to install a d-bus service file in `/usr/share/dbus-1/system-services/` for that to happen - without it the alias achieves nothing. In the case of bluealsa, I think it most unlikely that it should be started whenever a dbus client requests it anyway. For example, `aplay -D bluealsa music.wav` would cause dbus to launch bluealsa - but with no device connected what is the point?

graham8:
I wasn't around when the architecture for systemd in OSMC was devised, but yes we do use overrides and I'd always wondered why an environment file was necessary for bluez-alsa.

Even with this change, we will still have to package the other configuration files so I don't think it makes a lot of difference whether we have your unit files and override them or roll our own.  It would be different if bluez-alsa became part of Debian.

borine:
> I'd always wondered why an environment file was necessary for bluez-alsa.

Not "necessary" - its just the way that I do it, so that's how I wrote the wiki. The example file in the wiki was never intended to be part of a distribution, rather it was/is a simple start for users rolling their own from source. And as a public wiki, others are welcome to contibute alternative suggestions.

I think @Arkq 's suggestion of having no command line args is probably the best compromise. Distributions will have to make their own edits/overrides. My only concern is the potential for more issues to be raised here asking how to configure systemd which really is not what this project is for!

graham8:
Understood.  I nearly put quotes round "necessary" myself.  @Arkq 's suggestion is fine for distributions but end users of distros that don't package bluez-alsa will need their hand holding so there should still be a fully-working example in the wiki.  That should keep the issue count down.

Arkq:
> So systemd has an alternative method for modifying unit files, which is called "override". An override file is a "snippet" - only containing those lines that should be changed from the original. 

Many thanks for the explanation, now I see what is the problem here. Also, searching for override has yielded such result:
https://askubuntu.com/questions/659267/how-do-i-override-or-configure-systemd-services 

> That's because the bluealsa service file defines an alias "dbus-org.bluealsa.service". I cannot imagine why that alias is included. 

I've copied it from `bluez.service`. I've thought that since BlueALSA is also a D-Bus service such line should be in the unit file. I've removed it now. Thanks for the explanation.

> but end users of distros that don't package bluez-alsa will need their hand holding so there should still be a fully-working example in the wiki. That should keep the issue count down.

I will add some tips in distributed unit files. I hope that will suffice for not having issues raised because someone does not know how to change command line options.

Arkq:
Regarding some message for users, I've added something like this: https://github.com/Arkq/bluez-alsa/pull/473/files#diff-79a6fd2ea36c5c8ceefdfd6b1a2289e50267141718bb8785d0862ea07093fcd1
What do you think, will that be enough?

graham8:
Maybe add 'See the bluez-alsa wiki for options'.

PR #483, opened 2021-08-04
Umute PCM on incoming volume change from BT device

### LINKED ISSUE #202: Update mute status on volume update
A2DP Capture Switch is not updated based on volume update. This will create odd behaviour when controlling the volume and mute function. Consider below scenario:
1. BT Sink Device -> bluealsa capture switch - mute -> Bluez -> BT source volume 0
2. BT source volume 10% -> Bluez -> bluealsa capture volume - 10% -> BT Sink Device
Volume is not zero now, but the capture switch was not updated and is still muted. A solution is to update the capture switch everytime a volume update is received on bluez. i.e. mute on 0 volume and unmute on non-zero volume.

Log at end of step 2 above (Note: switch off = muted):
```
amixer -D bluealsa contents '<Device> - A2DP'
../../git/src/shared/bluealsa/ctl-client.c:105: Connecting to socket: /var/run/bluealsa/hci0
../../git/src/shared/bluealsa/ctl-client.c:105: Connecting to socket: /var/run/bluealsa/hci0
../../git/src/shared/bluealsa/ctl-client.c:132: Subscribing for events: 1111111111111111
numid=2,iface=MIXER,name='<Device> Capture Switch'
  ; type=BOOLEAN,access=rw------,values=2
  : values=off,off
numid=1,iface=MIXER,name='<Device> Capture Volume'
  ; type=INTEGER,access=rw------,values=2,min=0,max=127,step=1
  : values=10,10

```

Arkq:
Controlling audio volume should have two parameters: gain level (dB) and mute state. Setting gain level to very low value does not mean that audio will be muted. PCM power will go asymptotically to zero but will never reach it. For zeroing PCM power there should be a mute switch.

Unfortunately, Bluetooth specification (A2DP and HFP/HSP) does not define any standard way of muting audio. Audio volume level is controlled by a property which goes from 0 to MAX (e.g. 15) and by convention audio is muted when the volume level is set to zero.

On the other hand, BlueALSA defines gain level and mute state (like a hardware PCM device). When PCM is muted BlueALSA sets Bluetooth volume level as zero and internally stores PCM volume as muted.

The problem is that once volume is muted by BlueALSA, Bluetooth can not unmute it in any standard way (Bluetooth specification does not define "mute" property). To fix that, from now on, BlueALSA will unmute PCM when there is a signal from a Bluetooth device that the volume should be changed.

Fixes #202

### DIFF
--- a/.github/workflows/build-and-test.yaml
+++ b/.github/workflows/build-and-test.yaml
@@ -1,7 +1,6 @@
 name: CI - Build and Test
 on:
   push:
-    branches: [ master ]
   pull_request:
     branches: [ master ]
 jobs:
--- a/.github/workflows/codeql-analysis.yaml
+++ b/.github/workflows/codeql-analysis.yaml
@@ -1,7 +1,6 @@
 name: CodeQL Analysis
 on:
   push:
-    branches: [ master ]
   pull_request:
     branches: [ master ]
 jobs:
--- a/src/asound/bluealsa-pcm.c
+++ b/src/asound/bluealsa-pcm.c
@@ -207,7 +207,7 @@ static void *io_thread(snd_pcm_ioplug_t *io) {
 	/* We update pcm->io_hw_ptr (i.e. the value seen by ioplug) only when
 	 * a period has been completed. We use a temporary copy during the
 	 * transfer procedure. */
-	snd_pcm_uframes_t io_hw_ptr = pcm->io_hw_ptr;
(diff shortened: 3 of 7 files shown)


Arkq:
In the https://github.com/Arkq/bluez-alsa/issues/202#issuecomment-895928159 @borine has pointed out that approach proposed in this PR might confuse users. I agree with that. Hence, I'm closing this PR without merging it.

PR #485, opened 2021-08-12
Use ALSA config to define PCM initial volume and codec

borine:
I think both these changes are "nice to have" additions rather than essential function. The volume config works well for me, but the codec config is largely untested as I do not have access to suitable devices - it is tested only with A2DP playback with SBC and aptX.

### DIFF
--- a/src/asound/20-bluealsa.conf
+++ b/src/asound/20-bluealsa.conf
@@ -32,7 +32,7 @@ ctl.bluealsa {
 }
 
 pcm.bluealsa {
-	@args [ DEV PROFILE DELAY SRV ]
+	@args [ DEV PROFILE VOL DELAY SRV ]
 	@args.DEV {
 		type string
 		default {
@@ -46,6 +46,10 @@ pcm.bluealsa {
 			@func refer
 			name defaults.bluealsa.profile
 		}
+	}
+	@args.VOL {
+		type string
+		default ""
 	}
 	@args.DELAY {
 		type integer
@@ -64,10 +68,11 @@ pcm.bluealsa {
 	type plug
 	slave.pcm {
 		type bluealsa
-		service $SRV
 		device $DEV
 		profile $PROFILE
+		volume $VOL
 		delay $DELAY
+		service $SRV
 	}
 	hint {
 		show {
--- a/src/asound/20-bluealsa.conf
+++ b/src/asound/20-bluealsa.conf
@@ -32,7 +32,7 @@ ctl.bluealsa {
 }
 
 pcm.bluealsa {
-	@args [ DEV PROFILE CODEC VOL DELAY SRV ]
+	@args [ DEV PROFILE CODEC VOL SOFTVOL DELAY SRV ]
 	@args.DEV {
 		type string
 		default {
@@ -55,6 +55,14 @@ pcm.bluealsa {
 		type string
 		default ""
 	}
+	@args.SOFTVOL {
+		type string
+		default {
+			@func refer
+			name defaults.bluealsa.softvol
+			default ""
+		}
+	}
 	@args.DELAY {
 		type integer
 		default {
@@ -76,6 +84,7 @@ pcm.bluealsa {
 		profile $PROFILE
 		codec $CODEC
 		volume $VOL
+		softvol $SOFTVOL
 		delay $DELAY
 		service $SRV
 	}
--- a/src/asound/20-bluealsa.conf
+++ b/src/asound/20-bluealsa.conf
@@ -32,7 +32,7 @@ ctl.bluealsa {
 }
 
 pcm.bluealsa {
-	@args [ DEV PROFILE VOL DELAY SRV ]
+	@args [ DEV PROFILE VOL CODEC DELAY SRV ]
 	@args.DEV {
 		type string
 		default {
@@ -48,6 +48,10 @@ pcm.bluealsa {
 		}
 	}
 	@args.VOL {
+		type string
+		default ""
+	}
+	@args.CODEC {
 		type string
 		default ""
 	}
@@ -71,6 +75,7 @@ pcm.bluealsa {
 		device $DEV
 		profile $PROFILE
 		volume $VOL
+		codec $CODEC
 		delay $DELAY
 		service $SRV
 	}
--- a/src/asound/20-bluealsa.conf
+++ b/src/asound/20-bluealsa.conf
@@ -32,7 +32,7 @@ ctl.bluealsa {
 }
 
 pcm.bluealsa {
-	@args [ DEV PROFILE VOL CODEC DELAY SRV ]
+	@args [ DEV PROFILE CODEC VOL DELAY SRV ]
 	@args.DEV {
 		type string
 		default {
@@ -47,11 +47,11 @@ pcm.bluealsa {
 			name defaults.bluealsa.profile
 		}
 	}
-	@args.VOL {
+	@args.CODEC {
 		type string
 		default ""
 	}
-	@args.CODEC {
+	@args.VOL {
 		type string
 		default ""
 	}
@@ -74,8 +74,8 @@ pcm.bluealsa {
 		type bluealsa
 		device $DEV
 		profile $PROFILE
-		volume $VOL
 		codec $CODEC
+		volume $VOL
 		delay $DELAY
 		service $SRV
 	}
--- a/src/asound/20-bluealsa.conf
+++ b/src/asound/20-bluealsa.conf
@@ -6,6 +6,15 @@ defaults.bluealsa.device "00:00:00:00:00:00"
 # Default to A2DP connection because that is what
 # most people want to use - high quality audio.
 defaults.bluealsa.profile "a2dp"
+# By default allow bluealsa to negotiate the "best"
+# codec for each PCM.
+defaults.bluealsa.codec "unchanged"
+# By default do not change the volume or mute state
+# when a PCM is opened.
+defaults.bluealsa.volume "unchanged"
+# By default do not modify the software volume state
+# when a PCM is opened.
+defaults.bluealsa.softvol "unchanged"
 defaults.bluealsa.delay 0
 defaults.bluealsa.battery "yes"
 defaults.bluealsa.service "org.bluealsa"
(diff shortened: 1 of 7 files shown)


src/asound/20-bluealsa.conf
```diff
@@ -32,7 +32,7 @@ ctl.bluealsa {
 }
 
 pcm.bluealsa {
-	@args [ DEV PROFILE DELAY SRV ]
+	@args [ DEV PROFILE VOL CODEC DELAY SRV ]
```
Arkq, line 35: Since the order of arguments determines order of values without arg name, I would put `codec` before `vol`. I think that in most cases where PCM is given on command line the codec information will be more common, e.g.:
`bluealsa:33:22:33:22:11:33,a2dp,aac` rather than `bluealsa:33:22:33:22:11:33,a2dp,70`. The initial volume will be more likely in the custom asoundrc file, IMHO.
Also, one more thing. Maybe add codec names in the header file `shared/dbus-client.h` and add helper function for case insensitive mappings between the value for codec argument and actual codec name. The codec in BlueALSA D-Bus API is case sensitive, but for users it will be more convenient to use codec names as case insensitive. Also, I would add aliases for common variants, e.g. `aptx` and `apt-x`.

Arkq (commented):
I like that. Very nice addition.

borine:
I pushed a commit with the simple changes done. The codec refactoring will take a little longer, I'll come back to it when I have more free time.

Arkq:
> he codec refactoring will take a little longer, I'll come back to it when I have more free time.

OK, no hurry. In the mean time I'll try to move select_codec() function into the shared folder (+ small cleanup).

borine:
... actually I've already done a first draft of the refactoring, just won't be able to test it until tomorrow. It allows case-insensitive codec name matching, with aliases, and also accepts a codec configuration as a hex string appended to the codec name. I've used a slightly different signature for select-codec() than your suggestion, but otherwise I think it covers all your recommendations. Will push a new commit tomorrow, unless I get some serious failure in testing, I still can't properly test though, so it is possible some mistakes may slip through. I should have some more time for any required revisions next week.

Arkq:
>  It allows case-insensitive codec name matching, with aliases, and also accepts a codec configuration as a hex string appended to the codec name

OK, great.

OK, so I will wait until you will push your changes, Than I will merge my small "cleanups" with your commits.

Arkq:
One additional thought. Since there will be a possibility to control volume
and mute, maybe it would be worth adding parameter for software-volume. If
no such parameter, then soft-vol will not be changed, and when given it
will be updated before opening PCM. What do you think? With such options
for alsa there will be no need for adding per-device configuration into the
bluealsa service.

borine:
I did not need to add codec names to the client header file, so perhaps I have not correctly understood your idea. However, here is my shot at refactoring the duplicated code. I have added softvol support and codec config support.

Arkq:
> I did not need to add codec names to the client header file, so perhaps I have not correctly understood your idea.

Scratch my idea. I was thinking out loud without revising my stream of thoughts :D With aliasing there will be no (hard) need for codec names in header file.

I will review your latest commit but maybe tomorrow evening or on Monday evening because I might have busy Sunday...

src/asound/20-bluealsa.conf
```diff
@@ -47,6 +47,22 @@ pcm.bluealsa {
 			name defaults.bluealsa.profile
 		}
 	}
+	@args.CODEC {
+		type string
+		default ""
```
Arkq, line 52: Why there is no `name default.bluealsa....`?

src/asound/20-bluealsa.conf
```diff
@@ -47,6 +47,22 @@ pcm.bluealsa {
...
+	@args.VOL {
+		type string
+		default ""
+	}
+	@args.SOFTVOL {
+		type string
+		default {
+			@func refer
+			name defaults.bluealsa.softvol
+			default ""
```
Arkq, line 63: Is `default ""` required with `name default...`?

src/asound/bluealsa-pcm.c
```diff
@@ -1099,6 +1217,17 @@ SND_PCM_PLUGIN_DEFINE_FUNC(bluealsa) {
 		return ret;
 	}
 
+	if (!select_codec(pcm, codec, &err)) {
+		SNDERR("Couldn't select BlueALSA PCM Codec: %s", err.message);
+		dbus_error_free(&err);
+	}
+
+	if (!update_volume(pcm, volume, mute, &err))
+		SNDERR("Couldn't set PCM volume: %s", err.message);
```
Arkq, line 1226:
```suggestion
	if (!update_volume(pcm, volume, mute, &err)) {
		SNDERR("Couldn't set PCM volume: %s", err.message);
		dbus_error_free(&err);
	}
```

src/asound/bluealsa-pcm.c
```diff
@@ -1099,6 +1217,17 @@ SND_PCM_PLUGIN_DEFINE_FUNC(bluealsa) {
...
+
+	if (!update_softvol(pcm, softvol, &err))
+		SNDERR("Couldn't set PCM soft volume property: %s", err.message);
```
Arkq, line 1229:
```suggestion
	if (!update_softvol(pcm, softvol, &err)) {
		SNDERR("Couldn't set PCM soft volume property: %s", err.message);
		dbus_error_free(&err);
	}
```

src/shared/dbus-client.h
```diff
@@ -236,4 +236,12 @@ dbus_bool_t bluealsa_dbus_message_iter_get_pcm_props(
 		DBusError *error,
 		struct ba_pcm *pcm);
 
+dbus_bool_t bluealsa_dbus_pcm_select_codec(
+		struct ba_dbus_ctx *ctx,
+		const struct ba_pcm *pcm,
```
Arkq, line 241: Change this parameter to `const char *pcm_path` and place the definition just after `bluealsa_dbus_open_pcm`. The same thing in the C file.

utils/cli/cli.c
```diff
@@ -370,78 +370,34 @@ static int cmd_codec(int argc, char *argv[]) {
...
-		dbus_set_error(&err, DBUS_ERROR_NO_MEMORY, NULL);
-		goto fail;
-	}
-
-	DBusMessageIter props;
-	if (!dbus_message_iter_open_container(&iter, DBUS_TYPE_ARRAY, "{sv}", &props)) {
-		dbus_set_error(&err, DBUS_ERROR_NO_MEMORY, NULL);
-		goto fail;
-	}
+	uint8_t data[64];
```
Arkq, line 376: Name this variable as `configuration`. Now there is no `char *property = "Configuration"` so it will be harder to guess what kind of data it is.

src/asound/20-bluealsa.conf
borine: I can certainly add that if you prefer. Personally, I am uncomfortable with using the empty string that way. Perhaps an alternative would be to introduce a 'special' value such as "unchanged" as the default, and modify the plugin code to handle that. The other params could also be modified the same way.

... Now I wish I'd thought  of this idea originally, I think it is a more user-friendly approach, do you agree?

Arkq: > Perhaps an alternative would be to introduce a 'special' value such as "unchanged" as the default.

Yes, that seems to be a good idea. The problem with not having "name default" is the consistency with other arguments. Since there are "default.bluealsa..." names for other options, one might expect that all options might be configured that way.

src/shared/dbus-client.c
```diff
@@ -16,10 +16,57 @@
...
+	"apt-x-ll", NULL
+};
+static const char *aliases_aptx_tws[] = {
+	"apt-x-tws", NULL
+};
+
+static const struct ba_codec codecs[] = {
+	{ "SBC", NULL },
+	{ "MP3", NULL },
+	{ "AAC", NULL },
```
Arkq, line 51: I think that the simple structure for alias:name mapping should be enough. There will be not much aliases, and such an array will be more simple to maintain. Compiler should optimize duplicated strings anyway.

```suggestion
static const struct {
    const char *alias;
    const char *name;
} codecs[] = {
	{ "SBC", "SBC" },
	{ "MP3", "MP3" },
	{ "aptX", "aptX" },
	{ "apt-X", "aptX" },
```

borine:
I've drafted some updates for the wiki to document the features in this PR. It is currently visible here:
https://github.com/borine/bluez-alsa/wiki/Using-the-BlueALSA-ALSA-PCM-Plugin

Arkq:
> I've drafted some updates for the wiki to document the features in this
PR.

Great. Also, sorry for not giving you any more feedback regarding this PR,
but currently I've got no time for any open-source activity. Unfortunately,
this will last at least until the end of this month. I'd try to do a second
round review at the beginning of September. And I will have to try it
myself (I have not done this yet). Also, I'm struggling whether codec name
correction `bluealsa_codec_get_canonical_name()` should be embedded into
the bluealsa_dbus_pcm_select_codec() function. My thoughts are that the
dbus-client.h API should reflect the D-Bus API. So, maybe the
bluealsa_codec_get_canonical_name() can be also in the dbus-client.h but as
a separate function which one can call explicitly. I have not made up my
mind yet, though.

>

borine:
Ok. An advantage of making bluealsa_codec_get_canonical_name() accessible to clients is that the plugin would be able to properly check whether the codec needs to be changed. At present it does a simple strcasecmp() which obviously does not match aliases.

I also have another PR just about ready which I think makes some important improvements to both plugins. I will wait until this PR is finally settled before raising that one, however the code is visible in my fork here:

https://github.com/borine/bluez-alsa/tree/alsa-disconnect-improvements

with some draft user notes for the CTL change here:

https://github.com/borine/bluez-alsa/wiki/Using-the-BlueALSA-ALSA-CTL-Plugin.

Arkq:
> An advantage of making bluealsa_codec_get_canonical_name()
> accessible to clients is that the plugin would be able to properly
> check whether the codec needs to be changed.

This check is already done in the BlueALSA service. I'm not sure whether it
will make any difference in robustness if D-Bus communication is skipped.
And beside the codec name check, BlueALSA also compares used A2DP
configuration. I think that the bluealsa_codec_get_canonical_name() is more
like UI enhancement than codec selection logic.

>

Arkq (approved):

Arkq:
Finally, I'm back from long vacations :D In the meantime, I've tested this PR and it works nicely. I've updated code style a little bit + some minor fixes. In the final version I've decoupled `bluealsa_dbus_pcm_select_codec()` and `get_codec_canonical_name()` which should allow some flexibility in the future (maybe it will never be used, but it's nice to have an option).

PS.
In order to spot more warnings, please compile code with `-Wall -Werror`. Since it's very easy to forget about this I've got something like this in my `~/.profile` file:
```
export CFLAGS="${CFLAGS} -Wall -Wextra"
export CXXFLAGS="${CFLAGS}"
```

PR #494, opened 2021-09-17
Alsa plugin updates

borine:
Another round of fixes and enhancements for the ALSA plugins.

### DIFF
--- a/src/asound/bluealsa-ctl.c
+++ b/src/asound/bluealsa-ctl.c
@@ -8,6 +8,10 @@
  *
  */
 
+#if HAVE_CONFIG_H
+# include <config.h>
+#endif
+
 #include <ctype.h>
 #include <errno.h>
 #include <poll.h>
@@ -17,6 +21,7 @@
 #include <stdlib.h>
 #include <string.h>
 #include <sys/param.h>
+#include <unistd.h>
 
 #include <alsa/asoundlib.h>
 #include <alsa/control_external.h>
@@ -80,7 +85,31 @@ struct bluealsa_ctl {
 	/* if true, show battery meter */
 	bool battery;
 
+	/* Disconnection event pipe.
+	 * This allows us to generate a POLLERR event by closing the read end
+	 * then polling the write end. No actual reads or writes are performed
+	 * on this pipe, so no risk of SIGPIPE.
+	 * Many applications (including alsamixer) interpret POLLERR as
+	 * indicating the mixer device has been disconnected. */
+	int pipefd[2];
+
+	/* if true, this mixer is for a single bluetooth device */
+	bool single_device;
 };
+
+static int str2bdaddr(const char *str, bdaddr_t *ba) {
+
+	unsigned int x[6];
+	if (sscanf(str, "%x:%x:%x:%x:%x:%x",
+				&x[5], &x[4], &x[3], &x[2], &x[1], &x[0]) != 6)
+		return -1;
+
+	size_t i;
+	for (i = 0; i < 6; i++)
+		ba->b[i] = x[i];
+
+	return 0;
+}
 
 static int bluealsa_bt_dev_cmp(const void *p1, const void *p2) {
 	const struct bt_dev *d1 = *(const struct bt_dev **)p1;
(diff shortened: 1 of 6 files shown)


Arkq (commented):
I will try to review `bluealsa-ctl.c` tomorrow :)

Also, it seems not right that you've included "Copyright (c) 2016-2021 Arkadiusz Bokowy." in your work. Since this project is no longer developed by me only, I'm inclined to change the copyright statement to something like "Copyright (c) 2016-2021 bluez-alsa authors." in all files. What do you think?

Arkq:
I've just applied to master commits for PCM plugin fixes.

borine:
I will make all the requested changes to the manual page once the review of the ctl  code is complete, just in case that results in changes that impact to the man page.

> I'm inclined to change the copyright statement ... What do you think?

I don't know much about, and have little interest in, copyright management. So I am not a good source of advice on this subject. I have explicitly assigned copyright to you on any new files I have submitted to this project for two reasons:

* I prefer to remain pseudo-anonymous here, and claiming copyright for myself is not helpful in that. It does mean that you are accepting copyright ownership with any responsibilities that go with it (which I do not know what they are), but you have control over whether to commit any change so you can always refuse if you wish.
* It gives you freedom to modify the licence conditions at any time without needing my permission - that may or may not be important to you depending on whether you have any commercial contracts (or ambition for such) with this project.

I think if any other contributors were concerned about the copyright notice they would have said so already, so this matter probably comes down to whatever you personally are most comfortable with - I will not object whatever your decision.

Arkq:
> I prefer to remain pseudo-anonymous here, and claiming copyright for myself is not helpful in that. It does mean that you are accepting copyright ownership with any responsibilities that go with it (which I do not know what they are), but you have control over whether to commit any change so you can always refuse if you wish.

OK, if you are putting it this way, and that is your conscious decision, then let it be that way. I have no objections, but sometimes people do feel awkward of putting their names on work they do if they are contributing to an opensource project and I wanted to be sure that it's not that case with you :) And small side node: from the legal point of view, in Poland, copyright transfer is not possible (you can not create something and than tag it with someone else's name even though the other person will agree with that :D - copyright is non-transferable; however, you can transfer financial rights).

> I will make all the requested changes to the manual page once the review of the ctl code is complete, just in case that results in changes that impact to the man page.

OK, no problem.

src/asound/bluealsa-ctl.c
```diff
@@ -992,23 +1120,90 @@ SND_CTL_PLUGIN_DEFINE_FUNC(bluealsa) {
...
 	}
 
-	if (!bluealsa_dbus_get_pcms(&ctl->dbus_ctx, &ctl->pcm_list, &ctl->pcm_list_size, &err)) {
+	if (!bluealsa_dbus_get_pcms(&ctl->dbus_ctx, &pcm_list, &pcm_list_size, &err)) {
 		SNDERR("Couldn't get BlueALSA PCM list: %s", err.message);
 		ret = -ENODEV;
 		goto fail;
 	}
 
+	if (bacmp(&device, BDADDR_ALL) != 0) {
```
Arkq, line 1129: I think it will be easier to read if the if will be "true". More like `if (BDADDR_ALL) { ... }`
```suggestion
	if (bacmp(&device, BDADDR_ALL) == 0) {
		ctl->pcm_list = pcm_list;
		ctl->pcm_list_size = pcm_list_size;
	}
	else {
```

src/asound/bluealsa-ctl.c
```diff
@@ -938,10 +1043,17 @@ SND_CTL_PLUGIN_DEFINE_FUNC(bluealsa) {
 	snd_config_iterator_t i, next;
 	DBusError err = DBUS_ERROR_INIT;
 	const char *service = BLUEALSA_SERVICE;
+	const char *device_str = NULL;
+	bdaddr_t device;
```
Arkq, line 1047: Name these variables in the same way as in the `bluealsa-pcm.c`: `device` and `ba_addr`. Also you can use inline initialization:
```c
baaddr_t ba_addr = *BDADDR_ALL;
```

src/asound/bluealsa-ctl.c
```diff
@@ -68,6 +73,9 @@ struct bluealsa_ctl {
 	struct ba_pcm *pcm_list;
 	size_t pcm_list_size;
 
+	/* device address */
+	bdaddr_t addr;
```
Arkq, line 77: Is this used anywhere?

src/asound/bluealsa-ctl.c
```diff
@@ -992,23 +1120,90 @@ SND_CTL_PLUGIN_DEFINE_FUNC(bluealsa) {
...
 	strncpy(ctl->ext.driver, "BlueALSA", sizeof(ctl->ext.driver) - 1);
-	strncpy(ctl->ext.name, "BlueALSA", sizeof(ctl->ext.name) - 1);
+
+	if (ctl->single_device) {
+		if (ctl->dev_list_size != 1) {
+			SNDERR("No such audio device: %s", device_str);
+			ret = -ENODEV;
+			goto fail;
+		}
+		strncpy(ctl->ext.name, ctl->dev_list[0]->name, sizeof(ctl->ext.name));
```
Arkq, line 1202: I think that it will be safer to assume that ALSA expect zero-terminated string in `name`. So, please use `sizeof(ctl->ext.name) - 1`.

src/asound/bluealsa-ctl.c
```diff
@@ -992,23 +1120,90 @@ SND_CTL_PLUGIN_DEFINE_FUNC(bluealsa) {
...
+				if (pcm_list[n].sequence >= seq) {
+					seq = pcm_list[n].sequence;
+					latest = n;
+				}
+			}
+
+			bacpy(&device, &pcm_list[latest].addr);
+		}
+
+		for (n = 0; n < pcm_list_size; n++) {
```
Arkq, line 1157: Add a comment that here we are filtering PCMs which belong to selected BT device, or something like that.

src/asound/bluealsa-ctl.c
```diff
@@ -992,23 +1120,90 @@ SND_CTL_PLUGIN_DEFINE_FUNC(bluealsa) {
...
+			if (bacmp(&device, &pcm_list[n].addr) == 0) {
+				struct ba_pcm *tmp = _pcms;
+				if ((tmp = realloc(tmp, (n + 1) * sizeof(*tmp))) == NULL) {
+					SNDERR("Couldn't create PCM info: %s", strerror(ENOMEM));
+					ret = -ENOMEM;
+					free(pcm_list);
```
Arkq, line 1163: Move this `free()` into the `fail:` section.

src/asound/bluealsa-ctl.c
```diff
@@ -1028,6 +1223,10 @@ SND_CTL_PLUGIN_DEFINE_FUNC(bluealsa) {
 fail:
 	bluealsa_dbus_connection_ctx_free(&ctl->dbus_ctx);
 	dbus_error_free(&err);
+	if (ctl->pipefd[0] != -1)
```
Arkq, line 1226: Rebase on current master. These `close()` here will no longer be required.

Arkq (changes requested):

borine:
I've force-pushed an update which attempts to apply all the requested changes to the code, config, and manual page.

src/asound/bluealsa-ctl.c
```diff
@@ -344,19 +395,25 @@ static int bluealsa_create_elem_list(struct bluealsa_ctl *ctl) {
...
+
 		count++;
 
 		elem_list[count].type = CTL_ELEM_TYPE_SWITCH;
 		elem_list[count].dev = dev;
 		elem_list[count].pcm = pcm;
 		elem_list[count].playback = pcm->mode == BA_PCM_MODE_SINK;
-		bluealsa_elem_set_name(&elem_list[count], dev->name, -1);
+		if (ctl->single_device)
+			bluealsa_elem_set_name(&elem_list[count], NULL, -1);
```
Arkq, line 413: Use `name` defined above.

src/asound/bluealsa-ctl.c
```diff
@@ -366,7 +423,11 @@ static int bluealsa_create_elem_list(struct bluealsa_ctl *ctl) {
 			elem_list[count].dev = dev;
 			elem_list[count].pcm = pcm;
 			elem_list[count].playback = true;
-			bluealsa_elem_set_name(&elem_list[count], dev->name, -1);
+			if (ctl->single_device)
+				bluealsa_elem_set_name(&elem_list[count], NULL, -1);
```
Arkq, line 427: Use `name` defined above.

Arkq (changes requested):

Arkq (approved):

Arkq:
I've updated a little bit the code:
- fixed name creation for battery element in single-device mode
- fixed some typos in the bluealsa-plugins manual
- simplified BT device filtering (in-place filtering)

Other than that it was all good according to my tests :) Once again, many thanks for excellent new features!

PR #519, opened 2021-12-30
Command line option for selecting audio codecs

Arkq:
This commit introduces optional new syntax for BT profiles selection.
Like for BT audio codecs, from now on it will be also possible to modify
the list of profiles enabled by default instead of overwriting it.


misc/bash-completion/bluealsa
```diff
@@ -13,6 +13,18 @@ _bluealsa_profiles() {
...
 
+# helper function gets available codecs
+# @param $1 the bluealsa executable name
+_bluealsa_codecs() {
+	"$1" --help | while read -r line; do
+		[[ "$line" = "Available BT audio codecs:" ]] && start=yes && continue
+		[[ "$start" ]] || continue
+		[[ "$line" ]] || break
+		words=(${line//,/})
+		echo ${words[@]:1} | tr '[:upper:]' '[:lower:]'
```
borine, line 24: The `tr` process can be avoided with:
```sh
words="${words[@]:1}"
echo ${words,,}
```

Arkq: OK, thx for the tip. And since you've commented this bash completion script, I've got a question (in fact two questions).

What do you think about the `--codec=` syntax (and `--profile=`) with optional `+`, `-` prefixes? There has to be a default sane setup for both profiles and codecs. Also there should be a possibility to overwrite default values. So, previous approach was simple but maybe not intuitive (when enabling a single new profile, other profiles were gone). So, in order to fix the "not intuitive" part, I thought that maybe the better option would be to allow modifications of the default (compiled-in) setup. Hence, the +/- prefixes. But yet again, there might be a confusion that when using option without a prefix, the default setup is gone (previous behavior).... So, so be honest I'm not sure whether this whole syntax is OK. Maybe, no-prefix should equal to "+" prefix, and there should be a separate special value (or a separate option) for clearing default built-in setup? That's first question.

And the second question. Assuming that the +/- syntax will stay. Maybe you know how to incorporate such prefixes into the bash auto-completion script? I've experimented with `compgen -P + -W "aac sbc"`  but it seems that the `-P` options is not for such a use case... And I would like not to have all values shown 3 times, i.e.: "sbc +sbc -sbc" :D

borine: I'll have a think about the first question - I don't have any immediate suggestions.

The completion issue is because bluealsa completion relies on bash-completion's built-in `_parse_help` function, which recognizes only options prefixed by `-`. To add `+` options would require writing some new function, or perhaps finding some other project that has already solved that problem.

paul-1: Perhaps split the options
—enable_codec
—disable_codec
—enable_profile
—disable_profile

 I’m not sure if the defaulting a codec/profile without +/- is really that helpful.  Start with a set of default items, and then let the user add or subtract options.

Arkq: > I’m not sure if the defaulting a codec/profile without +/- is really that helpful.

Yes, I'm inclined in that direction as well. There will be no `+` prefix, but only `-` (or `+` will do nothing). So when using `bluealsa -p a2dp-sink` there will be a2dp-source, a2dp-sink, hfp-ag and hsp-ag profiles enabled. But it will be possible to disable e.g. hsp-ag by using `-p -hsp-ag`. I think that the most common use case is to add non-default profiles rather then disabling them. The same will be with codecs. But right now there comes the question: which codecs should be enabled by default and which should be optional? I think that reasonable is to enable SBC (mandatory codec) and AAC. I also thought about enabling MP3 codec, but it does not work as efficiently as I anticipated (I will have to revise the a2dp-mpeg.c code someday). Vendor codecs will be enabled only when explicitly specified on command line. This will break backward compatibility for command like options. But I've already removed/renamed some options, so the next release will be a major version bump (I think).

Arkq:
I've pushed revised version of this PR. Right now the `-p` and `-c` options work as follows: when using these options default profiles/codecs are not disabled. These options support "+" and "-" prefixes ("+" is not documented because it behaves as if there was no prefix). Codecs enabled by default: SBC, AAC.

Now I will have to sleep with that, so to speak, in order to think whether such change is OK. But I think it will stay like that (it will fit perfectly with configuration file which might be added in the future). Additionally, I will add new properties in the D-Bus API for querying enabled profiles and codecs (continuation of #437).

PS. I've managed to modify bash-completion script, so it completes `-p` and `-c` options correctly (with +/- prefixes).

Arkq:
I think that this feature mostly ready. The list of enabled codecs can be queried by `bluealsa-cli status`. E.g. when running `bluealsa -p a2dp-sink -c -aac -c aptx-hd -c ldac` the output of `bluealsa-cli status` will be as follows:
```
Service: org.bluealsa
Version: v3.1.0-84-g5bae03c
Adapters: hci0
Profiles:
 - A2DP-source : SBC aptX-HD LDAC
 - A2DP-sink   : SBC aptX-HD LDAC
 - HFP-AG      : CVSD mSBC
 - HSP-AG      : CVSD
```

@borine what do you think abut the output format? You've been using `bluealsa-cli` a lot with scripting, do you think that such output format will be usable?

borine:
> what do you think abut the output format?

I think the leading hyphens are not strictly necessary in the Profiles list, it's perfectly readable and parseable without them - just the indentation is sufficient. But if they are there for some other reason then that's OK, the output is stll simple enough to parse.

PR #522, opened 2022-01-10
Fix bluealsa-aplay when codec is changed

borine:
A change of codec by a connected device may result in a change in
the audio parameters, which in turn necessitates changing the
hw_params of the ALSA output device. For example if a SCO
connection changes from mSBC to CVSD the sampling rate changes
from 16kHz to 8kHz. If left running, the ouput device continues to
play at 16kHz resulting in double-speed playback and constant
underruns.

### DIFF
--- a/utils/aplay/aplay.c
+++ b/utils/aplay/aplay.c
@@ -616,12 +616,45 @@ fail:
 	return NULL;
 }
 
+static bool pcm_hw_params_equal(const struct ba_pcm *ba_pcm_1, const struct ba_pcm *ba_pcm_2) {
+	if (ba_pcm_1->format != ba_pcm_2->format)
+		return false;
+
+	if (ba_pcm_1->channels != ba_pcm_2->channels)
+		return false;
+
+	if (ba_pcm_1->sampling != ba_pcm_2->sampling)
+		return false;
+
+	return true;
+}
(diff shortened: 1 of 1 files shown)


utils/aplay/aplay.c
```diff
@@ -616,12 +616,40 @@ static void *pcm_worker_routine(struct pcm_worker *w) {
...
+		return false;
+
+	return true;
+}
+
+static void pcm_worker_stop(struct pcm_worker *worker) {
+	pthread_rwlock_wrlock(&workers_lock);
+	pthread_cancel(workers->thread);
+	pthread_join(workers->thread, NULL);
+	memcpy(worker, &workers[--workers_count], sizeof(*worker));
```
Arkq, line 636: Calling `memcpy` like this does not "tell" what is going on here. Overwriting input param with last item from workers array looks strange :D. I would opt for other signature for this function. Maybe pass `pcm_path`, so this function could search for a worker by itself.... However, such signature will be awkward to use in `supervise_pcm_worker_start()`...

Also, there is a typo in cancel and join: worker**S**->thread -> worker->thread

Arkq (changes requested):

Arkq (approved):

Issue #266, opened 2019-11-13
HFP plays silence, bluealsa to soundbar


killerkalamari:
When I play in HFP mode (no ofono) to my soundbar or vehicle it appears to play, but I hear no sound. However, when I play from my Android phone (either an actual phone call or using BTmono) I can play to the soundbar and hear it.

Note: When I say "HFP", I'm referencing the Hands-Free Bluetooth Profile Specification, Revision v1.7.2, available here: https://www.bluetooth.com/specifications/profiles-overview/

The solution to the problem is to send unsolicited +CIEV result code messages at the appropriate times, as part of setting up a "call" (audio playback) and then again when ending the call. HFP 4.11 describes the overall audio connection setup process.

HFP 4.18 describes in detail the +CIEV commands to send to the HF to inform it of events surrounding a call. In the bluealsa case, the ATD step providing the phone number from the HF device is not necessary, so that part can apparently be skipped.

HFP 4.15.2 describes how to terminate a call from the AG. 

**EDIT:** Removed all the exploration notes and partially broken proof of concept patch. I'm currently working on a better patch and I'll share that when ready.

killerkalamari:
@Arkq This new patch against 740886b09bc2df10bb1254de5fd6feda91f9ed5b seems to work reliably in my car now. I'm still not sure if transport_acquire_bt_sco is the right place for the pre-sleep. I'm not be covering all cases on the close; I only did happy path for that. Still, I hope this helps!

```
--- bluez-alsa-ce960be902ea9dd9d7dd211afd0ec232433b3eda.orig/src/ba-transport.c	2019-11-11 13:18:58.868404453 -0700
+++ bluez-alsa-ce960be902ea9dd9d7dd211afd0ec232433b3eda/src/ba-transport.c	2019-11-15 13:52:14.083809051 -0700
@@ -713,7 +713,7 @@ int ba_transport_drain_pcm(struct ba_tra
 	 * arbitrary time before releasing transport. In order to make it right,
 	 * there is a requirement for an asynchronous release mechanism, which
 	 * is not implemented - it requires a little bit of refactoring. */
-	usleep(200000);
+	usleep(2000000);
 
 	debug("PCM drained");
 	return 0;
@@ -878,6 +878,7 @@ static int transport_acquire_bt_sco(stru
 	debug("New SCO link: %s: %d", batostr_(&t->d->addr), t->bt_fd);
 
[...]

PR #525, opened 2022-01-26
ALSA Hook PCM for HFP-AG support

borine:
This is a tentative proposal, more a proof-of-concept or discussion piece than a finished work.

Issue #266 proposes adding basic HFP call transfer and termination support to the BlueALSA server when used with HFP-AG profile. This PR addresses the same requirement by adding optional extra functionality to the ALSA client.

I've deliberately kept the new function completely separated from the existing client by implementing it as a Hook PCM with a distinct config file: this is to simplify discussion, experimentation and initial testing. If the new functionality is deemed desirable then the implememtation could possibly be made more efficient by integrating it with the existing ALSA client.

There's no documentation yet, but the idea is that the user is given a new pre-defined PCM called "hfp" that adds call session handling and is used in much the same way as the `bluealsa` PCM:

```
aplay -D hfp sounds.wav
arecord -D hfp:DEV=XX:XX:XX:XX:XX:XX,CODEC=msbc -r 16000 -c 1 -f s16_le recording.wav
```

### DIFF
--- a/.github/spellcheck-wordlist.txt
+++ b/.github/spellcheck-wordlist.txt
@@ -48,6 +48,7 @@ TODO
 UI
 UUID
 VBR
+XDG
 XQ
 
 # Proper Names
@@ -78,6 +79,7 @@ endian
 endianness
 enum
 init
+inode
 middleware
 mutex
 natively
@@ -86,6 +88,7 @@ params
 proc
 reinitialization
 scalable
+shm
 signedness
 unexported
 unmute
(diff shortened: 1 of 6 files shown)


borine:
I've just force-pushed a re-factored version which will be easier to add into the bluealsa pcm plugin, should that be desired, and hopefully is also more robust

borine:
I now feel this product would sit better outside of the bluez-alsa project, and so it is now maintained as a distinct project here: https://github.com/borine/bluealsa-hfp-ag-plugin

Issue #400, opened 2020-11-30
No sound with HFP with USB BCM20702A1 adapter


xyzzy42:
I'm using a BCM20702A1 USB BT interface attempting to get HFP working on some Bose QC-35 headphones.  A2DP is working, but aplay with SCO just produces silence of the appropriate length.  The bluealsa log doesn't indicate any errors.  arecord never receives any data and hangs.

I've tested with another BT adapter, BCM4345C5, which is connected over a UART (with wifi on sdio) and that does work in HFP mode, both playback and recording.  Audio quality is vastly improved over a Macbook Air mid-2012 that doesn't seem to support wide band voice in HFP.

It seems like this is very similar to #258, but that was fixed with a commit to configure broadcom adapters to use the hci interface for audio instead of an I2S interface.  It does appear that the code in fix is active:
```
bluealsa: D: sco.c:166: Checking Broadcom internal SCO routing
bluealsa: D: sco.c:172: Current SCO interface setup: 1 2 0 1 1
bluealsa: D: sco.c:200: Created SCO dispatcher [ba-sco-dispatch]: hci0
```

Once SCO playback is started, I get these messages from the kernel:
```
Bluetooth: hci0: Device does not support ALT setting 6
Bluetooth: hci0: urb 00000000a9df4129 submission failed (90)
Bluetooth: hci0: urb 00000000a9df4129 submission failed (90)
Bluetooth: hci0: sending frame failed (-90)
```
The last two messages then repeat constantly.

xyzzy42:
It might be possible that this USB BT adapter just doesn't support mSBC.  Apparently ALT mode 6 is supposed to be used for mSBC over USB.

I disabled mSBC support in bluealsa, and then this time I remember to run the recompiled src/bluealsa instead of the system installed bluealsa.  With only 8kHz CVSD sound available, it worked.  Audio quality is poor of course.

It would be nice if there was a way to discover if the BT adapter can do WBS or not.  I also see a hack to get a realtek chipset to use alt1 for mSBC.  Perhaps that can work.

Doesn't really seem like a bluealsa bug, unless those `-90` errors are making there way back to bluealsa and it is ignoring them.

xyzzy42:
Hacking the btusb driver to use alt1 instead of alt6 was able to produce audio!

But it sounds bad.  Too loud and with a hiss in the background.  It sounds better with the correctly functioning bt adapter.  So I suspect there is still some problem.

borine:
[...]

PR #550, opened 2022-03-07
Improve SCO socket transfer size selection

borine:
It now seems that the linux btusb module developers will not modify btusb to report the HCI link MTU in the SCO socket options instead of reporting the HCI SCO buffer size. So this PR implements a work-around that obtains the correct write() size by waiting until the first incoming BT message has been read - this is similar to the strategy used by pulseaudio and pipewire.

If accepted, I think issue #400 can be closed since the code should work for all USB interface AlternateSetting values.  I've tested with QCA and Broadcom adapters (both use USB Alt1) and a RPi zero W (Broadcom UART)
Unfortunately I do not have an Alt3 adapter; hopefully some other volunteer can test that, since Alt3 mSBC support is really the whole point of this PR !

### DIFF
--- a/src/sco.c
+++ b/src/sco.c
@@ -31,6 +31,7 @@
 
 #include <glib.h>
 
+#include "audio.h"
 #include "ba-device.h"
 #include "ba-transport-pcm.h"
 #include "bluealsa-config.h"
@@ -149,7 +150,8 @@ static void *sco_dispatcher_thread(struct ba_adapter *a) {
 		pthread_mutex_lock(&t->bt_fd_mtx);
 
 		t->bt_fd = fd;
-		t->mtu_read = t->mtu_write = hci_sco_get_mtu(fd, a);
+		t->mtu_read = t->mtu_write = hci_sco_get_mtu(fd);
+
 		fd = -1;
 
 		pthread_mutex_unlock(&t->bt_fd_mtx);
@@ -235,23 +237,35 @@ static void *sco_cvsd_enc_thread(struct ba_transport_pcm *t_pcm) {
 	struct ba_transport_thread *th = t_pcm->th;
 	struct io_poll io = { .timeout = -1 };
 
-	const size_t mtu_samples = t->mtu_write / sizeof(int16_t);
-	const size_t mtu_write = t->mtu_write;
-
 	ffb_t buffer = { 0 };
 	pthread_cleanup_push(PTHREAD_CLEANUP(ffb_free), &buffer);
 
-	/* define a bigger buffer to enhance read performance */
-	if (ffb_init_int16_t(&buffer, mtu_samples * 4) == -1) {
+	/* The buffer will be resized once the SCO transfer size is known. We
+	 * choose a small value initially (10ms) to ensure that we discard as little
+	 * audio as possible. */
+	if (ffb_init_int16_t(&buffer, 80) == -1) {
 		error("Couldn't create data buffer: %s", strerror(errno));
 		goto fail_init;
 	}
 
 	debug_transport_pcm_thread_loop(t_pcm, "START");
-	for (ba_transport_thread_state_set_running(th);;) {
+	ba_transport_thread_state_set_running(th);
 
-		ssize_t samples = ffb_len_in(&buffer);
-		switch (samples = io_poll_and_read_pcm(&io, t_pcm, buffer.tail, samples)) {
+	/* The message size of the BT SCO socket */
+	uint16_t transfer_bytes = 0;
+
+	/* The number of PCM samples equivalent to one BT transfer. */
+	unsigned int transfer_samples = 0;
+
+	/* use blocking I/O on FIFO until SCO transfer size is known. */
+	ssize_t samples = 0;
+	size_t start_threshold = SIZE_MAX;
+	bool mtu_optimized = false;
+
+	asrsync_init(&io.asrs, t_pcm->sampling);
+
+	for (;;) {
+		switch (samples = io_poll_and_read_pcm(&io, t_pcm, buffer.tail, ffb_len_in(&buffer))) {
 		case -1:
 			if (errno == ESTALE) {
 				ffb_rewind(&buffer);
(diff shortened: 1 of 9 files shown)


src/sco.c
```diff
@@ -208,66 +217,118 @@ int sco_setup_connection_dispatcher(struct ba_adapter *a) {
...
 
 	debug_transport_thread_loop(th, "START");
-	for (ba_transport_thread_set_state_running(th);;) {
+	ba_transport_thread_set_state_running(th);
 
-		ssize_t samples = ffb_len_in(&buffer);
-		if ((samples = io_poll_and_read_pcm(&io, pcm, buffer.tail, samples)) <= 0) {
+	ssize_t samples;
+	/* discard samples until SCO transfer size is known. */
+	while (transfer.bytes == 0) {
```
Arkq, line 277: What will happen if there will be no incoming SCO stream? I now that such a case will not happen (or will it?) in real life, but in principle one can use two blualsa-based devices and try to transfer SCO data only in one direction (like with A2DP).

borine: I've tested bluealsa-bluealsa, and can confirm that a SCO transport *always* delivers packets in both directions between the adapters as soon as the transport is acquired, whether or not any valid audio data is supplied over the HCI. The only exception, I think, is if the adapter does not route incoming packets to the HCI, in which case bluealsa will not see them. I have included a timeout in the decoder thread to catch this case.

xyzzy42:
I tried this with a RTL8761b, which works with Alt3 + 72 byte MTU.  No luck.  Initially impressions is it's not detecting the incoming eSCO data.
```
bluealsa[170724]: /usr/bin/bluealsa: D: codec-msbc.c:82: Initializing mSBC codec
bluealsa[170724]: /usr/bin/bluealsa: D: sco.c:573: IO loop: START: sco_msbc_dec_thread: HFP Audio Gateway (mSBC)
bluealsa[170724]: /usr/bin/bluealsa: D: ba-transport.c:1402: PCM resumed: 14
bluealsa[170724]: /usr/bin/bluealsa: D: sco.c:580: SCO decoder read timeout
bluealsa[170724]: /usr/bin/bluealsa: D: sco.c:591: USB adjusted SCO MTU: 30: 24
bluealsa[170724]: /usr/bin/bluealsa: D: sco.c:224: Setting SCO write MTU 24
```
There is incoming data.  The timing of various points, in milliseconds, HCI xfers are USB, while rest are Bluetooth:
```
-57: HCI Setup Sync Connection (transparent audio, 7.81 kb/sec, only 3-EV3 and larger packets??)
-47: LMP eSCO Link Request (7.5 ms interval, 60 byte packets bidirectional, 2-EV3 packets bidirectional)
-44: LMP Accepted eSCO Link Request
-41: HCI Sync Connection Complete (7.5 ms interval, 60 byte packets)
-33: USB Set Interface Alt 0 (yes, alt 0)
-30: eSCO POLL/NULL packet pair (first eSCO traffic, repeats every 7.5 ms)
-23: eSCO POLL/NULL packet pair
-15: eSCO POLL/NULL packet pair
 -8: eSCO POLL/NULL packet pair
 -1: USB Set Interface Alt 3
  0: eSCO POLL/2-EV3 packet pair (no data out, 60 bytes in), repeats every 7.5 ms
  0: HCI SCO Data out (0 bytes), repeats every 1ms USB frame (not shown)
  2: HCI SCO Data in (0 bytes), repeats every 1ms USB frame (not shown)
  8: eSCO POLL/2-EV3 packet pair
15: eSCO POLL/2-EV3 packet pair
16: HCI SCO Data in (25 bytes), start of 1st HCI data packet, HCI packet size 72 bytes
17: HCI SCO Data in (25 bytes), continue packet from 16
18: HCI SCO Data in (25 bytes), finish packet from 16, repeat this 3 packet burst
```

Few interesting things.
* Basic concept, send data in same size packet that it arrives in, would work.  Data arrives in 72 byte HCI packets and should be sent the same way.
* Accepted packet types bitmask in HCI Setup Sync Connection only sets 3-EV3, 2-EV5, 3-EV5 to true, yet actual connection is using 2-EV3 packets.
* Initially interface is set to alt 0, then switched to alt 3 after 32 ms.
* No Bluetooth data is initially sent on eSCO link (internal audio buffer in headphones filling?) but just 1 ms after alt 3 is selected on USB, incoming data arrives on BT eSCO.  Is this just a coincidence?  Selecting Alt 3 didn't generate any BT traffic I can see, so I don't know how the headphones would know it happened.
* It takes ~59 ms from when the HCI connection complete status is returned from the RT8761 until the first 72 byte payload HCI SCO data packet is finished.
* Three 60 byte data packets arrive on BT before the 1st USB HCI data packet starts.

xyzzy42:
I increased the eSCO timeout from 30 ms to 90 ms.  This fixed it and a MTU of 72 is detected.

There is still a glitch at the beginning of the stream.

borine:
I've increased the SCO timeout to 100ms to make debugging easier. I need to read through the BT protocol layer interactions to  understand the traces for Alt-3. That will take some time as I am busy with other priorities at present.

borine:
I've obtained a ASUS USB_BT500 adapter (usbid 0b05:190e) which has a realtek chip:
```
hciconfig -a
hci0:	Type: Primary  Bus: USB
	BD Address: 04:42:1A:55:44:96  ACL MTU: 1021:5  SCO MTU: 255:11
	UP RUNNING PSCAN 
	RX bytes:1488 acl:0 sco:0 events:157 errors:0
	TX bytes:26835 acl:0 sco:0 commands:157 errors:0
	Features: 0xff 0xff 0xff 0xfe 0xdb 0xfd 0x7b 0x87
	Packet type: DM1 DM3 DM5 DH1 DH3 DH5 HV1 HV2 HV3 
	Link policy: RSWITCH HOLD SNIFF PARK 
	Link mode: SLAVE ACCEPT 
	Name: 'lenovo #2'
	Class: 0x48010c
	Service Classes: Capturing, Telephony
	Device Class: Computer, Laptop
	HCI Version: 5.1 (0xa)  Revision: 0xd9a
	LMP Version: 5.1 (0xa)  Subversion: 0x2883
	Manufacturer: Realtek Semiconductor Corporation (93)
```
The linux `btusb` driver selects Alt3 for this adapter, and this PR correctly sets the SCO socket MTU to 72 bytes for mSBC, and I get good results - no "glitches" at playback startup. However, there is some strangeness with firmware versions. I'm using Ubuntu 20.04 LTS with kernel 5.13.0-37-generic and the linux-firmware package does not include firmware for this adapter. The firmware files required bt the driver are `/lib/firmware/rtl_bt/rtl8761b_fw.bin` and `rtl8761b_config.bin`. However, these names belong to the UART variants; USB names should be `rtl8761bu_*`. I initially tried the firmware files from https://git.kernel.org/pub/scm/linux/kernel/git/firmware/linux-firmware.git/tree/rtl_bt (both the 'b' and 'bu' versions, renaming 'bu' to 'b') but they give an error in dmesg: `Bluetooth: hci0: RTL: extension section signature mismatch`. So then I downloaded the versions from the ASUS support site, and surprisingly both the 'bu' and 'b' variants work well (renaming 'bu' to 'b' so that the driver loads it).

So at present I am assuming that this PR is correct (-ish) , and possibly the problems reported are firmware related. I'll update my system next month and try again with a more recent kernel.

borine:
update on the firmware naming issue: this is a bug in the linux btrtl driver and was fixed in commit https://github.com/torvalds/linux/commit/9fd2e2949b43dea869f7fce0f8f51df44f635d59 which appears in linux 5.14. So for linux 5.13 and earlier, I guess we need to use the vendor's `rtl8761bu_fw` but rename it to `rtl8761b_fw.bin`

borine:
To answer some of the comment points above:

> Accepted packet types bitmask in HCI Setup Sync Connection only sets 3-EV3, 2-EV5, 3-EV5 to true, yet actual connection is using 2-EV3 packets.

The bitmask actually states which packet types will **not** be accepted. 

> Initially interface is set to alt 0, then switched to alt 3 after 32 ms.

`btusb` sends the HCI Setup Synchronous Connection request, then obtains a handle on the USB isochronous endpoint and sets its config to Alt0. I guess this is to gain exclusive access while not consuming any USB bandwidth. When/if a Synchronous Connection Complete response is received, `btusb` then changes the USB isochronous endpoint config to Alt3.

> No Bluetooth data is initially sent on eSCO link (internal audio buffer in headphones filling?) but just 1 ms after alt 3 is selected on USB, incoming data arrives on BT eSCO. Is this just a coincidence? Selecting Alt 3 didn't generate any BT traffic I can see, so I don't know how the headphones would know it happened.

Once the remote has sent Synchronous Connection Complete it begins sending eSCO packets. I'm not sure about the exact timing, but it has nothing to do with the local HCI switch to Alt3. I guess on the local side, once the switch to Alt3 has happened, 1ms represents the first "slot" on the isochronous endpoint.

> It takes ~59 ms from when the HCI connection complete status is returned from the RT8761 until the first 72 byte payload HCI SCO data packet is finished.

This seems to be quite variable. My USB-BT500 has this interval varying from 31ms to 40ms in the testing I have done. I think it best to leave the timeout at 100ms permanently. In a correctly functioning system the timeout will never be needed anyway.

Arkq:
@borine I see that you are working on this and some CI check are failing. After my recent changes to master (added mutexes here and there), test-alsa-pcm might fail on various occasions. I will try to fix that in the nearest future by fixing all warnings reported by TSAN. However, right now, I don't know how to run TSAN properly with bluez-alsa because of pthread cancellation - TSAN and cancellation do not go along very well :/

Also, really sorry for not taking care of this PR for such a long time... I'd like to test that properly, but for that I need a day or two of free time. I've checked this PR with my RPi and laptop and indeed it seems to work properly. However, I do remember that with some setup (I do not remember which), there was no incoming data from SCO socket if remote device was not sending any data. But maybe my memory is flaky....

borine:
With the recent changes to BlueALSA HFP support, and the latest realtek firmware from the linux-firmware , I'm now getting very good results with this branch. If anyone is watching this PR, please can they test and give feedback. Results from all adapter types and brands (not just USB and not just Realtek) would be very much appreciated.

borine:
I'm re-working this to simplify it and take advantage of changes made in release 4.1.0. Hopefully a complete re-write will be available soon.

borine:
This branch has not been maintained for a long time now. The current BlueALSA sources are now working well with Realtek USB devices, so this PR is no longer relevant for those. Intel USB Alt-6 devices (eg AX200) require mSBC MTU of 60 bytes, which is not currently supported by BlueALSA. However it is still not clear whether this PR is a correct approach to that, and as I do not have any such devices to test with, I am withdrawing this PR. I will leave the branch in my fork in case any ideas in it turn out to be relevant in the future.

PR #556, opened 2022-05-06
cli: filter out duplicate codecs

borine:


Arkq:
What do you think about filtering codecs in the server instead of client? I think that duplicated codecs are of no use for anyone...

borine:
> What do you think about filtering codecs in the server instead of client?

see #558

Arkq:
Closing it in favor of #558

PR #557, opened 2022-05-06
Do not enable any BT profile by default

Arkq:
Rationale:

The BlueALSA service can be used in many different ways. There is no
default usage for it. If one wants to build Bluetooth speaker then sink
profiles shall be enabled. If one wants to use Bluetooth headset with a
host on which BlueALSA is running then one shall use source profiles.
Also, there is no default strategy for which sink/source profiles shall
be enabled. We've got A2DP, HSP and HFP.

Taking into account the above statement, it is reasonable to allow a
user to enable suitable set of BT profiles. However, providing a default
set of profiles might be confusing or not easy to use. The first case is
when we provide default profiles, but if user specifies a profile on a
command line, default set of profiles is not used. The latter case is
when we provide a syntax for enabling/disabling BT profiles. Then, in
case when source profiles are enabled by default but user wants only
sink ones, it will be required to manually disable default set, which
is IMHO a configuration burden.

So we have to admit that in case of BT profiles there is no sane
default. From now on BlueALSA will require to specify which profiles
should be enabled.

### DIFF
--- a/doc/bluealsa.8.rst
+++ b/doc/bluealsa.8.rst
@@ -14,7 +14,7 @@ Bluetooth Audio ALSA Backend
 SYNOPSIS
 ========
 
-**bluealsa** [*OPTION*] ...
+**bluealsa** -p *PROFILE* [*OPTION*]...
 
 DESCRIPTION
 ===========
@@ -49,13 +49,10 @@ OPTIONS
     Without this option, the default is to use all available HCI devices.
 
 -p NAME, --profile=NAME
-    Enable or disable *NAME* Bluetooth profile.
-    May be given multiple number of times to enable (or disable) multiple profiles.
+    Enable *NAME* Bluetooth profile.
+    May be given multiple number of times to enable multiple profiles.
 
-    In order to disable given profile (remove it from the list of profiles enabled
-    by default), the *NAME* has to be prefixed with **-** (minus) character.
-
-    Without this option, **bluealsa** enables **a2dp-source**, **hfp-ag** and **hsp-ag**.
+    It is mandatory to enabled at leas one Bluetooth profile.
     For the list of supported profiles see the PROFILES_ section below.
 
 -c NAME, --codec=NAME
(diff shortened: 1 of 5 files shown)


doc/bluealsa.8.rst
```diff
@@ -49,13 +49,10 @@ OPTIONS
...
-    Enable or disable *NAME* Bluetooth profile.
-    May be given multiple number of times to enable (or disable) multiple profiles.
+    Enable *NAME* Bluetooth profile.
+    May be given multiple number of times to enable multiple profiles.
 
-    In order to disable given profile (remove it from the list of profiles enabled
-    by default), the *NAME* has to be prefixed with **-** (minus) character.
-
-    Without this option, **bluealsa** enables **a2dp-source**, **hfp-ag** and **hsp-ag**.
+    It is mandatory to enabled at leas one Bluetooth profile.
```
borine, line 55:
```suggestion
    It is mandatory to enable at least one Bluetooth profile.
```

src/main.c
```diff
@@ -520,6 +511,16 @@ int main(int argc, char **argv) {
 			return EXIT_FAILURE;
 		}
 
+	/* check whether at leas one BT profile was enabled */
```
borine, line 514:
```suggestion
	/* check whether at least one BT profile was enabled */
```

borine:
I personally much prefer this new approach; it is cleaner and easier to explain. I was very uncomfortable with the previous need to explicilty remove profiles when building sink devices, so big thumbs up from me for this!

Arkq:
> I personally much prefer this new approach; it is cleaner and easier to explain. I was very uncomfortable with the previous

Yeap, it was a standard problem when someone who is not using a product designs it :D. Previously I was using mostly bluez-alsa as a source for streaming music to my headphones. However, recently I've started to use it as a sink on my RPi. And to my surprise the command line UI was very annoying. So after some autoreflection I'm changing it to be more user friendly :) Also, I'm doing some preparations for releasing new version, so I'm more prone to backward incompatible changes.

Arkq:
Merged in b9f72d6a335dee3fcd6ed4467a31b1ce83948306

PR #562, opened 2022-05-29
fix ctl plugin single device mode

borine:
It turns out that changing the A2DP codec results in DBus
InterfaceRemoved and InterfaceAdded signals, and this breaks the
CTL plugin single device mode. So we now ignore both those signals
in single device mode and instead rely on the Bluez Device
"Connected" property change to indicate that a single device has
disconnected.

Additionally, other updates that would trigger removal/addition of
control elements in single device mode are also ignored - i.e.
changing of device Alias and addition of the Battery property
after initial creation of control elements.

### DIFF
--- a/doc/bluealsa-plugins.7.rst
+++ b/doc/bluealsa-plugins.7.rst
@@ -271,7 +271,7 @@ As a special case, a single device mixer can be opened with the address **00:00:
 The Predefined **bluealsa** CTL
 -------------------------------
 
-The **bluealsa** CTL has parameters DEV, BAT, and SRV. All the parameters have defaults.
+The **bluealsa** CTL has parameters DEV, BAT, DYN, and SRV. All the parameters have defaults.
 
 CTL Parameters
 ~~~~~~~~~~~~~~
@@ -280,7 +280,10 @@ CTL Parameters
     The device Bluetooth address in the form XX:XX:XX:XX:XX:XX. Device names or aliases are not valid here. The default value is **FF:FF:FF:FF:FF:FF** which selects controls from all connected devices (see `Default Mode` above). Also accepts the special address **00:00:00:00:00:00** which selects the most recently connected device.
 
   BAT
-    Causes the plugin to include a (read-only) battery level indicator, provided the device supports this. If the value is **yes** then the battery indicator is enabled, any other value disables it. The default is **yes**
+    Causes the plugin to include a (read-only) battery level indicator, provided the device supports this. If the value is **yes** then the battery indicator is enabled, any other value disables it. The default is **yes**.
+
+  DYN
+    Enables "dynamic" operation. The plugin will add and remove controls as profiles are connected or disconnected. This is the normal behaviour, so the default value is "**yes**". This argument is ignored in default mode; in that mode operation is always dynamic. There are some applications that are not programmed to handle dynamic addition or removal of controls, and can fail when such events occur. Setting this argument to "no" in single device mode with such applications can protect them from such failures. When dynamic operation is disabled, the plugin never adds or removes any controls. If a single profile is disconnected, then its associated volume control is put into an inactive state, i.e.: read-only with its value and playback/capture switch set to 0.
 
   SRV
     The D-Bus service name of the BlueALSA daemon. Defaults to **org.bluealsa**. See ``bluealsa(8)`` for more information.
@@ -303,6 +306,7 @@ You can define your own ALSA CTL in the ALSA configuration. To do this, create a
     type bluealsa # Bluetooth PCM
     [device STR]  # Device address (default "FF:FF:FF:FF:FF:FF")
     [battery STR] # Include battery level indicator (yes/no, default no)
+    [dynamic STR] # Enable dynamic operation (yes/no, default yes)
     [service STR] # D-Bus name of service (default "org.bluealsa")
   }
 
--- a/src/asound/bluealsa-ctl.c
+++ b/src/asound/bluealsa-ctl.c
@@ -45,6 +45,9 @@ struct ctl_elem {
 	char name[44 /* internal ALSA constraint */ + 1];
 	/* if true, element is a playback control */
 	bool playback;
+	/* For single device mode, if true then the associated profile is connected.
+	 * If false, the element value is zero, and writes are ignored. */
+	bool active;
 };
 
 struct ctl_elem_update {
@@ -88,10 +91,8 @@ struct bluealsa_ctl {
 	size_t elem_update_list_size;
 	size_t elem_update_event_i;
 
-	/* Disconnection event pipe.
-	 * This allows us to generate a POLLERR event by closing the read end
-	 * then polling the write end. No actual reads or writes are performed
-	 * on this pipe, so no risk of SIGPIPE.
+	/* Event pipe. Allows us to trigger events internally and to generate a
+	 * POLLERR event by closing the read end then polling the write end.
 	 * Many applications (including alsamixer) interpret POLLERR as
 	 * indicating the mixer device has been disconnected. */
 	int pipefd[2];
@@ -100,6 +101,8 @@ struct bluealsa_ctl {
 	bool show_battery;
 	/* if true, this mixer is for a single Bluetooth device */
 	bool single_device;
+	/* if true, this mixer adds/removes controls dynamically */
+	bool dynamic;
 
 };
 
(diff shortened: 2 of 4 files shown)


Arkq:
I think that more heuristic approach might be required here. The problem is that the codec change indeed results in disconnect and reconnect of A2DP profile. However, the disconnect signal might also be a legitimate disconnect call - e.g. someone would like to disconnect A2DP:

This will disconnect A2DP from connected BT sink (e.g. headphones):
```sh
gdbus call --system -d org.bluez -o /org/bluez/hci0/dev_1C_48_F9_9D_81_5C -m org.bluez.Device1.DisconnectProfile "0000110b-0000-1000-8000-00805f9b34fb"
```

Later one might connect this profile again (with the same setup):
```sh
gdbus call --system -d org.bluez -o /org/bluez/hci0/dev_1C_48_F9_9D_81_5C -m org.bluez.Device1.ConnectProfile "0000110b-0000-1000-8000-00805f9b34fb"
```

Maybe it's not popular in Linux GUI to disconnect or connect a single profile, but it's a common feature in Android phones, I think. So, maybe CTL in a single device mode should wait some time before removing A2DP controls.... When changing codec it's also possible that the operation will fail (e.g. invalid A2DP capabilities), so one might end up with disconnected A2DP. Or maybe the single device mode should in fact remove controls and add them again... This will allow to disconnect A2DP profile and connect it later.

borine:
> Maybe it's not popular in Linux GUI to disconnect or connect a single profile, but it's a common feature in Android phones, I think

OK, I did not realize that it was common to disconnect single profiles, so indeed this PR is invalid as it stands.

> Or maybe the single device mode should in fact remove controls and add them again...

The problem for me with this is that it would most likely change the numids  of ALSA control elements, which in turn will confuse, or even crash, applications which do not allow for this possibility - after all, the ALSA docs say:
```
Identifying Elements
Each element has the following identifying properties:
  - The numid (a numeric identifier, assigned when the sound card is
    detected, constant while the sound card is kept connected)
  - The interface type (e.g. MIXER, CARD or PCM)
  - The device
  - The subdevice
  - Its name
  - Its index
  ```
https://github.com/alsa-project/alsa-lib/blob/0418fd32c76228bc795d7968a12e3ad5337e40e9/src/control/control.c#L112

Honoring that promise for bluealsa, to make it more like hw, was the original motivation for introducing this feature.

I'll go away and think about this more carefully. I wonder if we could substitute dummy "placeholder" elements to reserve the numids? Why does everything in ALSA have to be so painful?

**Comment by Arkq** [comment]
> I wonder if we could substitute dummy "placeholder" elements to reserve the numids?

I think that it might work even without dummy elements. Maybe (I have not tried this) the element numid can be assigned in the `bluealsa_elem_list()` function using `snd_ctl_elem_id_set_numid()`. In the single device mode the number of controls is well known: [playback, capture] * [SCO, A2DP]. So maybe if some control is not there yet (the profile is not connected), we could reserve the control numid for it. So, in case of element remove/add elements ids will stay the same. With non-single-dev-mode it's not possible, because the number of connected devices is not known.

**Comment by borine** [comment]
The numid is generated by control_ext plugin: see
https://github.com/alsa-project/alsa-lib/blob/0418fd32c76228bc795d7968a12e3ad5337e40e9/src/control/control_ext.c#L92

But for single device mode we can allocate a fixed array of elements with identity fields pre-filled and match them up to bluealsa PCMs (and device battery property) as and when these are discovered via DBus. I am thinking of a boolean attribute "active" that we initialize to false and then set when the profile connects, reset when the profile disconnects. The bluealsa_read_integer and bluealsa_write_integer functions would always return 0 when "active" is false [I have not yet investigated how various applications would react to snd_ctl_elem_write() in that case, so this is still just a hypothesis]

A possible issue with that is that we would always have to allocate both Playback and Capture elements for A2DP even though very few BT devices actually support both.

Perhaps for Battery we could omit from initial allocation, so long as we always allocate it after A2DP and SCO elements it will always have numid=9

For default mode I think the existing behaviour is fine, as it works very well with alsamixer and amixer. Applications that want compatibility with hw ctls must use single-device mode.

**Comment by Arkq** [comment]
> The numid is generated by control_ext plugin:

Ehh... It might be not so easy to overcome this... :/

> A possible issue with that is that we would always have to allocate both Playback and Capture elements for A2DP even though very few BT devices actually support both.

If allocating playback and capture will mean that both these controls will be visible in the alsamixer, than I think that this might be confusing for most users.

**Comment by borine** [comment]
> Ehh... It might be not so easy to overcome this... :/

Yes, it seems impossible to guarantee consistent numids across different applications, or even 2 different mixer sessions within the same application. The best I've been able to do is maintain numids so long as the mixer / hctl handle is not closed. I think commit f1e23dee5c2ade53ac7b280a346640048f7610d4 works reasonably well within alsamixer and avoids invalidating snd_ctl_elem pointers in applications that do not check for element remove events. So it means that single device mode is still useful in some real use cases (eg, for me, MPD). However some scripts that rely on consistent numids from `amixer` invocations are still broken when a single profile disconnects; but as they already fail with bluealsa default mode, so I think that is not a reason to reject single-device mode.

If this approach is acceptable, then I think the manual page will need some text to explain the limitations; I can think about how to express that over the next few days.

**Comment by Arkq** [comment]
I've been thinking about this issue for some time and I've got some thoughts. First of all I've been trying to name the problem here. The problem is not with not-persistent IDs, but (if I'm not mistaken) the problem is that the client application does not handle add/remove events. You've basically stated the same:

> I think commit https://github.com/Arkq/bluez-alsa/commit/f1e23dee5c2ade53ac7b280a346640048f7610d4 works reasonably well within alsamixer and avoids invalidating snd_ctl_elem pointers in applications that do not check for element remove events.

So, in case where old IDs will not change, but new once will appear, app will still not know about that fact (but will it crash?). The same behavior is for normal and single-device mode. If the single-device mode was suppose to fix that issue, then it's definitely wrong approach. Anyway I like the single-device mode for other reasons - one can create mixer per device which is great (it is more like hwdev: single sound card -> single mixer).

But for the core problem, I would try to aim for a global fix.

Some final thoughts:

1. Since the problem is global and the single-device mode is not the solution for that I'd like the single-device mode to behave exactly like a normal mode - it should handle PCM add/remove events (PCM can appear and disappear, but only for a single BT device).
2. If the client application is properly written it will work with current CTL plugin. So, IMHO, default behavior should be as is. But we can support an optional fix for broken clients, e.g. using some ALSA config parameter "bluealsa:00:00:00:00:00:00,SOMEFIX=true"
3. When such a fix is enabled, CTL plugin should do its best to maintain the list of controls as constant. It means that we can not add new elements. Removed elements should be marked as inactive (like is this PR). When the element reappear it should be marked as active.

What do you think?

**Comment by borine** [comment]
> What do you think?

Your suggestions make good sense. I am pleased that you will accept a work-around for applications that ignore SND_CTL_EVENT_MASK_REMOVE (in my own use case this is `mpd`) because, annoyingly and unforgivably, alsa-lib calls `abort()` (via an assertion) instead of returning an EINVAL error when a handle to a removed element is used.

I will look at re-working this PR to implement your suggestions.

The numid issue is, I now think, unsolvable by bluealsa. Numid is used mostly by scripts to grep for a matching control element which is then later manipulated. It is chosen because it is supposed to be an identifier for a specific control element. The requirement imposed by the ext control plugin to have a contiguous series of numids makes it impossible to have a constant map of numids to removable elements. I believe this to be a design error in the ext plugin, because the control API does not impose this requirement. So the best we can do is to document this deviation from hw plugin behaviour in the manual page.

File: doc/bluealsa-plugins.7.rst [hunk 1]
```diff
@@ -280,7 +280,10 @@ CTL Parameters
...
+    Causes the plugin to include a (read-only) battery level indicator, provided the device supports this. If the value is **yes** then the battery indicator is enabled, any other value disables it. The default is **yes**.
+
+  DYN
+    Enables "dynamic" operation. The plugin will add and remove controls as profiles are connected or disconnected. This is the normal behaviour, so the default value is "**yes**". This argument is ignored in default mode; in that mode operation is always dynamic. There are some applications that are not programmed to handle dynamic addition or removal of controls, and can fail when such events occur. Setting this argument to "no" in single device mode with such applications can protect them from such failures.When dynamic operation is disabled, the plugin never adds or removes any controls. If a single profileis disconnected, then its associated volume control is put into an inactive state in which it is read-only; its value is always 0 and and muted.
```
Arkq, line 286:
```suggestion
    Enables "dynamic" operation. The plugin will add and remove controls as profiles are connected or disconnected. This is the normal behaviour, so the default value is "**yes**". This argument is ignored in default mode; in that mode operation is always dynamic. There are some applications that are not programmed to handle dynamic addition or removal of controls, and can fail when such events occur. Setting this argument to "no" in single device mode with such applications can protect them from such failures. When dynamic operation is disabled, the plugin never adds or removes any controls. If a single profile is disconnected, then its associated volume control is put into an inactive state, i.e.: read-only with its value and playback/capture switch set to 0.
```

src/asound/bluealsa-ctl.c
```diff
@@ -1083,6 +1164,13 @@ SND_CTL_PLUGIN_DEFINE_FUNC(bluealsa) {
 		return -EINVAL;
 	}
 
+	/* single Bluetooth device mode */
+	bool single_device_mode = (bacmp(&ba_addr, BDADDR_ALL) != 0);
+
+	/* non-dynamic operation requires single device mode. */
```
Arkq, line 1170: Do we really need this constraint? Maybe someone would like to use non-dynamic with default mode. Or maybe non-dynamic with default mode does not make any sense... :D Hmm...

borine: I guess it's not really necessary, we should allow each user to make her own choice. However, in that case we probably need separate alsa config default keys for default mode dynamic and single-device mode dynamic, since I think it will be common for users to require dynamic on for default mode and dynamic off for single-device. Or perhaps there should be no config key for the DYN arg and each CTL must explicitly disable it if required?

Arkq:
I've reviewed the code and made some changes. The code is in the https://github.com/Arkq/bluez-alsa/tree/dynamic branch. @borine could you sync (force push) this PR with that branch?

While testing this PR I've noticed that the battery control element was not added dynamically. I've fixed that (this bug is in current master) on top of your cleanup commit. Also, I've refactored a little bit activation/deactivation (code moved to sperate functions) for readability reasons. And lastly, I've changed a little bit add/remove skip logic for non-dynamic mode. In your PR the case when bluealsa service is killed (`pkill -9 bluealsa`) was not handled properly (device was not "unplugged"). I think that these are all the changes that I've introduced to your PR. I hope you don't mind :) Could you please test whether this non-dynamic logic still works for you as expected? I think that I've tested all cases but since you've got a real world usage case for that feature maybe you could check that as well before merging into master.

src/asound/bluealsa-ctl.c
Arkq: I think that we can change that later, if anyone will complain.

borine:
I've forced-pushed your changes as requested. My "real world" test case with MPD is not yet right. Whereas with "master" I get an abort in alsa-lib when a single profile is disconnected (which this PR now fixes), instead I get an abort in MPD code with non-dynamic operation when a device disconnects (which was working ok with master).

I need some time to debug this to see if the cause lies in mpd, bluealsa, or alsa-lib, and will get back to you.

Arkq:
>  I get an abort in MPD code with non-dynamic operation when a device disconnects (which was working ok with master)

I think that master is not triggering device "unplug" when BT device disconnects entirely. Maybe MPD does not like that at all?

Arkq (approved):

PR #565, opened 2022-06-02
Let BT profiles configurable in systemd service

liuming50:
We like to have a way to pass the preferred BT profiles to be enabled
in systemd service at build time, this is useful to the OSs that
generated from sources, like Yocto.

To achieve that, we now introduce a configure option:
--with-systemdbtprofiles, and it defaults to
'-p a2dp-source -p a2dp-sink' if not specified, so there is no
functional changes before and after this patch.

Signed-off-by: Ming Liu <liu.ming50@gmail.com>


configure.ac
```diff
@@ -320,6 +320,12 @@ AC_ARG_WITH([systemdsystemunitdir],
 	[systemdsystemunitdir=$($PKG_CONFIG --variable=systemdsystemunitdir systemd)])])
 AC_SUBST([SYSTEMD_SYSTEM_UNIT_DIR], [$systemdsystemunitdir])
 
+AC_ARG_WITH([systemdbtprofiles],
+	AS_HELP_STRING([--with-systemdbtprofiles], [profiles to be enabled in systemd service, defaults to '-p a2dp-source -p a2dp-sink' if not specified]),
```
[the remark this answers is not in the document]
liuming50: @Arkq Good suggestion! Will make a V2 soon.

configure.ac
```diff
@@ -320,6 +320,18 @@ AC_ARG_WITH([systemdsystemunitdir],
 	[systemdsystemunitdir=$($PKG_CONFIG --variable=systemdsystemunitdir systemd)])])
 AC_SUBST([SYSTEMD_SYSTEM_UNIT_DIR], [$systemdsystemunitdir])
 
+AC_ARG_WITH([systemdbluealsaargs],
+	AS_HELP_STRING([--with-systemdbluealsaargs], [profiles to be enabled in bluealsa.service, defaults to '-p a2dp-source -p a2dp-sink' if not specified]),
```
Arkq, line 324: Update help string to reflect purpose (not only profiles). Also, you do not have to check for `AM_COND_IF([ENABLE_SYSTEMD]`. This check is required in `--with-systemdsystemunitdir` because later we are reading data from systemd pkg-config file. That's not the case for this option.

liuming50:
Changes in patch set V3:

1 Adjusted the configure.ac according to @Arkq's suggestion to V2.

configure.ac
liuming50: Thanks for the review, have pushed a V3. @Arkq

Arkq (approved):

PR #571, opened 2022-06-15
Add note on using non-default ALSA PCM devices to bluealsa-aplay.1.rst

danboid:


borine:
I'm not too sure about this. The thing is that `front` ***is*** the name of ***a*** PCM, just not the PCM you waned to use. I have a slight concern that this may confuse users who may think that it is a bluealsa requirement to ignore ALSA configuration defaults and explicilty expand all ALSA configuration arguments, when that really is not true.

danboid:
The documentation should explain what `PCM` is (in the bluealsa/ bluealsa-aplay context) and how a user would find valid device names, no? This is current missing and the reader is expected to know what `pcm` refers to.

graham8:
Agreed, but @borine is correct, you don't **have** to use the full string.  front on it's own will give to the front device for the default card (if it's defined for that card which it usually is).

borine:
"front" on Ubuntu 22.04 is just fine, it is the out-out-the-box ALSA definition. You will find its definition here `/usr/share/alsa/pcm/front.conf`. If it did it work for you, then maybe you have 2 or more sound cards and card 0 was not the one you wanted, or maybe something else in your config is wrong. Also note that `front` typically requires `plug` , and often also `dmix`, and occasionally `softvol` before it can be used effectively by applications

graham8:
Then I guess front is not defined for your default alsa device.  If your default playback device was, say, a USB DAC front on its own would work.  The point is the full device string is not always required.  It depends on alsa's device definition, not on the OS release.

borine:
> that definition of front only applies if in the card conf it is included

Yes, but then I believe the card conf is included. When I plug in a generic USB card that identifies itself as "Device" into Ubuntu 22.04 then I do get exactly that definition for front:CARD=Device. I believe that it is brought in by the generic "USB-Audio" card config.

graham8:
> I believe that it is brought in by the generic "USB-Audio" card config.

It is indeed, but only about 2/3 of the card definitions in /usr/share/alsa/cards have a definition for front. If the user's default card does not `<include>` front.conf, front on its own results in a `Unable to find definition` error.

doc/bluealsa-aplay.1.rst
```diff
@@ -54,6 +54,9 @@ OPTIONS
     Select ALSA playback PCM device to use for audio output.
     The default is ``default``.
 
+    Available ALSA PCM devices can be listed by running **aplay -L**. You must
+    use the full PCM device name eg `front:CARD=Generic_1,DEV=0` not just `front`.
```
graham8, line 58: Suggest "To use any other device than the default audio card, you must use the full device name as listed eg 'plughw:CARD=Generic_1,DEV=0'.

borine: but ... does "plughw:..." appear in the `aplay -L` output for you? it does not for me. I think that there is no "simple" way to explain ALSA PCM configuration to inexperienced users in one or two lines. That, I am sure, is why no other ALSA application attempts to do it. If there is a reliable, comprehensive, beginners' guide out there then we could refer users to it. But I have not yet found one that does not raise more questions than it answers. So my gut feeling is that we should avoid any kind of ALSA tutorial, no matter how simple, within the BlueALSA manual pages. Wikis and Gists is where that kind of material belongs, IMHO. But of course, these are just the ramblings of an old man, what do I know?

Arkq: I've checked my 3 Linux distros and the output is not consistent :D

|Gentoo | Ubuntu 20.04 | Debian 11|
|- | - | -|
|no `plughw:` | `plughw:CARD=PCH,...` | `plughw:CARD=vc4hdmi0,...`
|no `hw:` | `hw:CARD=PCH,...` | `hw:CARD=vc4hdmi0,...`
|`front:CARD=PCH,...` | `front:CARD=PCH,...` | no `front:`

I kinda do agree with @borine , manual pages in bluez-alsa are not ALSA guides.... But also one more line will not hurt anyone unless it will create more doubts. In that case it's better to link proper ALSA tutorial.

borine: The output from `aplay -L` is actually the list of all PCMs ***that have a namehint with `description` defined and `show` set to true***. So it depends on various defaults that my be overridden by some distributions, config nodes created by various additional packages, and config nodes created or overridden by admins and users. There may be other useable PCMs that are not shown in the list simply because they do not have such namehints.

There just is no straightforward answer to this. Putting in nothing (for example `man aplay` says for its `-D` option:
``` 
 -D, --device=NAME
              Select PCM by name
```
is clearly meaningless to anyone using ALSA for the first time. But on the other hand, putting in a statement that is only correct in some cases but incomplete or wrong in others can make thinks worse by discouraging users from searching elsewhere for help. If only there was a "proper" ALSA tutorial, but I don't know of one.

danboid:
Wow!

This is really quite scary stuff. The conclusion seems to be that users can forget about using bluez-alsa unless you're an ALSA developer because there no reliable way to find out the name of the PCM device.

doc/bluealsa-aplay.1.rst
Arkq: > actually the list of all PCMs that have a namehint with description defined and show set to true

Yeap... I have forgotten about that... :D So, even stating that `aplay -L` is the list of PCMs might mislead someone :D But somehow I feel that we *can* add one statement about the `aplay -L` as a hint, but it should be worded differently.

@danboid so that you understand the scale of the problem here. When you run `aplay -L` you probably do not see PCM device named `bluealsa`... but after installing bluez-alsa, there should be such :D And that's one of reasons why we've got PulseAudio and PipeWire :D For a normal Linux user, ALSA is too difficult to use/configure.

borine:
> This is really quite scary stuff. The conclusion seems to be that users can forget about using bluez-alsa unless you're an ALSA developer because there is no easy and reliable way to find out the name of the PCM device.

That's a rather extreme conclusion, and does not match the reality that many thousands of non-developers have been using BlueALSA successfully for many years now.  Plus there is the fact that there are hundreds, possibly thousands of applications that require the user to specify an output or input device using ALSA configuration names, and as far as I am aware *none* of them explain the meaning of ALSA "PCM" in their manual pages.

If you obtain BlueALSA from a distribution repository then your first call for help should be to that distribution. If you obtained it here and built it yourself, then there is the issue list available to ask for help, and contributors here are generally very keen to help with users get the best out of BlueALSA. There is also the wiki, which is public and anyone can contribute, where there is clearly an opportunity for someone to make a big contribution by providing guidance on relevant ALSA topics. There has to be a limit to the scope of manual pages, otherwise they become manual "books" or even "libraries" !!

danboid:
>There has to be a limit to the scope of manual pages, otherwise they become manual "books" or even "libraries" !!

My commit added 2 sentences that would be very helpful in helping a new user get bluez-alsa running. That hardly constitutes a book.

You can'y just say `pcm` and expect people to know what that means. They MIGHT know it stands for Pulse Code Modulation but how does that mean they'd go "Oh yeah! I must need to run `aplay -L` to see the list of PCM devices!" from seeing the letter `pcm`, unless they already know ALSA quite well.

`-D` seems to be THE most important option to getting bluez-alsa running so we can offer it an extra 2 sentences in the man page, I think.

graham8:
This has got out of hand for a simple PR.  I support @danboid's point (but not his attitude). The relevant word here is **example**.  Users should no more expect to see front:, or apparently (new to me) hw: or plughw: in aplay -L than they should expect Generic_1.

At risk of wandering further off topic, I'm intrigued by two things above:

@Arkq on Gentoo, are you not able to specify hw:CARD=foobar?  AFAIK the hw: plugin is always available.  Maybe I'll have a look at Gentoo and see what their config looks like.

@borine 
> actually the list of all PCMs that have a namehint with description defined and show set to true
I know you've said this before, but with Debian at least, any custom pcm I define in .asoundrc shows up in aplay -L, whether or not I give it a namehint.  I haven't come across anything in alsa.conf by which a namehint can be inferred, so I wonder if you are strictly correct.

I'm asking because I'm wondering whether to attempt an ALSA 'tutorial' :P

danboid:
Users shouldn't be required to read any ALSA tutorials to use bluez-alsa as a BT speaker.

Under Pulseaudio, you install bluez and the PA bluetooth plugin, pair your phone and thats it, it just works. I realise there are a few extra complications using the terminal but basically `aplay -L` is the only ALSA command they should need to know except `alsamixer`.

graham8:
> Under Pulseaudio, you install bluez and the PA bluetooth plugin, pair your phone and thats it! BT audio 'just works'.

So it does with bluealsa if the distro has set it up with the correct unit files, which aren't difficult for people that are in charge of distros.

danboid:
I am providing commit and guides, it is constructive criticism. I'm not name calling , I've not gone there at all.

danboid:
If everyone has calmed down now, can we decide what wording we are going to use to explain what `pcm` is in this context and how to find out that value. We cannot expect users to know this and this should be explain here.

I suppose the question is mainly about the example to use for a pcm device. Do weed need to include an example value or not?

doc/bluealsa-aplay.1.rst
danboid: Yes, ALSA is a mess and hence PA, Pipewire etc.

Apparently NetBSD uses the same audio system that Solaris used. I wonder what it might be missing that I need?

Anyone here ever used the OpenBSD audio system, I've forgotten its name? It has be better than ALSA.

danboid:
I will update this commit when someone has worded it to all of the devs satisfaction.

borine:
> I'm wondering whether to attempt an ALSA 'tutorial'

That would be absolutely fantastic, if you have the time and determination to at least get it to the stage of a "beginners" guide - especially for what is precisely the point of this PR which is to enable users to select and specify the appropriate PCM id to achieve their desired input/output requirements. I would certainly be willing to provide review comments and possibly also some original content depending on how much time I can make available.

Arkq:
> On Gentoo, are you not able to specify hw:CARD=foobar? AFAIK the hw: plugin is always available

Yes, of course I'm able to use these, but they are not listed. That's the namehint case. It's not enabled for all names.

> Maybe I'll have a look at Gentoo and see what their config looks like.

I'm very often using Gentoo as a reference. IMHO it has the least number of changes compared with upstreams. So, I gather that the reason why the `hw` and `plughw` are visible in the `aplay -L` output in e.g. Ubuntu is distro-specific configuration (maybe due to complains that how the heck user can know what to use in the `-D` option.... which is kinda true).

doc/bluealsa-aplay.1.rst
Arkq, line 58: OK, so I will use the power of repo owner and will settle this :D (unless someone will say "veto")

```suggestion
    On most Linux distributions, available ALSA PCM devices can be listed by running **aplay -L**.
```

Put it at the end of the `--pcm` help message (after "bluealsa-aplay does not perform..." paragraph; for most users this paragraph will be more important than the `aplay` hint).
Also, please add `aplay(1)` to the "SEE ALSO" list.

If you are worried that this simple statement might not be enough, add "EXAMPLES" section just before "SEE ALSO" (to be honest I'd be glad if you could provide some examples). In that section you can provide real life usage of `bluealsa-aplay` (2-3 examples) where you can use `-D` option with e.g. `front:CARD=USB,DEV=0`.

danboid:
That works for me @Arkq !

I'm going to write a guide for the wiki now.

danboid:
No comments on my guide then?

danboid:
@Arkq I think we were discussing the different free UNIX OS sound systems somewhere but I can't find it now and I wanted to reply.

Yes, FreeBSD uses OSS which is what Linux used before ALSA but they've improved upon it from what Linux used to use.

According to  https://en.wikipedia.org/wiki/Open_Sound_System :+1: 

"[FreeBSD](https://en.wikipedia.org/wiki/FreeBSD) contains an independently developed implementation of the OSS API, which includes, among other things, in-kernel resampling, [mixing](https://en.wikipedia.org/wiki/Audio_mixing_(recorded_music)) (vchans), equalizer, surround sound, and independent volume control for each application. It also supports bit-perfect mode"

I tried Pulseaudio under FreeBSD about a year ago (not sure now if it was FreeBSD 12.x or 13.0) and it didn't handle swapping between multiple devices which is what I mainly use PA for. I'm going to be trying PA under FreeBSD again soon to test for this same thing but I'll file a bug report if its still broke this time.

FreeBSD aren't very good at keeping up with bug reports, unlike Haiku who are very good at that.

graham8:
> Here's the guide I promised:

Good start.  Just a few comments.

I don't understand why you would edit the unit file (by the way, you should probably refer to /lib/systemd/system/bluez-alsa.service rather than the link under /etc although the effect is the same).  I haven't installed Ubuntu server but I assume the distro unit file has sink, source and volume in the ExecStart line.  Is there a reason to remove source and volume?

>ad2p-sink is the only bluez-alsa profile required for a bluetooth speaker but if you are using a headset you would need to add -p a2dp-source to this command.

I got confused here.  Suggest you don't use the term headset which implies both a headphone and a microphone. Using a headset with Ubuntu server seems uber-niche but if a user wanted that, the headset or handsfree profiles would be appropriate, not a2dp.

If you want to send BT audio from, eg, your smart phone to Ubuntu, then you need to start bluealsa-aplay but the official way to do that is with the unit file that establishes bluealsa-aplay when you boot.  I don't know if that file is provided by the distro, but you can find it here in this repo.

There are actually three or four steps to making a BT connection.  In bluetoothctl, you will likely need `agent on` as well as `scan on`. Agent is the thing that deals with PIN codes which I'm surprised you didn't need to pair your smart phone.  After you've paired your device, type `trust XX;XX;XX;XX;XX:XX` before `connect XX;XX;XX;XX;XX:XX` (no need to quit bluetoothctl to connect).  With trust, you can just select your Ubuntu device from the list on your smart phone to reconnect.  You shouldn't need `bluetoothctl connect` on a regular basis or at all.

HTH

danboid:
Hi Graham

> I don't understand why you would edit the unit file (by the way, you should probably refer to /lib/systemd/system/bluez-alsa.service rather than the link under /etc although the effect is the same). 

Thats a good point. I've changed that path to the service now..

>I haven't installed Ubuntu server but I assume the distro unit file has sink, source and volume in the ExecStart line. Is there a >reason to remove source and volume?

This would seem be be a bug in the Ubuntu 22.04 package. Its ExecStart line looks like:

```
ExecStart=/usr/bin/bluealsa $OPTIONS
```

I have tried the bluez-alsa service like this but it doesn't work. I had to edit it like I described.


> There are actually three or four steps to making a BT connection. In bluetoothctl, you will likely need `agent on` as well as `scan on`. Agent is the thing that deals with PIN codes which I'm surprised you didn't need to pair your smart phone. After you've paired your device, type `trust XX:XX:XX:XX:XX:XX` before `connect XX:XX:XX:XX:XX:XX` (no need to quit bluetoothctl to connect). With trust, you can just select your Ubuntu device from the list on your smart phone to reconnect. You shouldn't need `bluetoothctl connect` on a regular basis or at all.
> 
> HTH

Ah right! I shall test that that now.

danboid:
Hi Graham

Yes, the trust command stops me having to manually connect but Ubuntu doesn't auto-launch bluealsa-aplay for me when I start playing audio after connecting.

graham8:
> Ubuntu doesn't auto-launch bluealsa-aplay

https://github.com/Arkq/bluez-alsa/wiki/Systemd-integration#running-bluealsa-aplay-as-a-systemd-service

danboid:
Hi Graham

I now have bluez-alsa correctly configured under Ubuntu 20.04.

I didn't have to edit the service file, instead it was a case of editing `/etc/default/bluez-alsa` to configure its options. I now have bluealsa-aplay starting as a systemd service but the example code for the service given on the wiki didn't quite work as given so I have fixed that as well as my Ubuntu 22.04 guide.

borine:
hi @danboid Please can you give more detail on what "didn't quite work" for you with the bluealsa-aplay systemd guide? It has been working fine for very many users for over two years and the original version is still working fine here for me now with Ubuntu 22.04. If there is some specific error in there that you have discovered I would prefer to debug it and put in a proper fix rather than requiring users to edit the service unit file. The whole point of using `/etc/defaults/<servicename>` is that is avoids editing files installed by the package manager - that has been the way it has been done on Debian based systems since Debian moved to systemd many years ago.

danboid:
> hi @danboid Please can you give more detail on what "didn't quite work" for you with the bluealsa-aplay systemd guide? It has been working fine for very many users for over two years and the original version is still working fine here for me now with Ubuntu 22.04. If there is some specific error in there that you have discovered I would prefer to debug it and put in a proper fix rather than requiring users to edit the service unit file. The whole point of using `/etc/default/<servicename>` is that is avoids editing files installed by the package manager - that has been the way it has been done on Debian based systems since Debian moved to systemd many years ago.

Yes, these were the problematic three line if used unedited, before I fixed them in the wiki example:

```
Environment="BT_ADDR=00:00:00:00:00:00"
EnvironmentFile=/etc/default/bluealsa-aplay
ExecStart=/usr/bin/bluealsa-aplay $OPTIONS $BT_ADDR
``

To fix, I commented out the two Environment lines because most uses will not need these and it stops the service launching if you don't create the `/etc/default/bluealsa-aplay` with the correct options. If the user used the example `/etc/default/bluealsa-aplay` it very likely wouldn't work hence these should be commented by default. I had to remove `$OPTIONS $BT_ADDR` from ExecStart for these same reasons.

**Comment by borine** [comment]
OK , I can see where you have gone wrong. The *original* service file has the line:
```
EnvironmentFile=-/etc/default/bluealsa-aplay
```
whereas the file you have has
```
EnvironmentFile=/etc/default/bluealsa-aplay
```
Notice that you have left out the minus sign before the file name? That is why systemd is failing. You need to use the original *as written* :smiley: 

I will restore the original. This stuff is really well tested you know !

In any case this way of specifying changes to system unit files is now deprecated and the new BlueALSA sources have systemd files included. The wiki  requires updating for this, and to document the "official" systemd way of overriding unit directives instead of the old Debian way. I will get round to that eventually, but I stll have a long list of other priorities to get through first.

**Comment by danboid** [comment]
Ah OK! I thought the minus was a typo but clearly it was intentional. I'm not a systemd expert by any means so I'll leave further edits and corrections up to you. The example didn't work for me under Ubuntu 22.04 anyway.

**Comment by danboid** [comment]
It would be good to see the `bluez-alsa-utils` include a service for `bluealsa-aplay` as standard, if not enabled by default. Do you know why it doesn't?

**Comment by borine** [comment]
You need to ask Ubuntu about that, this project plays no part in the packaging process for any distributions.

**Comment by borine** [comment]
> The example didn't work for me under Ubuntu 22.04 anyway.

I'm still keen to know what was wrong. If there is a bug lurking in there then it really does need to be fixed.

**Comment by borine** [comment]
Back to the PR :smile:

I still think that presenting a novice ALSA user with the output of `aplay -L` is just going to confuse them even more. If you don't know what an ALSA pcm is, then you have no chance of making sense of that output. Here is a patch for an alternative approach aimed specifically at users who have no ALSA experience.

```diff
diff --git a/doc/bluealsa-aplay.1.rst b/doc/bluealsa-aplay.1.rst
index 9541268..20598de 100644
--- a/doc/bluealsa-aplay.1.rst
+++ b/doc/bluealsa-aplay.1.rst
@@ -62,6 +62,32 @@ OPTIONS
     transformations to the stream. For this reason it is often necessary to use
     the ALSA **dmix** and **plug** plugins in the *NAME* PCM.
 
+    For most distributions, the installed definition of ``default`` for each
+    sound card includes both ``dmix`` and ``plug`` where necessary, so is
+    generally the best choice unless there is some specific reason to prefer
+    some other device. If there is more than one sound card attached then the
+    appropriate one can be selected with:
+
+    ::
+
+        -D default:CARDNAME
+
+    A list of attached sound card names can be obtained using **aplay -l**,
+    for example
+
+    ::
+
+        $ aplay -l
+        card 0: PCH [HDA Intel PCH], device 0: ALC236 Analog [ALC236 Analog]
+          Subdevices: 1/1
+          Subdevice #0: subdevice #0
+        card 1: Device [USB Audio Device], device 0: USB Audio [USB Audio]
+          Subdevices: 1/1
+          Subdevice #0: subdevice #0
+
+    Here the first word after the card number is the card name, so
+    **default:PCH** is the first card and **default:Device** is the second.
+
 --pcm-buffer-time=INT
     Set the playback PCM buffer duration time to *INT* microseconds.
     The default is 500000. It is recommended to choose a buffer time that is
```

Arkq:
@borine How about keeping the `-D` option as it is (even without `aplay -L`), but instead of that add some example which will explain in what circumstances `-D` needs to be used? I think that such approach would be more man-like.

borine:
> add some example which will explain in what circumstances -D needs to be used

Sure, if you have such an example in mind. To be honest I find it ***really*** difficult to come with an *example* that I am happy with (possibly because, at heart, I believe this hand-holding of ALSA novices is out of scope for BlueALSA manual pages, and belongs in Gists, Blogs and Wikis). Perhaps some others are better at that than me. What I can do is move the text from that patch into a new section, perhaps **Selecting An ALSA Playback PCM**, and then maybe others can tweak it and expand it as required.

Arkq:
Ok, I'll try to think about something. Personally I'm not using
bluealsa-aplay very much, but maybe my examples will make some sense :D

borine:
Here's a revised patch
```diff
diff --git a/doc/bluealsa-aplay.1.rst b/doc/bluealsa-aplay.1.rst
index 9541268..f01cb20 100644
--- a/doc/bluealsa-aplay.1.rst
+++ b/doc/bluealsa-aplay.1.rst
@@ -53,14 +53,8 @@ OPTIONS
 -D NAME, --pcm=NAME
     Select ALSA playback PCM device to use for audio output.
     The default is ``default``.
-
-    **bluealsa-aplay** does not perform any mixing of streams. If multiple devices
-    are connected it opens a new connection to the ALSA PCM device for each stream.
-    Therefore the PCM *NAME* must itself allow multiple open connections and
-    mix the streams together. See option **--single-audio** to change this
-    behavior. Similarly, **bluealsa-aplay** does not apply any
-    transformations to the stream. For this reason it is often necessary to use
-    the ALSA **dmix** and **plug** plugins in the *NAME* PCM.
+    See the section `Selecting An ALSA Playback PCM`_ below for more information
+    on ALSA playback PCM devices.
 
 --pcm-buffer-time=INT
     Set the playback PCM buffer duration time to *INT* microseconds.
@@ -132,6 +126,98 @@ OPTIONS
     PCM to be able to mix audio from multiple sources (i.e., it can be opened
     more than once; for example the ALSA **dmix** plugin).
 
+Selecting An ALSA Playback PCM
+==============================
+
+**bluealsa-aplay** does not apply any transformations to the audio stream.
+For this reason it is often necessary to use the ALSA **plug** plugin in the
+ALSA playback PCM.
+
+**bluealsa-aplay** does not perform any mixing of streams. If multiple bluetooth
+devices are connected it opens a new connection to the ALSA PCM device for each
+stream. Therefore the ALSA playback PCM must itself allow multiple open
+connections and mix the streams together (see option **--single-audio** to
+change this behavior). For this reason it is often necessary to use the ALSA
+**dmix** plugin in the ALSA playback PCM.
+
+For most distributions, the installed definition of ``default`` for each sound
+card includes both ``dmix`` and ``plug`` where necessary, so is generally the
+best choice unless there is some specific reason to prefer some other device.
+If there is more than one sound card attached then the appropriate one can be
+selected with:
+
+::
+
+    bluealsa-aplay -D default:CARDNAME
+
+A list of attached sound card names can be obtained using **aplay -l**, for
+example:
+
+::
+
+    $ aplay -l
+    **** List of PLAYBACK Hardware Devices ****
+    card 0: PCH [HDA Intel PCH], device 0: ALC236 Analog [ALC236 Analog]
+      Subdevices: 1/1
+      Subdevice #0: subdevice #0
+    card 1: Device [USB Audio Device], device 0: USB Audio [USB Audio]
+      Subdevices: 1/1
+      Subdevice #0: subdevice #0
+
+Here the first word after the card number is the card name, so that to use the
+HDA Intel PCH card:
+
+::
+
+    bluealsa-aplay -D default:PCH
+
+and to use the USB card:
+
+::
+
+    bluealsa-aplay -D default:Device
+
+
+Some sound cards offer more than one playback PCM device. ALSA identifies these
+devices by using an index number for each. Such a card might produce **aplay -l**
+output like:
+
+::
+
+    $ aplay -l
+    **** List of PLAYBACK Hardware Devices ****
+    card 0: PCH [HDA Intel PCH], device 0: ALC236 Analog [ALC236 Analog]
+      Subdevices: 1/1
+      Subdevice #0: subdevice #0
+    card 0: PCH [HDA Intel PCH], device 3: HDMI 0 [HDMI 0]
+      Subdevices: 1/1
+      Subdevice #0: subdevice #0
+    card 0: PCH [HDA Intel PCH], device 7: HDMI 1 [HDMI 1]
+      Subdevices: 1/1
+      Subdevice #0: subdevice #0
+
+Here the device index number is given by ``device N``, so in this example
+the Analog output is used with:
+
+::
+
+    bluealsa-aplay -D default:PCH,0
+
+and the first HDMI port (``HDMI 0``) is used with:
+
+::
+
+    bluealsa-aplay -D default:PCH,3
+
+If the device index is not specified, the default is ``0``.
+
+There will be some cases where the ``default`` device is not sufficient. These
+are generally special cases, such as selecting a specific subset of channels
+from a multi-channel device or to duplicate the stream across multiple output
+devices etc. These cases generally require some additional configuration by the
+user, and it is recommended to seek advice from your distribution as this can be
+quite complex and is beyond the scope of this manual.
+
 DMIX
 ====
 
@@ -168,7 +254,7 @@ Alternatively we can define a PCM with the required setting:
 SEE ALSO
 ========
 
-``bluealsa(8)``, ``bluealsa-rfcomm(1)``
+``amixer(1)``, ``aplay(1)``, ``bluealsa(8)``, ``bluealsa-rfcomm(1)``
 
 Project web site at https://github.com/Arkq/bluez-alsa
 ```

**Comment by graham8** [comment]
For a start, instead of "the default is default" suggest "if omitted, playback will be through your default audio device".  Then this is getting quite daunting.  I would put the examples first as the typical reader doesn't need to know the background if it just works with default:CARDNAME, then introduce the other options - subdevices first then the dmix issue.  I'm pretty sure default will always include a plug for samplerate conversion.  It does on my Debian-based distros (<confdir>/pcm/default.conf).

Next, on my systems, default doesn't take a DEV attribute so you can't say default:CARDNAME,DEV=1.

So something like this (with apologies for not formatting as a proper patch):

```diff
+Selecting An ALSA Playback PCM
+==============================
+
+If there is more than one sound card attached then the appropriate one can be
+selected with:
+
+::
+
+    bluealsa-aplay -D default:CARDNAME
+
+A list of attached sound card names can be obtained using **aplay -l**, for
+example:
+
+::
+
+    $ aplay -l
+    **** List of PLAYBACK Hardware Devices ****
+    card 0: PCH [HDA Intel PCH], device 0: ALC236 Analog [ALC236 Analog]
+      Subdevices: 1/1
+      Subdevice #0: subdevice #0
+    card 1: Device [USB Audio Device], device 0: USB Audio [USB Audio]
+      Subdevices: 1/1
+      Subdevice #0: subdevice #0
+
+Here the first word after the card number is the card name, so that to use the
+HDA Intel PCH card:
+
+::
+
+    bluealsa-aplay -D default:PCH
+
+or to use the USB card:
+
+::
+
+    bluealsa-aplay -D default:Device
+
+
+Some sound cards offer more than one playback PCM device. ALSA identifies these
+devices by using an index number for each. Such a card might produce **aplay -l**
+output like:
+
+::
+
+    $ aplay -l
+    **** List of PLAYBACK Hardware Devices ****
+    card 0: PCH [HDA Intel PCH], device 0: ALC236 Analog [ALC236 Analog]
+      Subdevices: 1/1
+      Subdevice #0: subdevice #0
+    card 0: PCH [HDA Intel PCH], device 3: HDMI 0 [HDMI 0]
+      Subdevices: 1/1
+      Subdevice #0: subdevice #0
+    card 0: PCH [HDA Intel PCH], device 7: HDMI 1 [HDMI 1]
+      Subdevices: 1/1
+      Subdevice #0: subdevice #0
+
+Here the device index number is given by ``device N``, so in this example
+the Analog output could be used with:
+
+::
+
+    bluealsa-aplay -D plughw:PCH,0
+
+and the first HDMI port (``HDMI 0``) is used with:
+
+::
+
+    bluealsa-aplay -D plughw:PCH,3
+
+For most distributions, the installed definition of ``default`` for most sound
+cards will automatically convert the samplerate and format of the BT audio to those
+supported by the playback card, using ALSA's  ``plughw`` plug-in.  ``default`` also refers
+to device 0 so we have to explicitly use the ``plughw`` plugin for other devices to get
+automatic samplerate and format conversion.
+
+The ``default`` for each sound card often also includes ALSA's ``dmix`` plug-in so that
+multiple sound sources (from **bluealsa-aplay** and/or other applications) can play
+simultaneously to the same card. It needs an additional configuration file to use both
+``dmix`` and ``plughw`` with a device other than 0.
+
+**bluealsa-aplay** does not perform any mixing of streams. If multiple bluetooth
+devices are connected it opens a new connection to the ALSA PCM device for each
+stream. Therefore the ALSA playback PCM must itself allow multiple open
+connections and mix the streams together (see option **--single-audio** to
+change this behavior). For this the ALSA ``dmix`` plugin must be used.
+
+If ``dmix`` is not provided by ``default`` in your distribution or if you have a special
+case, such as selecting a specific subset of channels from a multi-channel device or
+to duplicate the stream across multiple output devices etc. you will need some
+additional configuration. It is recommended to seek advice from your distribution
+as this can be quite complex and is beyond the scope of this manual.
```

borine:
> It needs an additional configuration file to use both ``dmix`` and ``plughw`` with a device other than 0.

@graham8 please can you explain that for me? I do not see that on a stock Ubuntu 22.04 (libasound 1.2.6). Here is an example where I force a audio format,rate and channel map that is not supported by my card and send this to device 3 using default:
```
$ aplay -v -D default:PCH,3 -t raw -c 4 -f u8 -r 96000 /dev/null
Playing raw data '/dev/null' : Unsigned 8 bit, Rate 96000 Hz, Channels 4
Plug PCM: Route conversion PCM (sformat=U8)
  Transformation table:
    0 <- 0
    1 <- 1
Its setup is:
  stream       : PLAYBACK
  access       : RW_INTERLEAVED
  format       : U8
  subformat    : STD
  channels     : 4
  rate         : 96000
  exact rate   : 96000 (96000/1)
  msbits       : 8
  buffer_size  : 32768
  period_size  : 2048
  period_time  : 21333
  tstamp_mode  : NONE
  tstamp_type  : MONOTONIC
  period_step  : 1
  avail_min    : 2048
  period_event : 0
  start_threshold  : 32768
  stop_threshold   : 32768
  silence_threshold: 0
  silence_size : 0
  boundary     : 4611686018427387904
Slave: Rate conversion PCM (48000, sformat=S32_LE)
Converter: libspeex (builtin)
Protocol version: 10002
Its setup is:
  stream       : PLAYBACK
  access       : MMAP_INTERLEAVED
  format       : U8
  subformat    : STD
  channels     : 2
  rate         : 96000
  exact rate   : 96000 (96000/1)
  msbits       : 8
  buffer_size  : 32768
  period_size  : 2048
  period_time  : 21333
  tstamp_mode  : NONE
  tstamp_type  : MONOTONIC
  period_step  : 1
  avail_min    : 2048
  period_event : 0
  start_threshold  : 32768
  stop_threshold   : 32768
  silence_threshold: 0
  silence_size : 0
  boundary     : 4611686018427387904
Slave: Soft volume PCM
Control: PCM Playback Volume
min_dB: -51
max_dB: 0
resolution: 256
Its setup is:
  stream       : PLAYBACK
  access       : MMAP_INTERLEAVED
  format       : S32_LE
  subformat    : STD
  channels     : 2
  rate         : 48000
  exact rate   : 48000 (48000/1)
  msbits       : 32
  buffer_size  : 16384
  period_size  : 1024
  period_time  : 21333
  tstamp_mode  : NONE
  tstamp_type  : MONOTONIC
  period_step  : 1
  avail_min    : 1024
  period_event : 0
  start_threshold  : 16384
  stop_threshold   : 16384
  silence_threshold: 0
  silence_size : 0
  boundary     : 4611686018427387904
Slave: Direct Stream Mixing PCM
Its setup is:
  stream       : PLAYBACK
  access       : MMAP_INTERLEAVED
  format       : S32_LE
  subformat    : STD
  channels     : 2
  rate         : 48000
  exact rate   : 48000 (48000/1)
  msbits       : 32
  buffer_size  : 16384
  period_size  : 1024
  period_time  : 21333
  tstamp_mode  : NONE
  tstamp_type  : MONOTONIC
  period_step  : 1
  avail_min    : 1024
  period_event : 0
  start_threshold  : 16384
  stop_threshold   : 16384
  silence_threshold: 0
  silence_size : 0
  boundary     : 4611686018427387904
Hardware PCM card 0 'HDA Intel PCH' device 0 subdevice 0
Its setup is:
  stream       : PLAYBACK
  access       : MMAP_INTERLEAVED
  format       : S32_LE
  subformat    : STD
  channels     : 2
  rate         : 48000
  exact rate   : 48000 (48000/1)
  msbits       : 32
  buffer_size  : 16384
  period_size  : 1024
  period_time  : 21333
  tstamp_mode  : ENABLE
  tstamp_type  : MONOTONIC
  period_step  : 1
  avail_min    : 1024
  period_event : 0
  start_threshold  : 1
  stop_threshold   : 4611686018427387904
  silence_threshold: 0
  silence_size : 4611686018427387904
  boundary     : 4611686018427387904
  appl_ptr     : 0
  hw_ptr       : 0
```
As you can see, `default` has automatically brought in all the following plugins without any configuration from me:
- route
- rate
- softvol
- dmix
- hw

> default doesn't take a DEV attribute

Yes, my bad. When a card has only one device then its `default` definition does not include the DEV argument. So the doc should still make it clear that the device index is for use only on multi-device cards and that  leaving out the index on those cards result in device 0 being chosen.

graham8:
I've decided my ALSA blog will be called ALSA in Wonderland in honour of the infinite rabbit holes we have to travel.

Arkq:
I've pushed proposed changes to PR https://github.com/Arkq/bluez-alsa/pull/572 Maybe it will be easier to work on it if the man will be rendered as a whole (e.g. using "add a suggestion"). Discussion can still happen here, though.

My notes:
I think that the info about **bluealsa-aplay** not performing any mixing and transformation should be still in the `--pcm` option. It's kind of essential info for anyone who is familiar with ALSA (and I think that this manual should be addressed mostly to such audience). Maybe it can be redacted as a brief note. Later in the examples or "Selecting An ALSA Playback PCM" section (I'm still in favor of "EXAMPLES", the other name fits more for a blog article which @graham8 is trying to write :)) we can say more about mixing and transformations and how it can be handled with ALSA plugins.

borine:
@Arkq, @graham8  All the text concerning device index with `default` in my patch is wrong. As @graham8 has pointed out, out-of-the-box `default` does ***not*** have a second argument, so `default:PCH,0` is not valid. Apologies again for this mistake.

Arkq:
I've updated manual page with the EXAMPLES section which contains a hint about the `aplay`/`amixer` and how to use it to get the device names. I hope that will point someone who is not familiar with ALSA in the right direction.

PR #579, opened 2022-07-09
Make default description more general

graham8:
We are updating the bluez-alsa we ship with OSMC and would like to use your 20-bluealsa.conf without patching.  Confused by the description, though. What is the thinking behind "hub"?

### DIFF
--- a/src/asound/20-bluealsa.conf
+++ b/src/asound/20-bluealsa.conf
@@ -133,6 +133,6 @@ pcm.bluealsa {
 			@func refer
 			name defaults.namehint.extended
 		}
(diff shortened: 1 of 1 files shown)


Arkq:
The resoning behind the "hub" was the analogue of USB-hub when a single "controller" has multiple number of USB devices. Similar here where a single mixer control controls not a single sound card but all connected BT devices. But if that's confusing I'm OK to change that.

Arkq (approved):

borine:
Yes, I agree this is an improvement. The hint description applies to a single PCM, not to the bluealsa service, so in that sense it is not a "hub"

PR #593, opened 2022-09-15
Add support for more recent spandsp versions

Batchyx:
The spandsp at https://www.soft-switch.org/ is old and not updated. A fork exist at https://github.com/freeswitch/spandsp and it turns out the old maintainer (Steve Underwood, https://github.com/coppice-git/) contributed to it until 2018:
https://github.com/freeswitch/spandsp/commits?author=coppice-git https://github.com/freeswitch/spandsp/commit/31504fd0c1ab968673234db4cc.patch

Trying to use this version currently fails, because plc_state_t is now a private structure which must be dynamically allocated:

In file included from codec-msbc.c:12:
codec-msbc.h:68:21: error: field 'plc' has incomplete type
   68 |         plc_state_t plc;
      |                     ^~~

Switching to plc_alloc()/plc_free() fixes the problem without breaking compatibility with the old version at www.soft-switch.org

### DIFF
--- a/src/codec-msbc.c
+++ b/src/codec-msbc.c
@@ -114,10 +114,10 @@ int msbc_init(struct esco_msbc *msbc) {
 	msbc->seq_number = 0;
 	msbc->frames = 0;
 
-	/* Initialize PLC context. When calling with non-NULL parameter,
-	 * this function does not allocate anything - there is no need
-	 * to call plc_free(). */
-	plc_init(&msbc->plc);
+	if (msbc->plc)
+		plc_init(msbc->plc);
+	else if (!(msbc->plc = plc_init(NULL)))
+		goto fail;
 
 	msbc->initialized = true;
 	return 0;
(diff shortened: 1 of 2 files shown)


src/codec-msbc.c
```diff
@@ -114,10 +114,7 @@ int msbc_init(struct esco_msbc *msbc) {
 	msbc->seq_number = 0;
 	msbc->frames = 0;
 
-	/* Initialize PLC context. When calling with non-NULL parameter,
-	 * this function does not allocate anything - there is no need
-	 * to call plc_free(). */
-	plc_init(&msbc->plc);
+	msbc->plc = plc_init(NULL);
```
Arkq, line 117: The `msbc_init` might be called more than once without calling `msbc_finish`. In order not to leak memory you have to free plc context if it is not NULL before allocating new one.

Arkq (changes requested):

src/codec-msbc.c
Batchyx: I'll just use `msbc_init` on the already-allocated pointer, even if it theory, `msbc_release` should probably be used (it does not do anything).

Arkq (approved):

Issue #536, opened 2022-02-15
E: Invalid BT socket & W: Couldn't register oFono


dr-ni:
- Are these warning relevant?
- Why are they always arising?

I have still very long delay for sent out audio on Raspberry zero W (buster) and Raspberry zero 2 (bullseye) with ofono see also #385
Could this be the reason?

raspberry zero 2 W:
```
root@rpi:/home/pi# bluealsa -p hfp-ofono
bluealsa: I: main.c:479: Disabling native HFP support due to enabled oFono profile
bluealsa: D: main.c:534: Starting main dispatching loop
bluealsa: D: main.c:111: Acquired D-Bus service name: org.bluealsa
bluealsa: D: bluealsa-dbus.c:426: Registering D-Bus manager: /org/bluealsa
bluealsa: D: bluez.c:669: Registering battery provider: /org/bluez/hci0/battery
bluealsa: D: bluez.c:689: BlueZ battery provider support not available
bluealsa: D: bluez.c:483: Creating media endpoint object: /org/bluez/hci0/A2DP/SBC/source/1
bluealsa: D: bluez.c:392: Registering media endpoint: /org/bluez/hci0/A2DP/SBC/source/1
bluealsa: D: bluez.c:483: Creating media endpoint object: /org/bluez/hci0/A2DP/SBC/source/2
bluealsa: D: bluez.c:392: Registering media endpoint: /org/bluez/hci0/A2DP/SBC/source/2
bluealsa: D: bluez.c:918: Creating hands-free profile object: /org/bluez/HSP/AudioGateway
bluealsa: D: bluez.c:842: Registering hands-free profile: /org/bluez/HSP/AudioGateway
bluealsa: D: ofono.c:466: Registering oFono audio agent: /org/bluez/HFP/oFono
bluealsa: D: ofono.c:247: Adding new oFono card: /card_1
bluealsa: D: ba-transport.c:1151: Starting transport: HFP Hands-Free (CVSD)
bluealsa: E: ba-transport.c:278: Invalid BT socket: -1
bluealsa: E: ba-transport.c:278: Invalid BT socket: -1
bluealsa: D: ofono.c:466: Registering oFono audio agent: /org/bluez/HFP/oFono
bluealsa: W: ofono.c:521: Couldn't register oFono: GDBus.Error:org.ofono.Error.InUse: The resource is currently in use
```
raspberry zero w:
```
[...]

PR #601, opened 2022-11-17
PCM "Running" property

borine:
This is a tentative proposal to implement the suggestion here:
https://github.com/Arkq/bluez-alsa/issues/536#issuecomment-1280767923

I've called the new PCM1 property "Running", but of course that can easily be changed.

I should say here that I only have a little previous experience coding the D-Bus and GLib APIs, so there may well be better ways of implementing this. However, I have found the new bluealsa-cli monitor functionality to be very useful in testing; I got good results using it in a script to launch `alsaloop` on a HF node as soon as the audio connection is established.

### DIFF
--- a/src/ba-transport.c
+++ b/src/ba-transport.c
@@ -99,6 +99,7 @@ static int transport_pcm_init(
 
 	pcm->t = t;
 	pcm->th = th;
+	pcm->th->pcm = pcm;
 	pcm->mode = mode;
 	pcm->fd = -1;
 	pcm->active = true;
@@ -215,6 +216,7 @@ static int transport_thread_init(
 
 	th->t = t;
 	th->state = BA_TRANSPORT_THREAD_STATE_TERMINATED;
+	th->pcm = NULL;
 	th->bt_fd = -1;
 	th->pipe[0] = -1;
 	th->pipe[1] = -1;
@@ -321,6 +323,8 @@ int ba_transport_thread_state_set(
 
 	pthread_mutex_lock(&th->mutex);
 
+	enum ba_transport_thread_state old_state = th->state;
+
 	/* Moving to the next state is always allowed. */
 	bool valid = state == th->state + 1;
 
@@ -351,6 +355,13 @@ int ba_transport_thread_state_set(
 		return errno = EINVAL, -1;
 
 	pthread_cond_broadcast(&th->cond);
+
+	if (state != old_state && th->pcm != NULL && th->pcm->ba_dbus_exported) {
+		if (old_state == BA_TRANSPORT_THREAD_STATE_RUNNING ||
+				state == BA_TRANSPORT_THREAD_STATE_RUNNING) {
+			bluealsa_dbus_pcm_update(th->pcm, BA_DBUS_PCM_UPDATE_RUNNING);
+		}
+	}
 	return 0;
 }
 
--- a/src/ba-transport.c
+++ b/src/ba-transport.c
@@ -218,7 +218,6 @@ static int transport_thread_init(
 
 	th->t = t;
 	th->state = BA_TRANSPORT_THREAD_STATE_TERMINATED;
-	th->pcm = NULL;
 	th->bt_fd = -1;
 	th->pipe[0] = -1;
 	th->pipe[1] = -1;
--- a/doc/bluealsa-api.txt
+++ b/doc/bluealsa-api.txt
@@ -140,6 +140,11 @@ Properties      object Device [readonly]
                         Possible A2DP values: 0-127
                         Possible SCO values: 0-15
 
+               boolean Running [readonly]
+
+                        This property is true when the Bluetooth transport for
+                        this PCM is acquired and able to transfer audio samples.
+
 RFCOMM hierarchy
 ================
 
--- a/doc/bluealsa-cli.1.rst
+++ b/doc/bluealsa-cli.1.rst
@@ -178,8 +178,8 @@ monitor [-p[PROPS] | --properties[=PROPS]]
 
     ``PropertyChanged PCM_PATH PROPERTY_NAME VALUE``
 
-    Property names than can be monitored are **Codec**, **SoftVolume** and
-    **Volume**.
+    Property names than can be monitored are **Codec**, **Running**,
+    **SoftVolume** and **Volume**.
 
     The value for Volume is a hexadecimal 16-bit encoding where data for
     channel 1 is stored in the upper byte, channel 2 is stored in the lower
--- a/misc/bash-completion/bluealsa
+++ b/misc/bash-completion/bluealsa
@@ -61,6 +61,7 @@ _bluealsa_pcm_codecs() {
 _bluealsa_cli_properties() {
 	local properties=(
 		Codec
+		Running
 		SoftVolume
 		Volume
 	)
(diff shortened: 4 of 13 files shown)


src/ba-transport.c
```diff
@@ -267,8 +276,15 @@ int ba_transport_thread_set_state(
 
 	pthread_mutex_unlock(&th->state_mtx);
 
-	if (!skip)
+	if (!skip) {
 		pthread_cond_signal(&th->changed);
+		if (state != old_state) {
+			if (old_state == BA_TRANSPORT_THREAD_STATE_RUNNING ||
+					state == BA_TRANSPORT_THREAD_STATE_RUNNING) {
+				eventfd_write(th->state_fd, 1);
```
Arkq, line 284: Why you are starting IO watch for property change notification? I think that it will be easier to call `bluealsa_dbus_pcm_update()` directly from here (with some "ifs" for checking which PCM is associated with the transport thread).

borine: I wasn't sure about libdus thread management internals, so I wanted this dbus blockng call to be made from the glib mainloop thread. However, if if is safe to call this from transport IO threads then that will simplify things a lot and I will make that change.

Arkq: Yes, it should be safe to call it from here, GIO is thread safe (exception from general rule in glib, that glib APIs are not thread-safe).

And one more thing. If this PR is supposed to be tested by someone else, I think that it will be better to base in on 1326cb1 (not on master). After commit c0b1d29, there is a rarely occurring dead-lock when opening PCM (added mutex when accessing pcm->fd in io.c). Right now I'm working on improvements for thread/mutex management and more patches are coming (it's quite possible that I will abandon pthraed cancellation.... - it seems that there is a bug in libc: dead-lock in the scenario like this: cancel signal -> running some cancellation internals (lock mutex) -> another signal (e.g. SIGTERM) -> signal handler runs libc write -> running [again] cancellation internals -> dead lock on mutex lock). If testing this PR is not critical, you can wait for a proper fix in master (I will push it in few days) - the fix will add dedicated mutex for pcm client in the `bluealsa_pcm_open()`, but I'm still testing it though.

borine: OK, I will remove the state_fd eventfd and update the D-Bus state from the IO threads. As far as testing goes, I have enough feedback already from my own tests and #536 to be confident that is is a worth while addition, so I will stop trying to keep it based on master and wait for your current re-design to be completed before re-basing.

Arkq: I've just pushed the commit with dedicated client mutex lock. I hope that CI will not fail so often any more (it still can fail because of mentioned libc signal issue, but it's manifesting itself only during bluealsa termination with SIGTERM/SIGINT).

Arkq: Hmm... I have not thought about that, but yes, you might be right. I will have to check that. Thx!

Arkq: With your suggestion (masking all signals) I've been running `test_playback_hw_set_free` test in a loop for ~12 hours and no deadlock has accrued. Without masking, I've got deadlock within ~30 minutes (running on RPi4 Debian 11.5). Once again thanks for suggestion!

borine:
> Can you look at the script and try to adopt it to new behavior? 

OK, I'll play around with that and get back to you. Not sure how long it will take, it will either be in next couple of days or not for a couple of weeks (I am busy this weekend and the following week).

borine:
The completion script for `bluealsa-cli` will have to be re-written, so that is unlikely to be ready until the New Year. There is one notable inconsistency in the global options: `--dus=` has to be positioned *before* the command, whereas `--quiet` and `--verbose` my be placed either before or after the command. Is that intentional?

Arkq:
Yes, that was intentional (but I'm not sure whether that's right decision :D). The case is that the D-Bus connection is initialized before launching commands, so I was truing to avoid move that piece of code into all subcommands. Adding verbose and quiet was cheaper. But as I think of that right now, maybe the `-v` and `-q` should not be added to subcommands.... For sure I want to keep dedicated `--help`.

borine:
The completion re-write proved much easier than expected. So (early Christmas present) I've created a new PR with a first draft. It's obviously had very little testing, but hopefully the basics are OK. See PR 604

EDIT oops - just seen your edit. Will update my patch tomorrow

src/ba-transport.c
```diff
@@ -1638,8 +1652,12 @@ void ba_transport_thread_cleanup(struct ba_transport_thread *th) {
 	ba_transport_thread_bt_release(th);
 
 	/* If we are closing master thread, release underlying BT transport. */
-	if (th->master)
+	if (th->master) {
 		ba_transport_release(t);
+		/* SCO encoder thread on HF cannot transfer audio when BT socket is closed */
+		if (t->profile & BA_TRANSPORT_PROFILE_MASK_HF)
+			ba_transport_thread_set_state(&t->thread_enc, BA_TRANSPORT_THREAD_STATE_NONE);
```
arkq, line 1659: Do you have any test case where this is required?

Currently I'm trying to "fix" IO threads start and termination in case of errors and this patch seems to be related :D For full duplex codecs (e.g. SCO, A2DP FastStream) we need both IO threads running, if one fails other should terminated and PCM client notified.

I will go back to this PR after IO threads fix, because maybe it will be more feasible to change property type from boolean to state/enum after all my changes.

borine: > Do you have any test case where this is required?

No, it appears to be a relic of some idea I was experimenting with, that was not cleaned up before I submitted the PR. As far as I can tell now, it is not making any difference to the propertieschanged signals.

There is an "annoyance" with HF/HS nodes that the sink PCM remains with "Running true" when the transport is released by the AG, then on the next acquisition it emits "Running false" followed by "Running true". This makes this signal difficult to use for starting/stopping clients on the HF/HS. I think I was trying to persuade bluealsa to emit a property changed signal even though the encoder thread is still running, but never achieved that successfully.

arkq (approved):

arkq:
I've merged it with slightly reworked fix for `test_capture_poll` test ca. Now it will not skip the "block forever" check.

borine:
That's great. Thank you so much for including this feature. I have a collection of small changes relating to HFP that fix minor issues I found while working on this, and will will raise a new PR with those once they are re-based. Then I'll update the wiki with some hints on using this new feature, with an example shell script that creates a simple audio connection helper using `aplay`, `arecord` and `bluealsa-cli`,  for use with either oFono or native BlueALSA based HFP-HF systems.

arkq:
> Thank you so much for including this feature.

No problem. For a while I was thinking about more generic "state" property, but I've got some troubles how to "design" it properly... so running is good enough.

Also, during tests I've also discovered some issues (I'm not sure whether theses issues are related to `bluealsa` or maybe only to the mock server). The problem is with "blocking forever case". When capture PCM is opened but the PIPE is not drained, the buffer will fill up and the `write()` (in bluealsa) will block. Later when ALSA plug-in wants to start pcm, it calls pcm resume which needs pcm mutex (in bluealsa), but the mutex is locked by the io_pcm_write and won't be released because `write()` blocks.... That's one case. The other case is with state cancellation logic. There is still possible to have more than one MANAGER_CANCEL_THREAD queued in the manager - when send command is more than one thread. And it pops up very frequently when running `test-io`. I think that I will have to redesign send_command() slightly... the simplest way would be to send unique thread counter and compare it the manager thread whether it has not changed...

PR #604, opened 2022-12-08
Cli monitor props

borine:


borine:
I'm now thinking that having quiet and verbose options after the command is very convenient for the user, so maybe its just the documentation that needs a bit more text to help explain the inconsistency? Anyway, have a think about it and I will fix up the completion script to match whatever you decide.

Arkq:
Nice. Thanks for that!

I was also looking at the autocompletion script, but I was wondering whether it would be possible to parse subcommand in the same way the command is parsed - somehow parse `bluealsa-cli COMMAND --help`. But bash-commpletion specific variables are a pure magic for me :D so I'm not sure whether it's even possible to have thigs like "bluelsa-cli --option1 --option2 COMMAND --some-option *\<TAB\>*" and autocompletion script will know what kind of argument is that and autocomplete it, e.g. `PCM-PATH`

Arkq:
> I'm now thinking that having quiet and verbose options after the command is very convenient for the user

Yes, that was my reasoning. Especially for adding `-v` to the command. But after your commend I'm not longer sure :D Maybe it could be changed later. I think that for now we are good with `--help` only (at least it's not a duplication for "global" `--help`.

I was using Python https://click.palletsprojects.com/en/8.1.x/ framework as a template for how to present subcommands and their options to the user.

PR #606, opened 2022-12-19
Alsa drain

borine:
1. Avoid potential deadlock in ALSA PCM plugin: close() is a thread cancellation point so should not be called while the mutex is locked.
2. Avoid checking the value of pcm->ba_pcm_fd in the application thread while the io thread is running.
3. Allow playing of audio files that are shorter than the start_threshold parameter.

Together these changes permit the use of BlueALSA with alsa-utils `speaker-test`.

### DIFF
--- a/src/asound/bluealsa-pcm.c
+++ b/src/asound/bluealsa-pcm.c
@@ -250,8 +250,6 @@ static void *io_thread(snd_pcm_ioplug_t *io) {
 
 			if (pcm->io_hw_ptr == -1)
 				continue;
-			if (pcm->ba_pcm_fd == -1)
-				goto fail;
 
 			asrsync_init(&asrs, io->rate);
 			io_hw_ptr = pcm->io_hw_ptr;
@@ -304,14 +302,17 @@ static void *io_thread(snd_pcm_ioplug_t *io) {
 					if (errno == EINTR)
 						continue;
 					SNDERR("PCM FIFO read error: %s", strerror(errno));
+					pcm->connected = false;
 					goto fail;
 				}
 				head += ret;
 				len -= ret;
 			}
 
-			if (ret == 0)
+			if (ret == 0) {
+				pcm->connected = false;
 				goto fail;
+			}
 
 			io_thread_update_delay(pcm, io_hw_ptr);
 
@@ -325,6 +326,7 @@ static void *io_thread(snd_pcm_ioplug_t *io) {
 						continue;
 					if (errno != EPIPE)
 						SNDERR("PCM FIFO write error: %s", strerror(errno));
+					pcm->connected = false;
 					goto fail;
 				}
 				head += ret;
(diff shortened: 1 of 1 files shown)


src/asound/bluealsa-pcm.c
```diff
@@ -345,10 +348,19 @@ static void *io_thread(snd_pcm_ioplug_t *io) {
...
+	}
+	if (pcm->ba_pcm_ctrl_fd != -1) {
+		close(pcm->ba_pcm_ctrl_fd);
+		pcm->ba_pcm_ctrl_fd = -1;
+	}
+
+	/* make sure we will not get stuck in the pause sync loop */
 	pthread_mutex_lock(&pcm->mutex);
-	pcm->pause_state = BA_PAUSE_STATE_PAUSED;
+	pcm->connected = false;
```
arkq, line 362: I would keep the `pcm->pause_state = BA_PAUSE_STATE_PAUSED` like it was before. I think that the `pause_state` is more related to the `pause_cond` than the `connected` fag. But since there is now a `connected` flag I would set it to false just after "disconnected" notification from `read()`/`write()` - just before `goto fail` in lines 307, 314 and 328. In that way it will be easier to read/understand the flow IMHO.

src/asound/bluealsa-pcm.c
```diff
@@ -345,10 +348,19 @@ static void *io_thread(snd_pcm_ioplug_t *io) {
 
 fail:
 
+	if (pcm->ba_pcm_fd != -1) {
```
arkq, line 351: Is it required to close file descriptors here? With the `connected` flag, we can relay on the `bluealsa_hw_free()` being called by the ALSA library - right now there is no function which checks for `ba_pcm_fd == -1`.

src/asound/bluealsa-pcm.c
```diff
@@ -770,11 +831,17 @@ static int bluealsa_pause(snd_pcm_ioplug_t *io, int enable) {
...
 		pthread_mutex_lock(&pcm->mutex);
 		pcm->pause_state |= BA_PAUSE_STATE_PENDING;
-		while (!(pcm->pause_state & BA_PAUSE_STATE_PAUSED) && pcm->ba_pcm_fd != -1)
+		while (!(pcm->pause_state & BA_PAUSE_STATE_PAUSED) && pcm->connected)
 			pthread_cond_wait(&pcm->pause_cond, &pcm->mutex);
 		pthread_mutex_unlock(&pcm->mutex);
 	}
 
+	if (!pcm->connected) {
+		pcm->pause_state = BA_PAUSE_STATE_RUNNING;
```
arkq, line 840: Do we need to set the `pause_state` in case of error? If so, it will have to be done under mutex, otherwise TSAN might report race warning (this flag is set in the IO thread as well).

arkq (commented):
Nice improvement!

arkq (approved):

PR #608, opened 2022-12-29
updates to manual pages

borine:


arkq (approved):

arkq:
> During this waiting time the PCM plugin behaves as if the device "clock" is stopped, it does not generate any poll() events, and the application will be blocked when writing or reading to/from the PCM.

Apart from this PR, I've got one idea. Maybe, we should generate silence in case when BT transport is not acquired yet (for source PCMs)? This of course can be an option for PCM with some sane default (I'm not such whether by default PCM should generate silence or not). I'm not sure, but maybe there is a similar case with some hardware ALSA PCM... so BlueALSA should behave in similar way (or at least ALSA PCM plug-in).

borine:
> Maybe, we should generate silence 

Actually, I am already doing that in my own fork. There is a branch https://github.com/borine/bluez-alsa/tree/alsa-capture which has just that one change. As you can see, my solution turns out to be quite "ugly" in many ways, but at least the code is doing what I need it to do. I can raise a PR from it if you think it may be workable as a general solution.

arkq:
> As you can see, my solution turns out to be quite "ugly" in many ways, but at least the code is doing what I need it to do

OK, I see. I will think about it. Maybe the solution could be simplified if implemented in the server.... or maybe server can support it somehow, so the client code can be simpler. However, one think I'd like to avoid, if I'm not mistaken. I haven't read your code very carefully, but it looks to me that you are outputting silence not only during the initial not-acquired state, but also later. I think that the PLC code in the server should be responsible for inserting e.g. silence in places of missing packets. The case is that Bluetooth connection might be unstable, the delay might accumulate, but eventually all packets will be delivered. So, received audio will be complete (if written to file). Using precise timing in the ALSA plugin layer might insert silence in places where it should not be inserted...

arkq (approved):

PR #616, opened 2023-02-22
Do not allow a slow client to block the decoder thread

borine:
Here's my rather simple solution to the slow client / full pipe problem mentioned in https://github.com/arkq/bluez-alsa/pull/601#issuecomment-1436569189


arkq:
Hmm... simple yet correct, I think :D

arkq (approved):

PR #618, opened 2023-02-26
fix IO test

borine:
Fix for random test suite io_test failures.

### DIFF
--- a/test/test-io.c
+++ b/test/test-io.c
@@ -596,6 +596,7 @@ static void test_io(struct ba_transport *t_src, struct ba_transport *t_snk,
 
 	ba_transport_thread_state_set_stopping(&t_src->thread_enc);
 	transport_thread_cancel(&t_src->thread_enc);
+	ba_transport_thread_state_wait_terminated(&t_src->thread_enc);
 
 	pthread_mutex_lock(&t_snk_pcm->mutex);
 	ba_transport_pcm_release(t_snk_pcm);
@@ -603,6 +604,7 @@ static void test_io(struct ba_transport *t_src, struct ba_transport *t_snk,
 
 	ba_transport_thread_state_set_stopping(&t_snk->thread_dec);
 	transport_thread_cancel(&t_snk->thread_dec);
+	ba_transport_thread_state_wait_terminated(&t_snk->thread_dec);
 
 }
 


arkq:
The `transport_thread_cancel` is suppose to be synchronous, so it actually waits for thread to be terminated :)

I have not checked this thoroughly, but I suspect that there is a case of "double" cancellation. However, my first guess was exactly like yours, that the test does not wait between stop and start.

Anyway, regardless of that fix, I've seen that you've been working on HFP improvements. If you are OK with that I'd like to cherry-pick some commits, in particular (I have not tested it by myself, I've only seen commit titles, though):
- Pre-select CVSD when mSBC is disabled on the command-line
- Only feature HFP codec negotiation if mSBC is enabled
- Enable oFono AG support for mSBC (this one I will have to review, but the commit title looks like a nice improvement)
- Fix SCO volume control
- ALSA PCM plugin: fail immediately when codec not known
- Improve HFP-HF to support gateways that use oFono

Also, if possible, I will add some test cases for these fixes/improvements.

borine:
I think the HFP changes are ready for PR now. The test suite failure made me think there might be some issue in them, so I decided to investigate that first. I am now convinced that the problem is in the test code, so I will raise the HFP PR for proper discussion soon.

Sorry for raising this one too soon - I had no failures on my own machine, and none from two runs in my github fork, but as soon as I created the PR it failed!. Mildly embarrassing for me, but no harm done I hope.

arkq:
>  I am now convinced that the problem is in the test code

Hmm.... OK. I have had not enough spare time to investigate that, though...

> Sorry for raising this one too soon - I had no failures on my own machine, and none from two runs in my github fork, but as soon as I created the PR it failed!.

Yeap, that happens to me often. On my local machine all tests are passing, but on CI 2 out of 8 tests fail from time to time. However, when I run test in a loop like this: `while ./test/test-io; do true; done` I hit fail after few minutes (it depends on the current load on the machine).

arkq:
I think that I've finally fixed the IO cancellation - no more random CI failures :D. The fix is here: https://github.com/arkq/bluez-alsa/commit/2f874bbf9a9b64c5c79e5b607d3daaeb7327ad5f

However, during tests I've discovered nasty bug introduced in 07701805607a191cdf20e3f4d765e9a2d6289fa4 :/ On slow machine the A2DP source profile might not start because transport state change to "active" might be delivered by D-Bus before `ba_transport_acquire()` sets IO threads state to "idle". And the problem is that `ba_transport_start()` checks whether IO threads are idle... I hope that this will be the LAST issue with IO threads...

PR #619, opened 2023-02-27
HFP fixes

borine:
A collection of changes for HFP, some minor bug fixes, some enhancements. The two oFono related commits need some carefult consideration.

"Improve HFP-HF to support gateways that use oFono" is a work-around for the way that oFono interprets the bluetooth HFP codec connection specification. oFono takes it to mean that the SCO connection must be completed *before* the codec connection is considered complete. It seems neither Android nor IOS have this interpretation, and nor does BueALSA. This commit attempts to allow for BlueALSA to work with both interpretations.

 "Enable oFono AG support for mSBC" is a work-around for the way oFono does codec connection when in AG mode. Its worth noting here that oFono has a bug that causes it to enter a busy loop (100%CPU) when a bluetooth device is disconnected. So to test this it is best to patch the oFono source first. Here is my fix:
```diff
diff --git a/plugins/hfp_ag_bluez5.c b/plugins/hfp_ag_bluez5.c
index 15acf413..e981a9fb 100644
--- a/plugins/hfp_ag_bluez5.c
+++ b/plugins/hfp_ag_bluez5.c
@@ -249,7 +249,7 @@ static DBusMessage *profile_new_connection(DBusConnection *conn,
 
 	fd_dup = dup(fd);
 	io = g_io_channel_unix_new(fd_dup);
-	g_io_add_watch_full(io, G_PRIORITY_DEFAULT, G_IO_HUP, io_hup_cb,
+	g_io_add_watch_full(io, G_PRIORITY_DEFAULT, G_IO_HUP|G_IO_ERR|G_IO_NVAL, io_hup_cb,
 						g_strdup(device), g_free);
 	g_io_channel_unref(io);
 
```

### DIFF
--- a/src/ofono.c
+++ b/src/ofono.c
@@ -20,7 +20,6 @@
 /* IWYU pragma: no_include "config.h" */
 
 #include <errno.h>
-#include <poll.h>
 #include <pthread.h>
 #include <stdbool.h>
 #include <stdint.h>
@@ -47,6 +46,7 @@
 #include "bluealsa-config.h"
 #include "dbus.h"
 #include "hci.h"
+#include "hfp.h"
 #include "ofono-iface.h"
 #include "ofono-skeleton.h"
 #include "shared/log.h"
@@ -67,28 +67,6 @@ struct ofono_card_data {
 static GHashTable *ofono_card_data_map = NULL;
 static const char *dbus_agent_object_path = "/org/bluez/HFP/oFono";
 static ofono_HFAudioAgentIfaceSkeleton *dbus_hf_agent = NULL;
-
-/**
- * Authorize oFono SCO connection.
- *
- * @param fd SCO socket file descriptor.
- * @return On success this function returns 0. Otherwise, -1 is returned and
- *	 errno is set to indicate the error. */
-static int ofono_sco_socket_authorize(int fd) {
-
-	struct pollfd pfd = { fd, POLLOUT, 0 };
-	char c;
-
-	if (poll(&pfd, 1, 0) == -1)
-		return -1;
-
-	/* If socket is not writable, it means that it is in the defer setup
-	 * state, so it needs to be read to authorize the connection. */
-	if (!(pfd.revents & POLLOUT) && read(fd, &c, 1) == -1)
-		return -1;
-
-	return 0;
-}
 
 /**
  * Ask oFono to connect to a card. */
@@ -132,6 +110,21 @@ static int ofono_acquire_bt_sco(struct ba_transport *t) {
 	GUnixFDList *fd_list = g_dbus_message_get_unix_fd_list(rep);
 	if ((fd = g_unix_fd_list_get(fd_list, 0, &err)) == -1)
 		goto fail;
+
+#if ENABLE_MSBC
+	if (codec != ba_transport_get_codec(t)) {
+		/* Although this connection has succeeded, it is not the codec expected
+		 * by the client. So we have to return an error ... */
+		error("Rejecting oFono SCO link: %s", "Codec mismatch");
+		/* ... but still update the codec ready for next client request. */
+		ba_transport_set_codec(t, codec);
+		if (fd != -1) {
+			shutdown(fd, SHUT_RDWR);
+			close(fd);
+		}
+		goto fail;
+	}
+#endif
 
 	t->bt_fd = fd;
 	t->mtu_read = t->mtu_write = hci_sco_get_mtu(fd, t->d->a->hci.type);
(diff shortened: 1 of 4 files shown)


arkq:
@borine I've got a question regarding "Fix SCO volume control". Are you sure that current master's got it wrong? I've just checked `-p hsp-hs` with an Android phone, and it seems to work correctly. I can see speaker volume changes in alsamixer, and changes made in alsamixer are reflected on phone's volume control (there is no mic control, though, as this feature is not very common).

src/ba-rfcomm.c
```diff
@@ -1283,11 +1303,19 @@ static void *rfcomm_thread(struct ba_rfcomm *r) {
 			switch (rfcomm_recv_signal(r)) {
 #if ENABLE_MSBC
 			case BA_RFCOMM_SIGNAL_HFP_SET_CODEC_CVSD:
+				if ((r->sco->profile & BA_TRANSPORT_PROFILE_HFP_AG) && !(r->hfp_features & HFP_HF_FEAT_CODEC)) {
```
arkq, line 1306: I'm trying to test "Fix HFP codec selection logic" commit, and I wonder why the `rfcomm_finalize_codec_selection()` is sent only in case of AG here? I've just looked through all the "if" branches, in here in order to fix potential deadlocks due to error being returned by `rfcomm_set_hfp_codec()`, and I'm kind of inclined to loosen the condition here to just `if (!(r->hfp_features & HFP_HF_FEAT_CODEC))`. For codec negotiation all involved parties (AG and HF) have to support it. So it seems reasonable to skip sending `+BCS` command if either of them does not support `FEAT_CODEC`.

arkq: After some tests I've decided to check `FEAT_CODEC` for both sides, i.e. AG and HF.

borine:
I've rebased against master, and added a new option for bluealsa-rfcomm to print the properties. 

In particular, the list of supported features is useful to know in a script, and now that can be got without needing to open a rfcomm terminal and exchange AT messages.

arkq:
I've just pushed a *little* refactor of  your "Include mSBC in HF supported codecs" commit. I thought that maybe it will be possible to guess a little bit better wheter mSBC codec is available or not :D I've tested it with various scenarios and despite one or two corner cases, it shows mSBC codec only when it is indeed supported.

src/asound/bluealsa-pcm.c
```diff
@@ -564,8 +564,11 @@ static int bluealsa_hw_params(snd_pcm_ioplug_t *io, snd_pcm_hw_params_t *params)
 	if (!bluealsa_dbus_pcm_open(&pcm->dbus_ctx, pcm->ba_pcm.pcm_path,
 				&pcm->ba_pcm_fd, &pcm->ba_pcm_ctrl_fd, &err)) {
 		debug2("Couldn't open PCM: %s", err.message);
+		ret = -EIO;
+		if (strcmp(err.message, strerror(EBUSY)) == 0)
```
arkq, line 568: I'm not sure whether string comparison will work here. The problem is that bluealsa daemon and ALSA application can run under different locales, so the strings will not match. I think that we need error code to errno mapping (and change error codes in `bluealsa_pcm_open()`.

arkq (changes requested):

src/ofono.c
```diff
@@ -632,15 +732,36 @@ static void ofono_signal_card_removed(GDBusConnection *conn, const char *sender,
...
 
+fail:
+	if (a != NULL)
+		ba_adapter_unref(a);
+	if (d != NULL)
+		ba_device_unref(d);
+	if (t != NULL)
+		ba_transport_unref(t);
+
+	g_hash_table_remove(ofono_card_data_map, ocd);
```
arkq, line 764: I'm not sure whether this is OK. The `g_hash_table_remove` removes element by the key, but the `ocd` is not a key but a value. Instead of duplicating `ofono_transport_lookup` logic here, I would simply call `g_hash_table_remove(ofono_card_data_map, card);`. As I see it now, the ` ofono_remove_all_cards` can also free resources by calling `g_hash_table_remove_all`.

src/sco.c
```diff
@@ -457,7 +459,8 @@ static void *sco_msbc_dec_thread(struct ba_transport_thread *th) {
 		if (!ba_transport_pcm_is_active(t_pcm))
 			continue;
 
-		ffb_seek(&msbc.data, len);
+		if (len > 0)
+			ffb_seek(&msbc.data, len);
```
arkq, line 463: Why msbc thread has got this `if (len>0)` check, where cvsd has got `continue` after read error? I think that the proper fix is to add `continue` after read error. It's definitely a bug to go forward with `-1` as length :D

arkq: OK, my bad.... indeed we need to feed msbc_decoder so PLC can work properly

arkq (approved):

arkq (approved):

PR #625, opened 2023-04-11
README update

borine:
extra detail in context diagram

### DIFF
--- a/README.md
+++ b/README.md
@@ -264,27 +264,9 @@ bluealsa-aplay -L
 
 ## Contributing
 
-This project welcomes contributions of code, documentation and testing. For
-code and manual page contributions, please use GitHub Pull Requests. There is
-no strict policy for PRs, each contribution will be evaluated individually. If
-you wish to help by testing PRs or by making review comments please do so by
-adding comments to the PR.
+This project welcomes contributions of code, documentation and testing.
 
-When preparing a pull request, if possible please configure with
-`--enable-test`, and to catch as many coding errors as possible please compile
-with:
-
-```sh
-make CFLAGS="-Wall -Wextra -Wshadow -Werror"
-```
-
-and then run the unit test suite:
-
-```sh
-make check
-```
-
-The project wiki is "public" and contributions there are also welcome.
+Please see the [CONTRIBUTING](./CONTRIBUTING.md) guide for details.
 
 ## Bug reports, feature requests, and requests for help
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -261,7 +275,7 @@ When preparing a pull request, if possible please configure with
 with:
 
 ```sh
-make CFLAGS="-Wall -Wextra -Werror"
+make CFLAGS="-Wall -Wextra -Wshadow -Werror"
```
arkq, line 278: I think that the entire "Contributing" should be moved to the CONTRIBUTING.md and in the README we should keep only a reference like: "See the [CONTRIBUTING.md file](CONTRIBUTING.md) for contribution guidelines." Having a separate contributing file will align better with GitHub Community Standards (https://docs.github.com/en/communities/setting-up-your-project-for-healthy-contributions/setting-guidelines-for-repository-contributors)

README.md
```diff
@@ -276,9 +290,16 @@ The project wiki is "public" and contributions there are also welcome.
 
 Before raising a new issue, please search previous issues (both open and
 closed), to see if your question has already been answered or problem resolved.
-If reporting a problem, please clearly state the version of BlueALSA that you
-are using, and give sufficient information for readers to be able to reproduce
-the issue.
+If reporting a problem, please clearly state:
+
+* the version of BlueALSA that you are using (`bluealsa --version`),
```
arkq, line 295: I'll try to tinker a little bit with issue and pull request templates. It will be better to provide such information as a context note. For PR template it's nice to have something like that:

```
> !!!!!!!!!! Instructions
>
> Please have a good description of the problem and the fix. Help the reviewer
> understand what to expect.
>
> For new feature, extensive change or non-trivial bug fix add a simple unit test
> or at least describe how to test the code manually.
>
> If you have an issue number, please reference it with a syntax `Fixes #123`.
>
> Please delete these instructions to prove you have read them.
>
> !!!!!!!!!! Instructions End
```

arkq (dismissed):

arkq:
@borine I've pushed here (maybe I should have opened a new PR...) GitHub templates for issues and PRs. What do you think about the content of these context tip notes? I want to keep them short and simple, I don't want to scare anyone with a wall of text. GitHub shows up to 10 lines without scrollbar, so it's the best to keep these notes below that number. Unfortunately, for bug report I've added more lines, but in that case, I think that it's best if user will follow some guidelines. I'm also thinking about adding something like "if possible, before reporting a bug, check whether the issue occurs on master branch" (but maybe with better wording).

PR #632, opened 2023-04-21
fix A2DP source acquistion on slow machines

### LINKED ISSUE #627: Fail to use google home mini as bluetooth speaker
### Problem

Hi, I'm trying to use my google home mini as a bluetooth speaker to play audio from a raspberry pi 4. The raspberry uses dietpi, a distro that doesn't have pulseaudio nor pipewire, just plain ALSA. After pairing and connecting to the speaker using `bluetoothctl`, running `aplay -D bluealsa beep.wav` works just the first time, then it fails to play further sounds until a reboot is performed, followed by a reconnect. Using the home mini as a bluetooth speaker from my android phone works fine with no issues.

Here are the debug logs of the bluealsa service, which contain the successful play attempt and a couple failed attempts: 
[logs.txt](https://github.com/arkq/bluez-alsa/files/11240344/logs.txt)

And here are the logs of aplay:
```
root@romeo:~# aplay -D bluealsa beep.wav 
[1491] D: ../../../src/asound/bluealsa-pcm.c:1380: Getting BlueALSA PCM: PLAYBACK 00:00:00:00:00:00 a2dp
[1491] D: ../../../src/asound/bluealsa-pcm.c:1133: /org/bluealsa/hci0/dev_F0_EF_86_6A_C6_3D/a2dpsrc/sink: Setting constraints
Playing WAVE 'beep.wav' : Signed 16 bit Little Endian, Rate 44100 Hz, Mono
[1491] D: ../../../src/asound/bluealsa-pcm.c:546: /org/bluealsa/hci0/dev_F0_EF_86_6A_C6_3D/a2dpsrc/sink: Initializing HW
[1491] D: ../../../src/asound/bluealsa-pcm.c:566: /org/bluealsa/hci0/dev_F0_EF_86_6A_C6_3D/a2dpsrc/sink: Couldn't open PCM: Did not receive a reply. Possible causes include: the remote application did not send a reply, the message bus security policy blocked the reply, the reply timeout expired, or the network connection was broken.
aplay: set_params:1407: Unable to install hw params:
ACCESS:  RW_INTERLEAVED
FORMAT:  S16_LE
SUBFORMAT:  STD
SAMPLE_BITS: 16
FRAME_BITS: 16
CHANNELS: 1
RATE: 44100
PERIOD_TIME: 125000
PERIOD_SIZE: (5512 5513)
PERIOD_BYTES: (11024 11026)
PERIODS: (3 5)
BUFFER_TIME: 500000
BUFFER_SIZE: 22050
BUFFER_BYTES: 44100
TICK_TIME: 0
[1491] D: ../../../src/asound/bluealsa-pcm.c:457: /org/bluealsa/hci0/dev_F0_EF_86_6A_C6_3D/a2dpsrc/sink: Closing
```

### Setup

**OS**: Linux 5.15.84-v8+, DietPi v8.14.2, RPi 4 Model B (aarch64)
**Bluealsa**: I'm using the master branch of `bluez-alsa` (541523ae1a3c4c2fb7264543777a48a93d66cd47), built using the following flags: `../configure --enable-mp3lame  --enable-manpages --enable-systemd --enable-debug --disable-payloadcheck`
**BlueZ**: 5.55
**ALSA**: aplay: version 1.2.4

borine:
My fix for the thread synchronization problem mentioned here https://github.com/arkq/bluez-alsa/pull/618#issuecomment-1483938339

Hopefully it also fixes #627 

I had to special-case the A2DP source profile because I was unable to find a more general solution.

Tested by inserting a call to usleep(3) to simulate a slow host:

```diff
diff --git a/src/ba-transport.c b/src/ba-transport.c
index d26ffef..cc82b45 100644
--- a/src/ba-transport.c
+++ b/src/ba-transport.c
@@ -1633,6 +1633,8 @@ int ba_transport_acquire(struct ba_transport *t) {
 final:
 	pthread_mutex_unlock(&t->bt_fd_mtx);
 
+usleep(1000);
+
 	if (acquired) {
 
 		if (t->profile != BA_TRANSPORT_PROFILE_A2DP_SOURCE) {
```

With that call I get 100% deadlocks with master HEAD and 0% deadlocks with this PR.

### DIFF
--- a/src/ba-transport.c
+++ b/src/ba-transport.c
@@ -325,6 +325,7 @@ int ba_transport_thread_state_set(
 	pthread_mutex_lock(&th->mutex);
 
 	enum ba_transport_thread_state old_state = th->state;
+
 	/* Moving to the next state is always allowed. */
 	bool valid = state == th->state + 1;
 
@@ -1482,12 +1483,24 @@ void ba_transport_set_codec(
  *   errno is set to indicate the error. */
 int ba_transport_start(struct ba_transport *t) {
 
+	/* For A2DP Source profile only, it is possible that BlueZ will
+	 * activate the transport following a D-Bus "Acquire" request before the
+	 * client thread has completed the acquisition procedure by initializing
+	 * the I/O threads state. So in that case we must ensure that the
+	 * acquisition procedure is not still in progress before we check the
+	 * threads' state. */
+	if (t->profile == BA_TRANSPORT_PROFILE_A2DP_SOURCE)
+		pthread_mutex_lock(&t->acquisition_mtx);
+
 	pthread_mutex_lock(&t->thread_enc.mutex);
 	bool is_enc_idle = t->thread_enc.state == BA_TRANSPORT_THREAD_STATE_IDLE;
 	pthread_mutex_unlock(&t->thread_enc.mutex);
 	pthread_mutex_lock(&t->thread_dec.mutex);
 	bool is_dec_idle = t->thread_dec.state == BA_TRANSPORT_THREAD_STATE_IDLE;
 	pthread_mutex_unlock(&t->thread_dec.mutex);
+
+	if (t->profile == BA_TRANSPORT_PROFILE_A2DP_SOURCE)
+		pthread_mutex_unlock(&t->acquisition_mtx);
 
 	if (!is_enc_idle || !is_dec_idle)
 		return errno = EINVAL, -1;


arkq:
Hmm... I'm not sure whether it's OK. As far as I remember, I've put setting IDLE in the 'final' block, so it will not be executed under `t->bt_fd_mtx` lock. Otherwise, TSAN has reported lock inversion for  `t->bt_fd_mtx` and `th->mutex`. But it's quite possible that it's not reported on CI, because there are more issues out there, so I had to disable TSAN check for tests which are known to fail.

Also, I'm surprised that my quick and dirty fix in https://github.com/arkq/bluez-alsa/issues/627 does not work (if the cause is the same), because I thought that `t->bt_fd_mtx` will synchronize that... but apparently I was wrong. I hope that I will have some time to look at that during the weekend.

Also, I've been thinking a little bit about lock inversion issues and a general "pattern" in bluealsa. The case is that we've got a hierarchy of objects:

```
adapter -> device -> transport -> io-thread -> pcm
                         ┖----------------------┚
```
So in order to prevent lock inversions we have to be consistent and follow that hierarchy when locking. We can lock from left to right and from right to left. To be honest I don't know which one will be the best approach. Currently, most locking is from right to left, because it allows io-threads and pcms to operate without locking transport.

borine:
> I'm surprised that my quick and dirty fix in #627 does not work

That fix sets the thread state only **after** acquire() has returned, and by then it is already too late. acquire() unlocks t->bt_fd_mtx and so ba_transport_thread_create() is free to create the encoder thread without waiting for any other thread.

>  TSAN has reported lock inversion for t->bt_fd_mtx and th->mutex

Ah yes, I see that now. So the fix will require changes at a higher design level after all. I'll look at it again next week, but if it turns out that major surgery is required then I will pass and leave the more serious engineering work to you.

arkq:
> So the fix will require changes at a higher design level after all

It's one possibility. But the most simple approach would be to add dedicated synchronization with a condition variable and a "state". But I deliberately have not fixed it like that yet, because maybe there is a cleaner solution :)

borine:
I've found 2 separate problems with thread synchronization during transport acquisition. The first affects both A2DP and SCO and very occasionally causes test failures in test-utils-cli and test-alsa-pcm. The second is the A2DP source profile issue which was the original focus of the PR. I've had a second go at fixing that. The original failed commit in the PR is now discarded and I've force-pushed 2 new commits to address these 2 problems.

arkq:
> Make transport acquisition an atomic operation

Nice catch! I will forward it to master right away :)

arkq:
As for the second issue, how about wrapping "bool is_enc_idle = ....; bool is_dec_idle = ...` with the new `acquisition_mtx` mutex? This will synchronize acquire with checks in the start function, and TSAN should also be OK. Maybe it will work for all cases?

borine:
> wrapping "bool is_enc_idle = ....; bool is_dec_idle = ..." with the new acquisition_mtx ... Maybe it will work for all cases?

Not as it stands. For SCO profiles ba_transport_acquire() calls ba_transport_start() whilst holding the acquisition mutex locked. So ba_transport_start() cannot then also lock that mutex.

Perhaps it will be safe for ba_transport_acquire() to unlock the mutex before calling ba_transport_start(). I will look into that later.

borine:
> Perhaps it will be safe for ba_transport_acquire() to unlock the mutex before calling ba_transport_start()

OK, so that is not safe. But I do prefer your idea of using the acquisition mutex to synchronize the A2DP source case. I've rebased and changed the implementation to use your suggestion.

arkq:
> OK, so that is not safe. 

OK, I was thinking exactly about that, moving start beyond mutexes (but I was not sure what about release() in case of start failure). So my other thought was that maybe there should be "internal" start and public start (which will add "bool is_enc_idle = ....; " checks)...

Anyway, many thanks for fixing that issue. It was the last big thing on my list before preparing a new release of bluez-alsa.

arkq (approved):

PR #641, opened 2023-06-13
Gitignore update

borine:
Since commit 2709a34, generated c source files are created in the src directory, and this can make some git actions such as switching branches awkward. These files also create problems when running spellcheck in a local work area. These
commits just make working with git a little bit easier.

### DIFF
--- a/.github/spellcheck.yaml
+++ b/.github/spellcheck.yaml
@@ -22,7 +22,7 @@ matrix:
 
 - name: C
   sources:
-  - 'src/**/*.{c,h}|!config.h'
+  - 'src/**/*.{c,h}|!config.h|!src/*-iface.c'
   aspell: *aspell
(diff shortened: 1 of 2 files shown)


arkq:
These files are generated in the source directory only if you are upgrading your current git build, because (I think) autotools somehow caches the location of these files and they are generated in the previous location. If you will distclean the project and regenerate everything from scratch, these files will not be generated in the source directory.

I've also seen these issue, but I did not add these files to gitignore, because I thought that it might be a "bug" only visible in case of "git pull". But maybe I'm wrong, because, to be honest, I do not know why these files are generated in the src dir... a cache thingy is only a wild guess...

borine:
Ah, thanks. So all I needed to do was `make distclean`. This PR is not needed, closing.

PR #644, opened 2023-06-15
aplay: properly handle a2dp codec change

borine:
See discussion #643 for description of problem.

### DIFF
--- a/utils/aplay/aplay.c
+++ b/utils/aplay/aplay.c
@@ -911,13 +911,19 @@ static DBusHandlerResult dbus_signal_handler(DBusConnection *conn, DBusMessage *
 			}
 			if (pcm.transport == BA_PCM_TRANSPORT_NONE)
 				goto fail;
-			struct ba_pcm *tmp = ba_pcms;
-			if ((ba_pcms = realloc(ba_pcms, (ba_pcms_count + 1) * sizeof(*ba_pcms))) == NULL) {
-				error("Couldn't add new BlueALSA PCM: %s", strerror(ENOMEM));
-				ba_pcms = tmp;
-				goto fail;
+
+			struct ba_pcm *stored_pcm;
+			if ((stored_pcm = get_ba_pcm(pcm.pcm_path)) == NULL) {
+				struct ba_pcm *tmp = ba_pcms;
+				if ((ba_pcms = realloc(ba_pcms, (ba_pcms_count + 1) * sizeof(*ba_pcms))) == NULL) {
+					error("Couldn't add new BlueALSA PCM: %s", strerror(ENOMEM));
+					ba_pcms = tmp;
+					goto fail;
+				}
+				stored_pcm = &ba_pcms[ba_pcms_count++];
 			}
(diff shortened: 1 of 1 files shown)


utils/aplay/aplay.c
```diff
@@ -911,13 +911,19 @@ static DBusHandlerResult dbus_signal_handler(DBusConnection *conn, DBusMessage *
...
 			if (pcm.transport == BA_PCM_TRANSPORT_NONE)
 				goto fail;
-			struct ba_pcm *tmp = ba_pcms;
-			if ((ba_pcms = realloc(ba_pcms, (ba_pcms_count + 1) * sizeof(*ba_pcms))) == NULL) {
-				error("Couldn't add new BlueALSA PCM: %s", strerror(ENOMEM));
-				ba_pcms = tmp;
-				goto fail;
+
+			struct ba_pcm *stored_pcm;
+			if ((stored_pcm = get_ba_pcm(pcm.pcm_path)) == NULL) {
```
arkq, line 916: Are you sure that this will resolve the issue? When the codec is changes bluealsa emits "InterfacesRemoved" followed by "InterfacesAdded". There should be no case where PCM with the same path is already in the bluealsa-aplay PCMs cache. One theoretical case is that the signals will be delivered in a reversed order, but it seems that that's not the case for the reference D-Bus implantation.

borine: Yes, sure. The "InterfacesRemoved" handler removes only the worker, it does not remove the pcm from the cache. In fact no line of code ever reduces `ba_pcms_count`, so the pcm cache array simply grows and grows. That is a memory leak, and in addition, because the array is always searched from the beginning, it means that when looking up a pcm from its path only the first connection is found, but the current connection may use a different codec and that is why we are getting the symptoms described in #643

arkq: >  it does not remove the pcm from the cache

Right! So, I think that the proper fix would be to address that memory leak. Right now it might be tricky because of the memory layout. But maybe after changing the `ba_pcms` to `struct ba_pcm **` it would be easier.

borine: > address that memory leak

This commit does "address" the leak in one way, in that any pcm can have only one entry in the cache. So the cache array size now represents the number of distinct pcms that have ever connected, rather than the total number of connections. But yes, it may be better create space in the array when a pcm disconnects, in the same way as is done for the workers array.

arkq: How about something like that (it's not tested though): https://github.com/arkq/bluez-alsa/commit/24f74d642a677ccd845a4e91045c5bb59e60cef2

arkq: OK, I think I've found the root cause of that issue. The problem causes the PropertiesChanged signal which is emitted with "Running true". Then the pcm struct in the worker array is compared with wrong ba_pcms pcm (due to new PCM being added at the end of the ba_pcms list, but `get_ba_pcm()` returns the first match) which triggers PCM stop and start, but it is started with updated pcm structure with old/staled pcm from the `ba_pcms`.

Also, it seems that removing and adding pcm to the ba_pcms works as well after all.

PR #645, opened 2023-06-17
SBC: Require sbc >= 1.5.

paul-1:
function sbc_reinit_a2dp was not exposed in the shared library until v1.5

See #639 

Use of sbc_reinit_a2dp was added in commit 44a5dd9a17eaf3b2a92ea2d1859b64f9837f2b4a


borine:
Please can you also mention the version requirement in INSTALL.md (line 32) ? Thanks

paul-1:
I was going to, but noticed only bluez had a version called out there.  Should we add versions to all libraries in Install.md?

arkq (approved):

arkq:
Many thanks for the patch!

PR #652, opened 2023-07-11
Provide option to set RT priority for IO threads

arkq:
This is an experimental feature which can improve audio quality with low-latency codecs such as mSBC.

### DIFF
--- a/.github/iwyu.imp
+++ b/.github/iwyu.imp
@@ -1,6 +1,7 @@
 [
 
 	{ include: [ '<bits/getopt_core.h>', private, '<getopt.h>', public ] },
+	{ include: [ '<bits/types/struct_sched_param.h>', private, '<sched.h>', public ] },
 	{ include: [ '<bsd/libutil.h>', private, '<bsd/stdlib.h>', public ] },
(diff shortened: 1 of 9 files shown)


borine:
I think the manual page reference should be `sched(7)`

Also, the systemd directives need modification for RT priority:
```
AmbientCapabilities=CAP_NET_RAW CAP_SYS_NICE
RestrictRealtime=false
```
and in addition the system call filter exclusion list must not include @resources, since the sched_* family of calls are included in that group, so change it to
```
SystemCallFilter=~@privileged
```

I can add a section to the systemd wiki page about this once the PR is merged, although perhaps it is worth adding some (commented-out) lines to the distributed systemd service file?

arkq:
> I think the manual page reference should be sched(7)

+1, nice catch :)

> although perhaps it is worth adding some (commented-out) lines to the distributed systemd service file?

Yes, good idea. If you are OK, I'll add your suggestions.

I'm still testing this, because elevated priority might uncover some not-found yet race-related bugs. However, my general experience with mSBC is a much better quality (practically no missing frames).

PR #653, opened 2023-07-11
Delay adjustment

borine:
See discussion #650 for rationale. Extends the D-Bus API org.bluealsa.PCM1 to allow setting a persistent adjustment that is applied to the Delay property of the PCM. Each codec supported by the PCM has its own delay adjustment.

The choice of control name for the CTL plugin is not yet agreed, and in this initial draft the name "Sync" is used.

### DIFF
--- a/doc/bluealsa-plugins.7.rst
+++ b/doc/bluealsa-plugins.7.rst
@@ -98,9 +98,11 @@ PCM Parameters
     softvol value. The default value is **unchanged**.
 
   DELAY
-    An integer number which is added to the reported latency value in order to
-    manually adjust the audio synchronization. It is not normally required and
-    defaults to **0**.
+    An integer number which is added to the reported delay (latency) value in
+    order to manually adjust the audio synchronization. It is not normally
+    required and defaults to **0**. See the **EXT** parameter of the CTL plugin
+    in the `CTL Parameters`_ section below for a more flexible and convenient
+    method of manually adjusting the reported delay by using a mixer control.
 
   SRV
     The D-Bus service name of the BlueALSA daemon. Defaults to
@@ -420,15 +422,15 @@ CTL Parameters
 
   EXT
     Causes the plugin to include extra controls. These are the controls for
-    Bluetooth codec selection, volume mode selection and/or battery level
-    indicator.
+    Bluetooth codec selection, volume mode selection, delay adjustment (sync)
+    and/or battery level indicator.
     If the value is **yes** then all of these additional controls are included;
     if the value is **no** then none of them are included. The default is
     **no**.
 
     This parameter can also select individual controls by using a colon (':')
     separated list of control names. The control names are **codec**, **mode**,
-    and **battery**. For example:
+    **sync** and **battery**. For example:
 
     ::
 
@@ -442,6 +444,19 @@ CTL Parameters
     playback control has index 0 and capture control has index 1.
     See the `Volume control` section in the ``bluealsa(8)`` for more
     information on the software volume setting.
+
+    The delay adjustment controls are called "Sync". They can be used to apply
+    a fixed adjustment to the delay reported by the associated PCM to the
+    application, and may be useful with applications that need to synchronize
+    the bluetooth audio stream with some some other stream, such as a video.
+    The values are in milliseconds from ``-3275 ms`` to ``+3275 ms`` in steps
+    of ``25 ms``. The playback control has index 0 and the capture control has
+    index 1. Each codec supported by a PCM has its own delay adjustment value.
+    Note that this control changes only the delay value reported to the
+    application by ALSA, it does not affect the actual delay (latency) of the
+    PCM stream. Values set by this control type are saved in the BlueALSA
+    persistent state files, and so are remembered and automatically applied
+    each time the PCM is used.
 
     The read-only battery level indicator will be shown only if the device
     supports battery level reporting.
--- a/src/asound/bluealsa-ctl.c
+++ b/src/asound/bluealsa-ctl.c
@@ -33,6 +33,11 @@
 #include "shared/dbus-client.h"
 #include "shared/defs.h"
 
+#define DELAY_SYNC_STEP 250
+#define DELAY_SYNC_MIN_VALUE (INT16_MIN / DELAY_SYNC_STEP * DELAY_SYNC_STEP)
+#define DELAY_SYNC_MAX_VALUE (INT16_MAX / DELAY_SYNC_STEP * DELAY_SYNC_STEP)
+#define DELAY_SYNC_NUM_VALUES (1 + (DELAY_SYNC_MAX_VALUE - DELAY_SYNC_MIN_VALUE) / DELAY_SYNC_STEP)
+
 /**
  * Control element type.
  *
@@ -43,6 +48,7 @@ enum ctl_elem_type {
 	CTL_ELEM_TYPE_VOLUME,
 	CTL_ELEM_TYPE_VOLUME_MODE,
 	CTL_ELEM_TYPE_CODEC,
+	CTL_ELEM_TYPE_DELAY_SYNC,
 	CTL_ELEM_TYPE_BATTERY,
 };
 
@@ -124,6 +130,8 @@ struct bluealsa_ctl {
 	bool show_codec;
 	/* if true, show volume mode control */
 	bool show_vol_mode;
+	/* if true, show delay adjustment sync control */
+	bool show_delay_sync;
 	/* if true, show battery level indicator */
 	bool show_battery;
 	/* if true, append BT transport type to element names */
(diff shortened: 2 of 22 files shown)


src/asound/bluealsa-ctl.c
```diff
@@ -137,6 +143,99 @@ static const char *soft_volume_names[] = {
 	"software",
 };
 
+static const struct {
+	const char *name;
+	int16_t value;
+} delay_adjustment_map[] = {
```
arkq, line 149: How about making sync a slider instead of enum control element?

Disclaimer: I have not check your branch yet, so maybe indeed enum is a better choice. I'm mostly concerned about the maintenance. If enum is better, how about generating such a list dynamically?

doc/bluealsa-plugins.7.rst
```diff
@@ -416,13 +419,28 @@ CTL Parameters
...
-    and take values "software" and "pass-through"; the playback control has
-    index 0 and capture control index 1. See ``bluealsa(8)`` for more on the
-    soft volume setting , and `Codec selection`_ in the **NOTES** section below
-    for more information on the Codec control.
+    Causes the plugin to include controls for codec, delay adjustment and/or
+    software volume selection. If the value is **yes** then all of these
+    additional controls are included; if the value is **no** then none of them
+    are included. The default is **no**. This parameter can also select a
+    single control, or a colon (':') separated list of controls by using the
+    values **codec**, **sync**, or **softvol**. For example:
```
arkq, line 427: I like the idea of colon separated list! Maybe BAT option should be merged with the EXT? I will experiment with that. If it "feels" OK I'd like to cherrypick this change before merging delay adjustment.

src/ba-transport.c
```diff
@@ -1733,11 +1736,69 @@ bool ba_transport_pcm_is_active(struct ba_transport_pcm *pcm) {
...
+		return;
+
+	for (size_t n = 0; n < pcm->delay_adjustments_count; n++) {
+		if (pcm->delay_adjustments[n].codec_id == codec_id) {
+			pcm->delay_adjustments[n].adjustment = adjustment;
+			return;
+		}
+	}
+
+	struct ba_transport_pcm_delay_adjustment *tmp = reallocarray(
```
arkq, line 1778: Maybe using `GArray` for the `delay_adjustment_count` will be easier? Or maybe event `GHashTable`....

src/shared/dbus-client.c
```diff
@@ -814,9 +814,10 @@ dbus_bool_t bluealsa_dbus_pcm_select_codec(
 		const char *codec,
 		const void *configuration,
 		size_t configuration_len,
+		dbus_bool_t blocking,
```
arkq, line 817: Why blocking/non-blocking mode is required?

arkq: Ooo, it's explained in the commit message

doc/bluealsa-plugins.7.rst
arkq: I've just cherry-picked this feature into the https://github.com/arkq/bluez-alsa/tree/ext branch. I'd be glad if you could review it if I didn't mess up anything :D I know that it's somehow backward incompatible, but I think that we can live with that.

src/asound/bluealsa-ctl.c
borine: >How about making sync a slider instead of enum control element

AFAIK `alsamixer` always shows volume type controls as 0-100%, which is not appropriate for delay adjustment. It also colours lowest 40% of bar height, middle 40% white and top 20% red, which is also not appropriate in this case. If  there is some way to persuade it to show -37250 to +37250 (and without colouring) then I will change to slider.

> how about generating such a list dynamically?

I guess that would not be too difficult, although I haven't done any work on that.

doc/bluealsa-plugins.7.rst
borine: I'll try to do a review in this coming week.

src/ba-transport.c
borine: I have no previous experience with glib, but will have a play with both GArray and GHashTable to see what I can come up with.

src/asound/bluealsa-ctl.c
arkq: > AFAIK alsamixer always shows volume type controls as 0-100%, which is not appropriate for delay adjustment

OK, fair enough :)

src/ba-transport.c
borine: I've now changed the implementation to use GHashTable.

borine:
I've had second thoughts about the non-blocking D-Bus calls, because amixer does not wait for the update to complete and therefore does not report the updated value. So I've now reverted that commit and instead invoke the D-Bus event handling code explicitly after codec and sync updates. This approach seems to keep both amixer and alsamixer happy.

I plan to rebase against master this weekend to pull in the revised EXT argument code, and at the same same squash some of these commits to make each logical change a single commit. I will then mark this PR as ready for review.

BTW I've been searching through dictionaries and thesauruses  to find better names for the delay adjustment control, but given there is only room for 4 characters I still haven't found anything better than "Sync".

arkq:
I've just reviewed commits for CTL plugin and for `bluealsa-cli`, and they look good. I've fixed some typos and reordered sync and battery (battery is the least related with audio, so I'd like to place it as the last item) and force pushed the changes (to avoid conflicts in the nearest future).

Now, I need to find some free time to review the core part of this PR...
However, already I've spotted one thing that I'd like to change. The `ba_transport_pcm_delay_adjustment_set` should get the `codec_id` not a string representation. It will be consistent with the rest of the internal API.

src/bluealsa-dbus.c
```diff
@@ -764,6 +769,62 @@ static void bluealsa_pcm_select_codec(GDBusMethodInvocation *inv, void *userdata
...
+	const struct ba_transport *t = pcm->t;
+
+	GVariantBuilder adjustments;
+	g_variant_builder_init(&adjustments, G_VARIANT_TYPE("a{sn}"));
+
+	GHashTableIter iter;
+	g_hash_table_iter_init(&iter, pcm->delay_adjustments);
+
+	gpointer key, value;
+	while (g_hash_table_iter_next(&iter, &key, &value)) {
```
arkq, line 814: Iteration here needs a mutex to protect against race with `bluealsa_pcm_set_delay_adjustment()`. All BlueALSA D-Bus API calls are handled in a separated thread, so they can run in parallel.

src/storage.c
```diff
@@ -252,6 +287,32 @@ int storage_pcm_data_update(const struct ba_transport_pcm *pcm) {
...
 	GKeyFile *keyfile = st->keyfile;
 	const char *group = pcm->ba_dbus_path;
+	const size_t num_codecs = g_hash_table_size(pcm->delay_adjustments);
+	gchar **list = calloc(num_codecs + 1, sizeof(gchar*));
+	GHashTableIter iter;
+	g_hash_table_iter_init(&iter, pcm->delay_adjustments);
+
+	gpointer key, value;
+	size_t index = 0;
+	while (g_hash_table_iter_next(&iter, &key, &value)) {
```
arkq, line 297: I think that mutex here is also required.

src/ba-transport.c
```diff
@@ -1734,11 +1738,34 @@ bool ba_transport_pcm_is_active(struct ba_transport_pcm *pcm) {
...
+
+	g_hash_table_insert(pcm->delay_adjustments, GINT_TO_POINTER(codec_id), GINT_TO_POINTER(adjustment));
+
+}
+
+int16_t ba_transport_pcm_delay_adjustment_get(const struct ba_transport_pcm *pcm) {
+	struct ba_transport *t = pcm->t;
+	pthread_mutex_lock(&t->codec_id_mtx);
+	int16_t adjustment = 0;
+	gpointer val = g_hash_table_lookup(pcm->delay_adjustments, GINT_TO_POINTER(t->codec_id));
```
arkq, line 1764: Simply use `ba_transport_get_codec(t)` which takes care of mutex lock.

src/ba-transport.c
```diff
@@ -845,11 +849,11 @@ struct ba_transport *ba_transport_new_a2dp(
 	t->acquire = transport_acquire_bt_a2dp;
 	t->release = transport_release_bt_a2dp;
 
-	ba_transport_set_codec(t, codec->codec_id);
-
 	storage_pcm_data_sync(&t->a2dp.pcm);
 	storage_pcm_data_sync(&t->a2dp.pcm_bc);
 
+	ba_transport_set_codec(t, codec->codec_id);
```
arkq, line 855: Why this reorder is required?

arkq (commented):
I think that the core change looks good as well!

src/ba-transport.c
borine: I think its a hangover from previous draft - now reverted

arkq (approved):
I've just white-reviewed the code and it looks OK.

I've added `InvalidArguments` error to the `SetDelayAdjustment`, so now the call like this will return proper D-Bus error:
```sh
gdbus call --system --dest org.bluealsa \
   --object-path /org/bluealsa/hci0/dev_04_5D_4B_ED_D7_00/a2dpsrc/sink \
   --method org.bluealsa.PCM1.SetDelayAdjustment xxx 33
```

I will perform few runtime tests and I think that we are ready to go :)

Also, I like very much the idea of "mutable" from C++! I'll try to adopt that in the whole project. Maybe with a bit different semantic. I thought about adding define just for mutexes (I don't see any other ba-structs member that can be "mutable"), something like `#define MUTABLE_MUTEX(mtx) ...`.

arkq (dismissed):

PR #657, opened 2023-07-29
src/Makefile.am: Install D-Bus policy in /usr/share, not /etc

gioele:
From https://bugs.debian.org/1006631:

> dbus supports policy files in both `/usr/share/dbus-1/system.d` and
> `/etc/dbus-1/systemd`. [The] recently released dbus 1.14.0, officially
> deprecates installing packages' default policies into `/etc/dbus-1/systemd`,
> instead reserving it for the sysadmin. This is the same idea as the
> difference between `/usr/lib/udev/rules.d` and `/etc/udev/rules.d`.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -342,7 +342,7 @@ AC_SUBST([ALSA_CONF_DIR], [$alsaconfdir])
 AC_ARG_WITH([dbusconfdir],
 	AS_HELP_STRING([--with-dbusconfdir=DIR], [path to D-Bus system bus configuration files]),
 	[dbusconfdir="${withval}"],
-	[dbusconfdir=$($PKG_CONFIG --variable=sysconfdir dbus-1)/dbus-1/system.d])
+	[dbusconfdir=$($PKG_CONFIG --variable=datadir dbus-1)/dbus-1/system.d])
 AC_SUBST([DBUS_CONF_DIR], [$dbusconfdir])
 
 AC_ARG_WITH([systemdsystemunitdir],


arkq:
I wonder how this will affect systems with older versions of dbus. Will it read the configuration correctly?

Also, the right place to change the default would be here:
https://github.com/arkq/bluez-alsa/blob/master/configure.ac#L345 (according to the dbus-1.pc file on Debian 12)

`--variable=sysconfdir` -> `--variable=datadir`

But also, it will be required to check whether the `datadir` variable was provided in older versions of dbus.

And another question is whether we need to provide customization for that directory in the configure.ac after all. Maybe configuration should just read the value from the pkg-config file and that's all...

gioele:
> Also, the right place to change the default would be here: https://github.com/arkq/bluez-alsa/blob/master/configure.ac#L345 According to the dbus-1.pc file on Debian 12: `--variable=sysconfdir` -> `--variable=datadir`

Right. I amended the PR. 

> I wonder how this will affect systems with older versions of dbus. Will it read the configuration correctly?

D-Bus reads from /usr since version 1.10, released in 2015: https://sources.debian.org/src/dbus/1.10.32-0%2Bdeb9u1/NEWS/#L483-L499

> But also, it will be required to check whether the `datadir` variable was provided in older versions of dbus.

`datadir` has been added to `dbus-1.pc`  in 2014: https://salsa.debian.org/utopia-team/dbus/-/commit/95fe17a96d0488da79c5773da9b86b5431fa28c9

arkq:
> D-Bus reads from /usr since version 1.10, released in 2015

OK, so I think that we can change the default then.

arkq (approved):

PR #661, opened 2023-08-16
Fix adjust volume invalid in using soft volume for sco

xiaoyao888888:
When using software volume, the sco volume cannot be adjusted, while it is normal in a2dp mode, so rename a2dp.volume to volume, and both a2dp/sco profiles can use it.


arkq:
So in your case scenario you would like to have software volume by default for A2DP and for HFP/HSP ? Because currently BlueALSA uses hardware volume (it simply forwards volume requests to connected BT device) by default for HFP (because it is a feature covered by very early standard of HSP, and it works reliably as far as I know), and software volume for A2DP by default. The A2DP is software by default because it was not reliably in the past (due to bugs in BlueZ). But maybe in the future I will change it to "hardware" by default as well, because everything works fine with recent versions of BlueZ (I do not remember now in which BlueZ version the bug was fixed). Anyway the software volume can be selected with alsamixer or with bluealsa-cli.

PR #664, opened 2023-09-01
bluealsa-aplay: option to select volume mode

borine:
A proposal to improve `bluealsa-aplay` volume control. Inspired by the discussion in PR #662, this proposal gives the `bluealsa-aplay` utility the ability to select soft volume or pass-through volume control independently of the `bluealsa` service defaults.

Default behavior remains unchanged, so backwards compatibility for existing installations is maintained.

See the changes to `doc/bluealsa-aplay.1.rst` for details of the new command-line option and the various new behaviors made available.

### DIFF
--- a/doc/bluealsa-aplay.1.rst
+++ b/doc/bluealsa-aplay.1.rst
@@ -6,7 +6,7 @@ bluealsa-aplay
 a simple bluealsa player
 ------------------------
 
-:Date: January 2023
+:Date: September 2023
 :Manual section: 1
 :Manual group: General Commands Manual
 :Version: $VERSION$
@@ -20,8 +20,7 @@ DESCRIPTION
 ===========
 
 Capture audio streams from Bluetooth devices (via ``bluealsa(8)``) and play
-them to an ALSA
-playback device.
+them to an ALSA playback device.
 
 By default **bluealsa-aplay** captures audio from all connected Bluetooth
 devices.  It is possible to select specific Bluetooth devices by providing a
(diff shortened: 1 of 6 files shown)


doc/bluealsa-aplay.1.rst
```diff
@@ -107,20 +107,44 @@ OPTIONS
...
+      then operate the same as for **auto** above.
+
+    - **none** - **bluealsa-aplay** will force the BlueALSA PCM volume mode
+      setting to native ("pass-through") before starting the PCM
+      stream.  It will not operate its configured ALSA mixer. This can be used
+      to effectively disable remote volume control; or it can be used to allow
+      some other application to apply remote volume change requests.
+
+    - **software** - **bluealsa-aplay** will force the BlueALSA PCM volume mode
+      setting to ``software`` and then will not operate its configured ALSA
```
arkq, line 131: Why different quoting is used here: ``software`` vs "pass-through" ?

utils/aplay/aplay.c
```diff
@@ -1139,6 +1174,21 @@ int main(int argc, char *argv[]) {
...
+	if (strcmp(volume_type_str, "mixer") == 0)
+		volume_type = VOL_TYPE_MIXER;
+	else if (strcmp(volume_type_str, "software") == 0)
+		volume_type = VOL_TYPE_SOFTWARE;
+	else if (strcmp(volume_type_str, "none") == 0)
+		volume_type = VOL_TYPE_NONE;
+	else if (strcmp(volume_type_str, "auto") == 0)
+		volume_type = VOL_TYPE_AUTO;
+	else {
+		fprintf(stderr, "Invalid volume control type '%s'. Try '%s --help' for more information.\n", volume_type_str, argv[0]);
```
arkq, line 1186: How about printing the error message in a similar way `bluealsa` does it? The user will get the info about correct values right away. Also (IMHO) it will simplify autocompletion for this option.

utils/aplay/aplay.c
```diff
@@ -1139,6 +1174,21 @@ int main(int argc, char *argv[]) {
...
+		return EXIT_FAILURE;
+	}
+	if (volume_type == VOL_TYPE_NONE || volume_type == VOL_TYPE_SOFTWARE)
```
arkq, line 1189: I would set this in the "if .. else .." block above.

utils/aplay/aplay.c
```diff
@@ -1148,18 +1198,26 @@ int main(int argc, char *argv[]) {
...
 				"  ALSA mixer device: %s\n"
-				"  ALSA mixer element: '%s',%u\n"
+				"  ALSA mixer element: %s%s%s%s\n"
 				"  Bluetooth device(s): %s\n"
 				"  Profile: %s",
 				dbus_ba_service,
 				pcm_device, pcm_buffer_time, pcm_period_time,
-				mixer_device, mixer_elem_name, mixer_elem_index,
+				volume_type_str,
+				mixer_device ? mixer_device : "(not used)",
```
arkq, line 1216: Can you prepare this string with standard "if" logic, something like:
```c
char mixer_element[128] = "(not used)";
if (mixer_device != NULL)
    snprintf(mixer_element_name, sizeof(mixer_element_name), "'%s',%u",
        mizer_elem_name, mixer_elem_index);

```

utils/aplay/aplay.c
```diff
@@ -1148,18 +1198,26 @@ int main(int argc, char *argv[]) {
 		for (i = 0; i < ba_addrs_count; i++, tmp += 19)
 			ba2str(&ba_addrs[i], stpcpy(tmp, ", "));
 
+		char index_str[sizeof(",NNN")];
+		snprintf(index_str, sizeof(index_str), ",%u", mixer_elem_index % 1000);
```
arkq, line 1202: Why `mixer_elem_index` is wrapped here? If it needs to be constrained to [0, 1000], is should be done in the "getopt" logic.

arkq (dismissed):
It's hard for me to judge the user experience with such `--volume=` option (volume control is not my every day use case here), but from the code point of view it looks great!

borine:
> volume control is not my every day use case

To be honest, nor is it mine. I do not use `bluealsa-aplay` with SCO, and I always use software volume control with A2DP. Given how few issues/feature requests/PRs there are relating to `bluealsa-aplay` I think my use case is possibly the most common, and those users/distributions who do use mixer volume control seem to be generally satisfied with the existing implementation. It's at this point in discussions that I always suggest that we solicit opinions from the user community; but we know from experience that rarely generates much interest.

The motivation for this PR comes from PR #662, and is three-fold:

1. The creator of PR #662 @xiaoyao888888 wishes to use `bluealsa-aplay` with SCO, and have software volume control by default. This PR offers an alternative solution to that requirement, and I would be very interested to hear any comments from @xiaoyao888888 on his opinion of the relative merits of both solutions for his own specific use case.

2. In PR #662, you hinted that you may at some point in the future switch A2DP volume control to native by default, which would be inconvenient for my own use case. So this PR is a blatantly selfish attempt to pre-empt such a change by having some mitigation already in place.

3. In order to make the solution as general as possible, I tried to imagine other likely use cases, and came up with the "auto" and "none" options as potentially useful, but without any concrete evidence that any real use cases exist.

I've pushed a new commit to address the review comments so far, and am happy to make further changes is requested.

arkq:
> In order to make the solution as general as possible, I tried to imagine other likely use cases, and came up with the "auto" and "none" options as potentially useful

Yes, I think that this feature provides very generic solution without much complexity. I'll play with it a bit and I think it's ready to go to master. Many thanks for that!

arkq (approved):

PR #667, opened 2023-09-25
Improve PCM plugin drain

borine:
Implement support for applications that use snd_pcm_abort() or snd_pcm_nonblock(pcm, 2) to request that blocking calls are aborted by a signal. Also apply a timeout to the local drain so that an application can not be blocked indefinitely should the bluealsa server stop reading from the PCM FIFO.

This PR is an attempt to address the plugin issues revealed indirectly by #666

### DIFF
--- a/src/asound/bluealsa-pcm.c
+++ b/src/asound/bluealsa-pcm.c
@@ -675,42 +675,94 @@ static int bluealsa_drain(snd_pcm_ioplug_t *io) {
 	 * always either finishes in state SND_PCM_STATE_SETUP or returns an error.
 	 * It is not possible to finish in state SND_PCM_STATE_DRAINING and return
 	 * success; therefore is is impossible to correctly implement capture
-	 * drain logic. So for capture PCMs we do nothing and return success. */
+	 * drain logic. So for capture PCMs we do nothing and return success;
+	 * ioplug will stop the PCM. */
+	if (io->stream == SND_PCM_STREAM_CAPTURE)
+		return 0;
 
-	if (io->stream == SND_PCM_STREAM_PLAYBACK) {
+	/* We must ensure that all remaining frames in the ring buffer are flushed
+	 * to the FIFO by the I/O thread. It is possible that the client has called
+	 * snd_pcm_drain() without the start_threshold having been reached, or
+	 * while paused, so we must first ensure that the IO thread is running. */
+	if (bluealsa_start(io) < 0) {
+		/* Insufficient resources to start a new thread - so we have no choice
+		 * but to drop this stream. */
+		bluealsa_stop(io);
+		snd_pcm_ioplug_set_state(io, SND_PCM_STATE_SETUP);
+		return -EIO;
+	}
 
-		/* We must ensure that all remaining frames in the ring buffer are
-		 * flushed to the FIFO by the I/O thread. It is possible that the
-		 * client has called snd_pcm_drain() without the start_threshold
-		 * having been reached, or while paused, so we must first ensure that
-		 * the IO thread is running. */
-		if (bluealsa_start(io) < 0)
-			return 0;
+	/* For a non-blocking drain, we do not wait for the drain to complete. */
+	if (io->nonblock == 1)
+		return -EAGAIN;
 
-		/* Block until the drain is complete. */
-		struct pollfd pfd = { pcm->event_fd, POLLIN, 0 };
-		while (bluealsa_pointer(io) >= 0 && io->state == SND_PCM_STATE_DRAINING) {
-			if (poll(&pfd, 1, -1) == -1) {
-				if (errno == EINTR)
+	struct pollfd pfd = { pcm->event_fd, POLLIN, 0 };
+	bool aborted = false;
+	int ret = 0;
+
+	snd_pcm_sframes_t hw_ptr;
+	while ((hw_ptr = bluealsa_pointer(io)) >= 0 && io->state == SND_PCM_STATE_DRAINING) {
+
+		snd_pcm_uframes_t avail = snd_pcm_ioplug_hw_avail(io, hw_ptr, io->appl_ptr);
+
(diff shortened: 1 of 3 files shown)


src/asound/bluealsa-pcm.c
```diff
@@ -666,47 +666,69 @@ static int bluealsa_drain(snd_pcm_ioplug_t *io) {
...
-				eventfd_t event;
-				eventfd_read(pcm->event_fd, &event);
-				if (event & 0xDEAD0000)
-					break;
-			}
+	/* We must ensure that all remaining frames in the ring buffer are flushed
+	 * to the FIFO by the I/O thread. It is possible that the client has called
+	 * snd_pcm_drain() without the start_threshold having been reached, or
+	 * while paused, so we must first ensure that the IO thread is running. */
+	if (bluealsa_start(io) < 0)
```
arkq, line 681: > we must first ensure that the IO thread is running

ALSA behaves in that way with other PCMs as well, i.e. it starts it in case of PCM being not stated or paused? Also, what in case of nonblock? On one stack overflow I've read that the drain does not perform anything in such case...

Sorry for such questions, but I'm off my PC until the end of this week.

borine: > ALSA behaves in that way with other PCMs as well, i.e. it starts it in case of PCM being not stated or paused?

ALSA pcm.c delegates drain behaviour to the specific plugin. For `hw` it is determined by the driver, drain is a call to `ioctl(hw->fd, SNDRV_PCM_IOCTL_DRAIN)`. However, looking at the alsa-utils `speaker-test` application, by default that writes less than `start-threshold` frames then calls `snd_pcm_drain()`, so clearly it expects that the plugin will auto-start.

In the ioplug built-in implementation there is an explicit call to `snd_pcm_ioplug_start()`, but that is only used when the plugin does not implement its own callback.

For the `PAUSED` state then ioplug does the same as for the `PREPARED` state, so this bluealsa implementation matches the ioplug built-in one. For `hw` I don't have any pausable devices to experiment with, so would have to study the kernel code; but I suspect that starting the PCM is the correct behaviour. After all, in a single threaded application using default blocking behaviour, there is no other way to start the PCM once drain() is called.

> what in case of nonblock?

For non-block, snd_pcm_drain() should simply set the state to SND_PCM_STATE_DRAINING and return with -EAGAIN. It is then the responsibility of the application to manage the drain itself using snd_pcm_avail() and snd_pcm_writei() etc. It is not possible for bluealsa to implement that without some re-design because the call to `bluealsa_dbus_pcm_ctrl_send_drain()` is always blocking.

When used with the rate plugin, rate will honour the nonblock flag until its own buffer is drained, and only then call the slave drain callback. So if an application using nonblocking calls snd_pcm_drain() only once then most likely bluealsa_drain() is never called, and bluelasa_stop() will be called without ever requesting the server drain the FIFO and encoder.

Perhaps the plugins manual page should mention that drain is blocking? In practice I do not know of any applications that use non-blocking drain. Even those that use non-blocking writes clear the non-block flag before calling drain (eg MPD, and also the alsa-lib sample test programs). I think there may have been issues with ALSA non-blocking drain in the past which may or may not still exist.

Some other thoughts:

1. Perhaps the poll timeout should be increased to give the IO thread more time than it can possibly need, for example add an arbitraty 100ms.

2. ioplug invokes the plugin drain with the ALSA PCM mutex unlocked. So it is possible for the application to change the avail_min parameter during the drain. I think that is unlikely to happen in any sensible application, so I have ignored the possiblility here.

borine: > For the PAUSED state ... would have to study the kernel code;

The kernel sound core does indeed unpause a paused PCM in response to a playback drain ioctl, see:
https://github.com/torvalds/linux/blob/9ed22ae6be817d7a3f5c15ca22cbc9d3963b481d/sound/core/pcm_native.c#L2115

src/asound/bluealsa-pcm.c
```diff
@@ -666,47 +666,84 @@ static int bluealsa_drain(snd_pcm_ioplug_t *io) {
...
-		/* Block until the drain is complete. */
-		struct pollfd pfd = { pcm->event_fd, POLLIN, 0 };
-		while (bluealsa_pointer(io) >= 0 && io->state == SND_PCM_STATE_DRAINING) {
-			if (poll(&pfd, 1, -1) == -1) {
-				if (errno == EINTR)
+	/* We must ensure that all remaining frames in the ring buffer are flushed
+	 * to the FIFO by the I/O thread. It is possible that the client has called
+	 * snd_pcm_drain() without the start_threshold having been reached, or
+	 * while paused, so we must first ensure that the IO thread is running. */
+	if (bluealsa_start(io) < 0)
```
arkq, line 681: In case of `bluealsa_start()` failure the error is not returned intentionally? If not, maybe it could be a good idea to return error from drain function as well?

arkq (approved):

src/asound/bluealsa-pcm.c
borine: The current code simply copies the ioplug built-in drain behaviour by returning zero for any error - ioplug then always calls snd_pcm_ioplug_drop().

If we return an error code, then ioplug will return that code to the client, without changing the PCM state or dropping the stream. So we would need to determine and set the correct state for each error, and possibly also drop the stream if that is appropriate.

borine:
Hmm - the test for io->state == SND_PCM_STATE_SUSPENDED in that last commit was silly; there is no way the system can have set that state. Do you know if there is any way to test if the system has been suspended during a call to poll()? If not then perhaps we should just drop the stream and report -EIO for any timeout error.

arkq:
> Do you know if there is any way to test if the system has been suspended during a call to poll()

I have no idea how to do it properly... the only thing that comes to my mind is some heuristic based on wall clock time vs monotonic time check. But I guess that simple EIO will suffice here.

arkq (approved):

PR #669, opened 2023-10-06
Alsa plugin resilience

borine:
Some more tweaks to the PCM plugin to help isolate it from failures in the server.
Also includes a fix for a D-Bus test failure (but it seems there is still at least one other bug in there that causes occasional failures in github actions).

### DIFF
--- a/.github/spellcheck-wordlist.txt
+++ b/.github/spellcheck-wordlist.txt
@@ -146,6 +146,7 @@ dpsnk
 dpsrc
 DYN
 EAGAIN
+EINTR
 EINVAL
 ENODEV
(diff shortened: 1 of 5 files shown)


arkq:
> Also includes a fix for a D-Bus test failure (but it seems there is still at least one other bug in there that causes occasional failures in github actions).

It fails because `g_test_dbus_down(conn->dbus)` hangs... and to be honest I do not know why. It seems that the problem is in the g_test_dbus itself (based on issues found on the Internet).

arkq (approved):

arkq (approved):

PR #672, opened 2023-11-12
Add support for LHDC V3 A2DP source and sink

### LINKED ISSUE #588: LHDC support
There are both decoder and encoder ARM libraries for [AOSP](https://gitlab.com/savitech-lhdc/savitech-lhdc-codec-aosp-q100r20/-/tree/lhdc_aosp_q100r20_main/external).
AFAIK they work for V2, V3 and V4. There's also [AOSP 12.1](https://gitlab.com/savitech-lhdc/savitech-lhdc-codec-aosp-12.1.0r1/-/tree/lhdc_aosp_1210r1_main) repo with V5, but libraries are not there.

anonymix007:
Closes #588

~~Source does not work, only some buzzing noise is produced instead of music. I've tried saving pcm.data as raw pcm and it seems to be incorrect already.
Library expects interleaved 24 or 16 bit data depending on selected bit depth, so probably related to the use of BA_TRANSPORT_PCM_FORMAT_S24_4LE.~~ It seems to be somewhat working now, but not yet perfect.


anonymix007:
I still have no idea why this happens even with BA_TRANSPORT_PCM_FORMAT_S32_4LE. Apparently, some variables in `a2dp_lhdc_enc_thread` aren't what I thought. liblhdcBT_enc.so built from same sources works fine in Android. PCM dump appears to have same issues as sound in headphones. liblhdcBT_enc.so built from same sources works fine in Android, so I can't see what's wrong.

arkq:
I'd like to test it locally. From where I can get the ldhcBT-dec/enc libraries? If this PR were to go to master I need to have a possibility to test the code for potential regressions.

src/a2dp-lhdc.c
```diff
@@ -0,0 +1,460 @@
...
+
+void a2dp_lhdc_init(void) {
+}
+
+void a2dp_lhdc_transport_init(struct ba_transport *t) {
+
+	const struct a2dp_codec *codec = t->a2dp.codec;
+
+	if (codec->dir == A2DP_SINK) {
+		t->a2dp.pcm.format = BA_TRANSPORT_PCM_FORMAT_S24_4LE;
```
arkq, line 122: Why sink and source use different bit-depth? From the A2DP config I can see that only 16 and 24 bits are supported...

anonymix007:
[libs.zip](https://github.com/arkq/bluez-alsa/files/13403536/libs.zip)
[headers.zip](https://github.com/arkq/bluez-alsa/files/13403544/headers.zip)

Source code for libraries will be available later.

```console
LHDC_INCLUDE_DIR=/path/to/headers

export LHDC_DEC_CFLAGS="-I$LHDC_INCLUDE_DIR/liblhdc/include -I$LHDC_INCLUDE_DIR/liblhdc/inc"
export LHDC_DEC_LIBS="-llhdcBT_dec"
export LHDC_ENC_CFLAGS="-I$LHDC_INCLUDE_DIR/liblhdcdec/include -I$LHDC_INCLUDE_DIR/liblhdcdec/inc"
export LHDC_ENC_LIBS="-llhdcBT_enc"


autoreconf --install
mkdir build && cd build
../configure --enable-aac --enable-lhdc --enable-debug
make
sudo make install
```

arkq (dismissed):

src/a2dp-lhdc.c
anonymix007: Decoder and encoder libraries have different APIs for producing/consuming samples, you may check headers for reference. I can change both to use S32_4LE, but internally encoder/decoder uses 16 or 24 bits depending on the configuration.

src/shared/a2dp-codecs.h
```diff
@@ -338,18 +340,28 @@
...
-#define LHDC_VENDOR_ID                  BT_COMPID_SAVITECH
-#define LHDC_CODEC_ID                   0x4C32
+#define LHDC_V2_VENDOR_ID               BT_COMPID_SAVITECH
+#define LHDC_V2_CODEC_ID                0x4C32
+
+#define LHDC_V3_VENDOR_ID               BT_COMPID_SAVITECH
+#define LHDC_V3_CODEC_ID                0x4C33
+
+#define LHDC_V5_VENDOR_ID               BT_COMPID_SAVITECH
+#define LHDC_V5_CODEC_ID                0x4C35
```
arkq, line 350: I've got some question about these. There is v1, v2, v3 and v5. The v4 is non existent? Or you just reversed only these 4 variants? I'm asking because I'd like to pull the updated codec configurations as a first commit (before pulling the rest of it).

My Huawei phone reports 2 codec which very much resembles LHDC v4 and v6:
```
$ a2dpconf -x 7d020000344c03060150
<hex:7d020000344c03060150> {
  vendor-id:32 = 0x27d [HUAWEI Technologies Co., Ltd.]
  vendor-codec-id:16 = 0x4c34
  data:32 = hex:03060150
}
```

```
$ a2dpconf -x 7d020000364c03060150
<hex:7d020000364c03060150> {
  vendor-id:32 = 0x27d [HUAWEI Technologies Co., Ltd.]
  vendor-codec-id:16 = 0x4c36
  data:32 = hex:03060150
}
```

anonymix007: V4 and LLAC are just flags for V3. 
> My Huawei phone reports 2 codec which very much resembles LHDC v4 and v6

Screenshot from 2023-12-31 02-21-12

arkq: Many thanks for explanation!

arkq:
In the last few days I've done long overdue refactoring around A2DP codecs. Now, the logic is kept mostly in one file, which should simplify maintenance and addition of new codecs. Due to extensive changes, I've rebased your work on top of current master. The code seems to compile, but I've got no time yet to test whether it works :D

Also, I've left some commented code (I'm not sure whether it will be required or not). In the A2DP configuration for LHDC you've added `#define LHDC_CHANNEL_MODE_STEREO 0x03` which suggests that the channel mode can be selected, is that true? If yes, the channel mode selection will need to be done right in `a2dp_lhdc_configuration_select()` and `a2dp_lhdc_configuration_check()` functions, otherwise, maybe the define should be removed?

anonymix007:
Most constants were directly copied from [AOSP patches](https://gitlab.com/savitech-lhdc/savitech-lhdc-codec-aosp-13.0.0_r2). This must be a remnant of that. Seems to be unused, so might be removed probably.

The commented out code seems to be for channel mode, so could be removed as well.

Another issue would be to implement LLAC/V3/V4 selection logic, which is basically the following:

1. If LLAC is supported and LLAC-friendly bitrate is selected (ABR or < 500 kbps) then default to 48 kHz and enable LLAC (V4 flag needs to be set to 0).
2. If LLAC is not supported then enable V4 if it's supported, otherwise go for V3.

At least sink seems to still work.

anonymix007:
I've tried to run it once again, but now it doesn't seem to work (even AAC doesn't work).
```
$ bluealsa-aplay
bluealsa-aplay: [8934] W: ../../../utils/aplay/aplay.c:1278: Couldn't get BlueALSA PCM list: Sender is not authorized to send message
```
or from root:
```
bluealsa-aplay: [9413] D: ../../../utils/aplay/aplay.c:876: Creating IO worker XX:XX:XX:XX:XX:XX
bluealsa-aplay: [9413] D: ../../../utils/aplay/aplay.c:1287: Starting main loop
bluealsa-aplay: [9414] D: ../../../utils/aplay/aplay.c:542: Opening BlueALSA source PCM: /org/bluealsa/hci0/dev_XX_XX_XX_XX_XX_XX/a2dpsnk/source
bluealsa-aplay: [9414] D: ../../../utils/aplay/aplay.c:568: Starting IO loop
bluealsa-aplay: [9414] D: ../../../utils/aplay/aplay.c:692: Opening ALSA playback PCM: name=default channels=2 rate=44100
bluealsa-aplay: [9414] W: ../../../utils/aplay/aplay.c:696: Couldn't open ALSA playback PCM: Open PCM: Host is down
```

Probably related to #681, though I haven't recompiled PipeWire, just renamed /usr/lib/spa-0.2/bluez5.

arkq:
> But it segfaults with LHDC

It seems that the problem is somewhere within encoder code. The reason of abort is heap corruption due to `memmove()` beyond allocated memory in `ffb_t bt` buffer. I've tested with new library and the old one, but the result is all the same:
```
==1263640== Invalid write of size 2                                                                                                                                                                                  
==1263640==    at 0x48529E3: memmove (in /usr/libexec/valgrind/vgpreload_memcheck-amd64-linux.so)                                                                                                                    
==1263640==    by 0x4F1C7E4: ??? (in /mnt/storage1t/Downloads/balsa/build/lhdc/liblhdcBT_enc.new.so)                                                                                                                 
==1263640==    by 0x125B5E: a2dp_lhdc_enc_thread (a2dp-lhdc.c:179)                                                                                                                                                   
==1263640==    by 0x5309AC2: start_thread (pthread_create.c:442)                                                                                                                                                     
==1263640==    by 0x539AA03: clone (clone.S:100)                                                                                                                                                                     
==1263640==  Address 0x5af146a is 1,002 bytes inside a block of size 1,003 alloc'd                                                                                                                                   
==1263640==    at 0x48487A9: malloc (in /usr/libexec/valgrind/vgpreload_memcheck-amd64-linux.so)                                                                                                                     
==1263640==    by 0x111C9C: ffb_init (ffb.c:26)                                                                                                                                                                      
==1263640==    by 0x125801: a2dp_lhdc_enc_thread (a2dp-lhdc.c:125)                                                                                                                                                   
==1263640==    by 0x5309AC2: start_thread (pthread_create.c:442)                                                                                                                                                     
==1263640==    by 0x539AA03: clone (clone.S:100)                                                                                                                                                                     
==1263640==                                                                                                                                                                                                          
==1263640== Invalid write of size 1                                                                                                                                                                                  
==1263640==    at 0x4852A13: memmove (in /usr/libexec/valgrind/vgpreload_memcheck-amd64-linux.so)                                                                                                                    
==1263640==    by 0x4F1C7E4: ??? (in /mnt/storage1t/Downloads/balsa/build/lhdc/liblhdcBT_enc.new.so)                                                                                                                 
==1263640==    by 0x125B5E: a2dp_lhdc_enc_thread (a2dp-lhdc.c:179)                                                                                                                                                   
==1263640==    by 0x5309AC2: start_thread (pthread_create.c:442)                                                                                                                                                     
==1263640==    by 0x539AA03: clone (clone.S:100)                                                                                                                                                                     
==1263640==  Address 0x5af1478 is 13 bytes after a block of size 1,003 alloc'd                                                                                                                                       
==1263640==    at 0x48487A9: malloc (in /usr/libexec/valgrind/vgpreload_memcheck-amd64-linux.so)                                                                                                                     
==1263640==    by 0x111C9C: ffb_init (ffb.c:26)                                                                                                                                                                      
==1263640==    by 0x125801: a2dp_lhdc_enc_thread (a2dp-lhdc.c:125)                                                                                                                                                   
==1263640==    by 0x5309AC2: start_thread (pthread_create.c:442)                                                                                                                                                     
==1263640==    by 0x539AA03: clone (clone.S:100) 
```

anonymix007:
I was testing the decoder, but encoder having the same issue is really strange. None of them should call `memmove` directly (except third-party submodules). I'll look into that.

Do you have any suggestions for coexisting with pipewire?

arkq:
> I was testing the decoder, but encoder having the same issue is really strange.

I've added a test code which encodes PCM and then decodes it, so due to encoder failure I was not able to check decoder, not yet :D

> Do you have any suggestions for coexisting with pipewire?

What do you mean "coexisting"? If you have BlueALSA and PipeWire running on the same host at the same time, then you should disable (somehow) Bluetooth support in PipeWire (I'm not sure whether renaming BlueZ plugin is sufficient, it might be though). Then, remove paired device and pair it again, so BlueZ will not reuse cached SDP information.

anonymix007:
> What do you mean "coexisting"? 

`bluealsa-aplay` doesn't play any sound: `Couldn't open ALSA playback PCM: Open PCM: Host is down`

Though, I guess, for debugging this issue I might just run that test.

arkq:
> bluealsa-aplay doesn't play any sound: Couldn't open ALSA playback PCM: Open PCM: Host is down

I guess that you have to use PCM playback which is not grabbed by PipeWire or use some playback PCM exposed by PipeWire itself (if there is any). However, I'm not PipeWire user (not yet), so I've got zero knowledge about PipeWire configuration. Anyway, the error code "Host is down" is kinda strange...

anonymix007:
Apparently, it doesn't like the following line:
```
memcpy(out_put, enc, handle->host_mtu_size);
```
And this is exactly what original library does.

So this, per say, is not a library error. Should be fixed now.

src/a2dp-lhdc.c
```diff
@@ -0,0 +1,516 @@
...
+
+	uint8_t seq_num = 0;
+
+	debug_transport_pcm_thread_loop(t_pcm, "START");
+	for (ba_transport_pcm_state_set_running(t_pcm);;) {
+
+		switch (io_poll_and_read_pcm(&io, t_pcm, &pcm)) {
+		case -1:
+			if (errno == ESTALE) {
+				ffb_rewind(&pcm);
```
arkq, line 157: Is there any function for flushing encoder internal buffers? If yes, it should be used here. Please, see other A2DP codecs for reference.

anonymix007: Nope, there are none.

arkq: Is there any internal state of the encoder? The use case looks like this: application sends PCM and then it calls "drop" call which should drop all previously written PCM samples (not necessarily transmitted over the Bluetooth link). Then, the application sends new samples and these new samples should be transmitted without any residual samples from the previous write. Most encoders keep some internal state which is based on some samples history. So, in order to have a clean start, this state needs to be flushed. Otherwise, in case of writing silence after the "drop" call and encoding it, the decoder will decode some non-zero samples. Some encoders do have flush API, but some need a trick to do it (e.g.: lc3plus).

anonymix007: There is a circular buffer, which accumulates samples. But there is no API to drop samples from it.

arkq: OK, so some workaround will have to be applied here.

arkq:
> So this, per say, is not a library error. Should be fixed now.

OK, now the test passes but only with the old version of the library. With the new one (UPDATED Mar 1, 2024) encoder produces only 78 bytes per packet, where the old one produces 270 bytes. It looks like the output is truncated before these 78 bytes matches with the first 78 bytes in the packet from the old encoder.

anonymix007:
You're right, there was a stupid bug. I've updated the libraries.

O2C14:
> *UPDATED Apr 5, 2024*
> [libs.zip](https://github.com/arkq/bluez-alsa/files/14885476/libs.zip)
> [headers.zip](https://github.com/arkq/bluez-alsa/files/14885524/headers.zip)
> 
> Source code for libraries will be available later.
> 
> ```console
> LHDC_INCLUDE_DIR=/path/to/headers
> 
> export LHDC_DEC_CFLAGS="-I$LHDC_INCLUDE_DIR/liblhdcdec/include -I$LHDC_INCLUDE_DIR/liblhdcdec/inc -L$LHDC_INCLUDE_DIR"
> export LHDC_DEC_LIBS="-llhdcBT_dec"
> export LHDC_ENC_CFLAGS="-I$LHDC_INCLUDE_DIR/liblhdc/include -I$LHDC_INCLUDE_DIR/liblhdc/inc -L$LHDC_INCLUDE_DIR"
> export LHDC_ENC_LIBS="-llhdcBT_enc"
> 
> autoreconf --install
> mkdir build && cd build
> ../configure --enable-aac --enable-lhdc --enable-debug
> make
> sudo make install
> ```

Where did you get the library liblhdcBT_dec.so from?

anonymix007:
> Where did you get the library liblhdcBT_dec.so from?

From the same place as liblhdcBT_enc.so. I compiled it from sources.

arkq:
@anonymix007 Do you have any plans of opensourcing the library? I'm OK with the library being closed-sorce and even commercial, but the case is that I'd like to merge this PR sooner or later because syncing it is a burden I'd like to avoid. However, I need some way of verifying whether the codec code compiles and the unit test passes. I'm mostly developing on arm64, so it would be more convenient for me to have arm64 libraries instead of x64 (sometimes I'm doing some checks on x64, so such version is also required). If you don't want to opensource the lib and you don't want to release arm64 builds to the public, could you at least send them to me (on my gmail address)?

If there will be no way for me to verify the code during further development of bluez-alsa I'm not going to merge it, because it would be impossible for me to maintain it... I hope that you understand that :)

anonymix007:
Yes, I'm going to publish all sources once V5 encoding library starts working. Maybe I should publish V2/V3/V4/LLAC already even in the current state, but I haven't decided yet. I kinda want to make them at ABI-compatible with latest official libraries (4.1.0). Encoder is more or less at parity with 4.0.6p1, while decoder is in a worse state being only a bit better than 4.0.2, but the latest official LHDC libraries aren't available anywhere.

Sure, uploaded aarch64 builds along with the same x86-64 versions to https://github.com/arkq/bluez-alsa/pull/672#issuecomment-1817877044. Please let me know if they work, these were compiled on aarch64 (X Elite) without any special flags.

I wasn't able to make bluez-alsa work at all last time I checked, so I shifted efforts to [pipewire](https://gitlab.freedesktop.org/anonymix007/pipewire/-/tree/lhdc) and [ExtA2DP](https://github.com/anonymix007/ExtA2DP) implementations and both worked fine, so you may use them for tests.

arkq:
> Please let me know if they work

I was able to compile LHDC support and run tests, but it seems that something is not right with the encoder/decoder itself. When running IO test, the encoder produced payloads with lengths: 266, 262, 250, 230, 226, 222, 222, 214 So, it's strange that payload length is not constant. After decoding it I have almost all silence...

I can see one warning reported by the library:
> WARNING: cal_frame_size_and_frames_in_packet: max_mtu_limit(989)

And yes, I've used MTU greater than that in the IO test, so maybe that's the case...

Making MTU smaller yields different warning:
> WARNING: cal_frame_size_and_frames_in_packet: max_mtu_limit(789)

I will have to verify with old x64 libraries to see whether the problem is on bluez-alsa side or maybe with the arm64 version, or in general with your new libraries...

arkq:
@anonymix007 I've been trying to figure out what is wrong with the encoder/decoder, so I've created a [simple recoder](https://github.com/arkq/bluez-alsa/blob/lhdc/test/lhdcrecode.c) which feeds audio through lhdc_encode and lhdc_decode. Unfortunately, the code does not work as expected (logic is copied from the src/a2dp-lhdc.c). If you have some time, could you have a look at it? Maybe you will find out what is wrong with the usage of LHDC API.

You can compile the code from my https://github.com/arkq/bluez-alsa/tree/lhdc branch as follows:

1. Make sure that you have libsndfile installed: `pkg-config sndfile --cflags`
2. Setup env vars required for your libs
3. Configure bluez-alsa with: `../configure --enable-test --enable-lhdc`
4. Compile tests: `make check TESTS=`
5. Run `./test/lhdcrecode some-2-channel-audio.ogg`  (audio is read via libsndfile, so mp3, wav and other formats are supported)
6. This program will produce `recoded.wav` (which should be the same as original, but unfortunately is not....)

I've check with valgrind and it reports lots of `Conditional jump or move depends on uninitialised value` warnings in your lib. But it might be due to invalid API usage.... For the record, I'm testing it on arm64.

arkq:
Also, I have another question. Is the decoder thread safe? I see that there is `int lhdcBT_dec_deinit_decoder(void);` which looks not-thread-safe (if it uses some static internal storage instead of e.g. `thread_local`).

anonymix007:
I'll take a look at these issues.

> Is the decoder thread safe

I'm quite sure it is not. However, there's not much I can do about it without changing ABI. This is what the original library does. I guess I can provide separate thread-safe API version if it's really needed.

anonymix007:
I wrote a few simple utils for testing my library, you may refer to them for the library usage.
These were verified to be working.
`lhdcenc.c`, `lhdcenc_stereo.c` are used for encoding (difference is that the latter uses `lhdcBT_encode_stereo` API) specially named .wav files.
`lhdcdec.c` is used for decoding of whatever is produced by `lhdcenc*.c`.
`lhdc-parser.c` is a simple yet functional LHDC V2/V3/V4 frame header parser. 
[lhdc-utils.zip](https://github.com/user-attachments/files/17739938/lhdc-utils.zip)

anonymix007:
Your test program is wrong in many ways, in fact, it doesn't even compile (though I tried compiling it alongside liblhdc, not in bluez-alsa):

```C
lhdcrecode.c: In function ‘main’:
lhdcrecode.c:186:50: error: passing argument 2 of ‘lhdcBT_encode_stereo’ from incompatible pointer type [-Wincompatible-pointer-types]
  186 |                 if (lhdcBT_encode_stereo(handle, pcm_ch1, pcm_ch2, encoded_data.data, &encoded, &frames) < 0) {
      |                                                  ^~~~~~~
      |                                                  |
      |                                                  int16_t * {aka short int *}
In file included from lhdcrecode.c:10:
liblhdc/inc/lhdcBT.h:359:58: note: expected ‘int *’ but argument is of type ‘int16_t *’ {aka ‘short int *’}
  359 | int lhdcBT_encode_stereo(HANDLE_LHDC_BT hLhdcParam, int *left_pcm, int *right_pcm, unsigned char *out_put, uint32_t *written, uint32_t *out_frames);
      |                                                     ~~~~~^~~~~~~~
lhdcrecode.c:186:59: error: passing argument 3 of ‘lhdcBT_encode_stereo’ from incompatible pointer type [-Wincompatible-pointer-types]
  186 |                 if (lhdcBT_encode_stereo(handle, pcm_ch1, pcm_ch2, encoded_data.data, &encoded, &frames) < 0) {
      |                                                           ^~~~~~~
      |                                                           |
      |                                                           int16_t * {aka short int *}
liblhdc/inc/lhdcBT.h:359:73: note: expected ‘int *’ but argument is of type ‘int16_t *’ {aka ‘short int *’}
  359 | ncode_stereo(HANDLE_LHDC_BT hLhdcParam, int *left_pcm, int *right_pcm, unsigned char *out_put, uint32_t *written, uint32_t *out_frames);
      |                                                        ~~~~~^~~~~~~~~

lhdcrecode.c:209:35: error: passing argument 1 of ‘lhdcBT_dec_decode’ from incompatible pointer type [-Wincompatible-pointer-types]
  209 |                 lhdcBT_dec_decode(&encoded_data, encoded_data_len, decoded_pcm, &decoded_len, bit_depth);
      |                                   ^~~~~~~~~~~~~
      |                                   |
      |                                   struct <anonymous> *
In file included from lhdcrecode.c:11:
liblhdcdec/inc/lhdcBT_dec.h:24:38: note: expected ‘const uint8_t *’ {aka ‘const unsigned char *’} but argument is of type ‘struct <anonymous> *’
   24 | int lhdcBT_dec_decode(const uint8_t *frameData, uint32_t frameBytes, uint8_t* pcmData, uint32_t* pcmBytes, uint32_t bits_depth);
      |                       ~~~~~~~~~~~~~~~^~~~~~~~~
lhdcrecode.c:209:68: error: passing argument 3 of ‘lhdcBT_dec_decode’ from incompatible pointer type [-Wincompatible-pointer-types]
  209 |          lhdcBT_dec_decode(&encoded_data, encoded_data_len, decoded_pcm, &decoded_len, bit_depth);
      |                                                             ^~~~~~~~~~~
      |                                                             |
      |                                                             int16_t * {aka short int *}

liblhdcdec/inc/lhdcBT_dec.h:24:79: note: expected ‘uint8_t *’ {aka ‘unsigned char *’} but argument is of type ‘int16_t *’ {aka ‘short int *’}
   24 | ode(const uint8_t *frameData, uint32_t frameBytes, uint8_t* pcmData, uint32_t* pcmBytes, uint32_t bits_depth);
      |                                                    ~~~~~~~~~^~~~~~~

```

With these warnings and a few other logic errors fixed, it appears to be working.
[0001-Fix-LHDC-recode.zip](https://github.com/user-attachments/files/17740481/0001-Fix-LHDC-recode.zip)
Screenshot from 2024-11-14 02-42-27
Screenshot from 2024-11-14 02-42-10

> When running IO test, the encoder produced payloads with lengths: 266, 262, 250, 230, 226, 222, 222, 214 So, it's strange that payload length is not constant. 

It is not. LHDC V2/V3/V4 are quite similar to LossyFLAC/LossyWavpack, all are by default VBR. Bitrate only defines the upper limit, actual bitrate might be lower. For example, it'll always encode "constant frames" into 12 bytes or something about it. I think you may set `need_padding` in lhdcBT_init_encoder to make it zero-pad the frames, but there isn't much sense in doing that for real use and I don't think this was extensively tested.

> I can see one warning reported by the library:
> WARNING: cal_frame_size_and_frames_in_packet: max_mtu_limit(989)

I guess I shouldn't have made this a warning... This is normal, library just reports the actual MTU used internally.

> For the record, I'm testing it on arm64.

And so did I. Fortunately, I now have a good arm64 machine so it's quite easy for me to test it.

> This program will produce recoded.wav (which should be the same as original, but unfortunately is not....)

I'm quite sure it won't be in most cases, though it *should* sound the same. LHDC is a lossy codec.

arkq:
> Your test program is wrong in many ways, in fact, it doesn't even compile

It might be, however, it should compile without `-Werror`. Anyway, thanks for sending your test encoder and decoder. I was able to compile it and they recode audio correctly (the implementation in the `lhdcrecode.c` still has some bug...). I will try to compare a2d-lhdc.c implementation with your apps to see what is not right.

I have more questions, though (in order to better understand the API). What is the purpose of the `bitPerSample` in `lhdcBT_init_encoder()` and `bits_depth` in `lhdcBT_dec_decode()`. I thought that it controls the bit-depth of PCM input/ouput samples (i.e. 16 := s16_le, 24 := s24_le (or s24_4le). But apparently I was wrong :D So, regardless of the bit-depth, the `lhdcBT_encode_stereo()` takes s32_le samples, and the `lhdcBT_dec_decode()` produces s24_4le. Is that right? I've been trying to change the  `bits_depth` in `lhdcBT_dec_decode()`, but it seems it does nothing at all...

arkq:
>> Is the decoder thread safe

> I'm quite sure it is not. However, there's not much I can do about it without changing ABI. This is what the original library does. I guess I can provide separate thread-safe API version if it's really needed.

The are some implication of that, though. BlueALSA runs a dedicated thread for encoder/decoder for every connected transport. So, in case when two A2DP source devices will try to use LHDC, bad things will happen... The encoder looks thread-safe, but it would be nice to know it for sure. As for decoder, I will have to add some restrictions, or maybe you can mark decoder global state as thread_local (C11), so at least for cases where there is a single decoder instance per-thread (like in bluealsa), there will be no side effects (with your copy of the library). I can also add some warning message which will warn user about the thread-safety concerns regarding the LHDC decoder library.

anonymix007:
> What is the purpose of the `bitPerSample` in `lhdcBT_init_encoder()` and `bits_depth` in `lhdcBT_dec_decode()`

`bitPerSample` in `lhdcBT_init_encoder()` actually controls bit depth of the encoded stream, but it's done internally (32-bit samples from `lhdcBT_encode_stereo()` are right-shifted and saved into internal buffers). As for `bits_depth` in `lhdcBT_dec_decode()`, it only controls output bit depth (so samples are left shifted) and there's no point in setting it to anything except 24 (though maybe I should add support for 32).
~~640 KiB is enough for everyone~~ `s24_4le` is enough for everyone.

> The encoder looks thread-safe, but it would be nice to know it for sure

Well, it probably isn't either. Worked fine in AOSP which also creates a separate thread for encoding though.

> As for decoder, I will have to add some restrictions, or maybe you can mark decoder global state as thread_local (C11)

There is more than variable for the global state and I don't really want to fix theoretical issue while there are more practical ones.

I might do something about this after the public release though, but for now let's assume that only 1 instance of LHDC encoder/decoder could be running in the same process and maybe add some runtime checks for that.

arkq:
>  it only controls output bit depth (so samples are left shifted) and there's no point in setting it to anything except 24 (though maybe I should add support for 32)

Thanks for the explanation. As for support for 32, I think that I will produce s32_le for both encoder and decoder anyway. It will allow easier internal testing (my entire testing infrastructure relies on decoder and encoder be able to consume/produce the same PCM sample format). So, for now I will add right-shift for produces 24-bit output.

> Worked fine in AOSP which also creates a separate thread for encoding though.

As far as I know Android is not able to stream audio to more than one remote device at a time. So, for Android that's not a problem.

arkq (approved):

arkq:
I was able to test the codec on my side. It seems that everything works as expected (within tested scope). The a2dp-lhdc.c code was OK after all, but the issue was with my test harness. However, after understanding how the codec API works I was able to fix that. Many thank for contribution!

Also as a side note. I've seen that the lhdcBT.h file contains commented out `lhdcBT_get_error_code(HANDLE_LHDC_BT handle)` function. It would be nice to have a way of retrieving any errors and print them to the user by the application instead of the library.

arkq:
@anonymix007 I have another question :D Is the decoder capable of decoding data encoded by devices with support for [LHDC-v2](https://github.com/arkq/bluez-alsa/tree/lhdc-v2) A2DP capabilities type? I have a Huawei phone which seems to support that. So, I've made an adaptation layer for LHDC-v2: https://github.com/arkq/bluez-alsa/tree/lhdc-v2 However, it seems that the codec payload is bit different than in case of [v3](https://github.com/arkq/bluez-alsa/blob/lhdc-v2/src/shared/a2dp-codecs.h#L620-L651). There is some header though, because second byte is a counter.

Here is an example of captured payload when playing some sound followed by a silence (without RTP header), maybe you will spot some syncword or something in these bytes :D :
```
10e0564c414c0402010160feffffccd98100cea82870b6b9320cc53fac61ff5451c50500aefb00e6019049821a1192ea4304d7621e1020501f1553f6f1c7d05b2f8f068181a91e8e3b42d3c8983870a92b57300032224dd734213a39149e11554cc0cebe0521b2b9302de01c708039651f64058588084731b195be945809a09c24107c0d436db77b58ed5fe326cf9ef3ddeca2e0e8db5f6c3a1543187d4b14c73402eb4a9055bc90127822219950fdec22115fd73187de3ea5412b6a4960b032d38f8f4f0424b25b216669b6c2a24e0aa2a2ba7fd75ad5871c9b4a777452d64897189aefe43e99bb20a5bf76525219c5a7797069bbf4534d73840b659e15a7ad704285b6ea2b74d21ba1922c959d4992244924499224b4e9a48010427b2b420821840821841021841042499200082449922492244992218424498410420810420821420821840821841000001042000c004c0000010100000000000c004c0000010100000000000c004c0000010100000000
10e1000c004c0000010100000000000c004c0000010100000000000c004c0000010100000000000c004c0000010100000000
0ce2000c004c0000010100000000000c004c0000010100000000000c004c0000010100000000
10e3000c004c0000010100000000000c004c0000010100000000000c004c0000010100000000000c004c0000010100000000
10e4000c004c0000010100000000000c004c0000010100000000000c004c0000010100000000000c004c0000010100000000
```

anonymix007:
> Is the decoder capable of decoding data encoded by devices with support for LHDC-v2 A2DP capabilities type

It should be capable, V3 is V2+lpc+different frame size and V4 is basically V3+LLAC. I definitely tested that once though, but something might've broke since then. 

> There is some header though, because second byte is a counter.

There is more than one header: first is 2 bytes (`rtp_lhdc_media_header_t`) and the second one is 5..7 bytes. So to "decode" the frame with `lhdc_parser.c`, you would need to strip first 2 bytes.

> syncword

`0x4C`, but it's usually not the first byte because frames are byte-swapped in groups of 4 after encoding. You may use `lhdc_parser.c` from `lhdc-utils.zip` to learn more about these frames.

arkq:
> It should be capable, V3 is V2+lpc+different frame size and V4 is basically V3+LLAC. I definitely tested that once though, but something might've broke since then.

OK, so something is not right, because only silence is decoded (and some random other frames).

Do you have some real data with encoded v3 (from some real retail device)? I'd like to see a single frame with RTP header to double check that lhdc media header is correct in bluez-alsa code base.

anonymix007:
Funnily enough, I don't have *real retail device* with LHDC. My phone doesn't support it out of the box (though is seems that LHDC is somewhat added in a newer version), so I'm using ExtA2DP. 

You may refer to the [official AOSP integration](https://gitlab.com/savitech-lhdc/savitech-lhdc-codec-aosp-13.0.0_r2/-/blob/AOSP13_LHDC_V5.2.5_DocV5.2.1/src/packages/modules/Bluetooth/system/stack/a2dp/a2dp_vendor_lhdcv3_encoder.cc?ref_type=tags#L960) to verify that payload header is correct. Looks good to me though.

Will capture a few frames a bit later.

arkq:
> You may refer to the [official AOSP integration](https://gitlab.com/savitech-lhdc/savitech-lhdc-codec-aosp-13.0.0_r2/-/blob/AOSP13_LHDC_V5.2.5_DocV5.2.1/src/packages/modules/Bluetooth/system/stack/a2dp/a2dp_vendor_lhdcv3_encoder.cc?ref_type=tags#L960) to verify that payload header is correct. Looks good to me though.

But looking [here](https://gitlab.com/savitech-lhdc/savitech-lhdc-codec-aosp-13.0.0_r2/-/blob/3ac44e21e92983bb800eee8cb9d938522e3852d2/src/external/liblhdcdec/src/lhdcBT_dec.c#L380-384) it seems that the LHDC 2-bytes header is in fact a byte header (number of frames + latency) followed by a byte sequence counter, which matches the header that I've got from my phone (which is a real retail device with LHDC v2 support). So [this structure](https://gitlab.freedesktop.org/anonymix007/pipewire/-/blob/2bbada1dc190db8df758621a00a53e7a917908ff/spa/plugins/bluez5/a2dp-codec-lhdc.c#L30) seems to be incorrect after all.

arkq:
If you have a device which supports that I'd be glad if you could dump Bluetooth traffic from your phone when streaming short (1-2 seconds) audio from your phone to headset which supports LHDC. You can use this https://www.wireshark.org/docs/man-pages/androiddump.html or any manufacturer specicif way to obtain HCI log from android. If you have a device with bluealsa, you can use this patch (applied on top of https://github.com/arkq/bluez-alsa/tree/lhdc-v2 branch) to capture incoming packets:

```patch
diff --git a/src/a2dp-lhdc.c b/src/a2dp-lhdc.c
index da394c4..5c67989 100644
--- a/src/a2dp-lhdc.c
+++ b/src/a2dp-lhdc.c
@@ -422,6 +422,8 @@ void *a2dp_lhdc_dec_thread(struct ba_transport_pcm *t_pcm) {
                if ((rtp_lhdc_media_header = rtp_a2dp_get_payload(rtp_header)) == NULL)
                        continue;

+               hexdump("PAYLOAD", rtp_header, len);
+
                int missing_rtp_frames = 0;
                rtp_state_sync_stream(&rtp, rtp_header, &missing_rtp_frames, NULL);

diff --git a/src/bluez.c b/src/bluez.c
index d4a0d69..d17332a 100644
--- a/src/bluez.c
+++ b/src/bluez.c
@@ -1367,6 +1367,7 @@ static void bluez_signal_interfaces_added(GDBusConnection *conn, const char *sen
                debug("Adding new Stream End-Point: %s: %s: %s",
                                batostr_(&addr), sep_cfg.type == A2DP_SOURCE ? "SRC" : "SNK",
                                a2dp_codecs_codec_id_to_string(sep_cfg.codec_id));
+               hexdump("SEP capabilities", &sep_cfg.capabilities, sep_cfg.caps_size);

                GArray *sep_cfgs = bluez_adapter_get_device_sep_configs(&bluez_adapters[dev_id], &addr);
                g_array_append_val(sep_cfgs, sep_cfg);
```

Compile bluez-alsa with debug enabled: `../configure --enable-debug --enable-lhdc` then run `bluealsad` with `-p a2dp-sink -c lhdc-v2 -c lhdc-v3` arguments. And finally post logs here. With bluealsa setup try to play audio when using LHDC v2 and LHDC v3.

arkq:
@O2C14 thanks for the dump! So, the header for v3 is the same as for v2, where the first byte is a "header" and the second one is a sequence.

arkq:
@O2C14 I've just updated the https://github.com/arkq/bluez-alsa/tree/lhdc-v2 branch, if you have a system with bluez-alsa on it could you verify that LHDC v3 works as expected? Use setup from https://github.com/arkq/bluez-alsa/pull/672#issuecomment-1817877044 to compile bluez-alsa with LHDC. Then run it as `bluealsad -p a2dp-source -c lhdc-v3`. Also, you can verify decoder part with `bluealsad -p a2dp-sink -c lhdc-v2 -c lhdc-v3` and `bluealsa-aplay -vv -D <PCM-for-audio-output>`.

EliasTheBigMasterofIssues:
Where i can find liblhdc source? or liblhdcv5 blob?

anonymix007:
Source code is not public yet. Prebuilt libraries are available in a message above: https://github.com/arkq/bluez-alsa/pull/672#issuecomment-1817877044

EliasTheBigMasterofIssues:
> Source code is not public yet. Prebuilt libraries are available in a message above: [#672 (comment)](https://github.com/arkq/bluez-alsa/pull/672#issuecomment-1817877044)

Thank you, i has tried on pipewire with your commits with disabling -wno-error=incompatible-point-*, and lhdcv3 is working , just lhdcv5 that is not identified because lacks a header to build time, but just lhdcv3 patches works very well

zeroninefourfour:
I tried play a 48K wav using aplay and mpv, theres no sounds but i see 
`WARNING: lhdc_encoder_adjust_bitrate: Down bitrate condition fails, new rate:320, current rate:320`
in bluezalisa

Has anyone encountered same problem?

zeroninefourfour:
bluealsad: [310303] D: ../../src/bluez.c:575: A2DP Source (LHDC v3) configured for device CC:AF:68:9B:64:DE
bluealsad: [310303] D: ../../src/bluez.c:578: A2DP selected configuration blob [len=9]: 3a050000334c140181
bluealsad: [310303] D: ../../src/bluez.c:580: PCM configuration: channels=2 rate=48000
bluealsad: [310303] D: ../../src/bluez.c:582: Delay reporting: unsupported
bluealsad: [310303] D: ../../src/bluez.c:748: Exporting media endpoint object: /org/bluez/hci0/A2DP/LHDCV3/source/3
bluealsad: [310341] D: ../../src/dbus.c:47: Called: org.bluealsa.PCM1.GetCodecs() on /org/bluealsa/hci0/dev_CC_AF_68_9B_64_DE/a2dpsrc/sink
bluealsad: [310341] D: ../../src/dbus.c:47: Called: org.bluealsa.PCM1.Open() on /org/bluealsa/hci0/dev_CC_AF_68_9B_64_DE/a2dpsrc/sink
bluealsad: [310303] D: ../../src/bluez.c:1485: Signal: org.freedesktop.DBus.Properties.PropertiesChanged(): org.bluez.MediaTransport1: State
bluealsad: [310341] D: ../../src/ba-transport.c:319: New A2DP transport: 18
bluealsad: [310341] D: ../../src/ba-transport.c:320: A2DP socket MTU: 18: R:1021 W:679
bluealsad: [310303] D: ../../src/ba-transport.c:1084: Starting transport: A2DP Source (LHDC v3)
bluealsad: [310303] D: ../../src/ba-transport-pcm.c:291: Created BT socket duplicate: [18]: 19
bluealsad: [310303] D: ../../src/ba-transport-pcm.c:370: Created new IO thread [ba-a2dp-lhdc]: A2DP Source (LHDC v3)
WARNING: cal_frame_size_and_frames_in_packet: max_mtu_limit(665)
bluealsad: [310342] D: ../../src/a2dp-lhdc.c:217: PCM IO loop: START: a2dp_lhdc_enc_thread: A2DP Source (LHDC v3)
bluealsad: [310303] D: ../../src/ba-transport-pcm.c:477: PCM resume: 14

Heres some llog

Issue #678, opened 2023-12-27
Fedora includes a fork of libopenaptx, libfreeaptx


lightdot:
Apparently Fedora decided to remove [libopenaptx](https://github.com/pali/libopenaptx) and include a fork, [libfreeaptx](https://github.com/iamthehorker/libfreeaptx) instead.

libfreeaptx 0.1.0 was a fork of libopenaptx 0.2.0.

Sadly from a programmer's or a packager's view, the package version and the file names were deliberately changed. Otherwise, the library is the same.

Would it be viable to add --with-libfreeaptx, with appropriate adjustments to the code, of course?

As to why this happened, there is some [discussion here](https://github.com/pali/libopenaptx/issues/13) and it can be researched further from there. I did not go into the details and have no opinion on the matter...

borine:
I've just submitted PR #679 as a proposal for adding libfreeaptx support (in addtion to libopenaptx support)

PR #679, opened 2024-01-21
Add support for libfreeaptx

borine:
Some distributions now package libfreeaptx for apt-X support (because pipewire uses it rather than libopenaptx). It appears that Fedora has **only** libfreeaptx, whereas Debian includes both libopenaptx and libfreeaptx.

See #678 for more information.


arkq (approved):
Many thanks for that!

PR #680, opened 2024-02-15
Command line option --loglevel

borine:
've been playing around with HFP recently, and discovered that my journal log was being spammed with thousands of `bluealsa: W: Missing mSBC packets` messages. This is not good for an embedded system with limited log space, so I thought it would be good to be able to filter out info and warning messages. This PR is my attempt to achieve that.


utils/aplay/aplay.c
```diff
@@ -1073,6 +1076,28 @@ int main(int argc, char *argv[]) {
 			syslog = true;
 			break;
 
+		case 9 /* --loglevel */ : {
```
arkq, line 1079: How about moving this to the second `getopt_long()`? In case of invalid value there is an `error()` call which will happen before the log is initialized. It's not a big issues, but maybe it will be better to call `error()` after `log_open()`.

arkq (dismissed):

arkq (approved):

arkq (approved):

arkq:
I've added support for bashcompletion and made small cleanup (after some consideration I don't think that reporting error from the `log_set_min_priority()` is required at all... so I've removed the check in that function, I hope that you don't mind).

Once again, many thanks for the patch!

PR #709, opened 2024-05-22
update dbus config documentation

borine:
Try to improve the description of locations for the bluealsa service d-bus policy file.

### DIFF
--- a/TROUBLESHOOTING.md
+++ b/TROUBLESHOOTING.md
@@ -14,8 +14,14 @@ starting the service may fail with the
 - The BlueALSA D-Bus policy file is not installed, or is in the wrong
 location.\
 In a default install, the file should be
+`/usr/share/dbus-1/system.d/bluealsa.conf`; but if BlueALSA was installed from
+a distribution package then the distribution may have moved it to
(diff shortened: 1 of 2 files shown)


arkq:
I've just checked files in `/usr/share/dbus-1/system.d/` on my Debian host, and there is a lot of files which names correspond to D-Bus service name. What do you think about renaming `bluealsa.conf` to `org.bluealsa.conf`?

See "INTEGRATING SYSTEM SERVICES" in: https://dbus.freedesktop.org/doc/dbus-daemon.1.html :

> The policy file for com.example.Example1 should normally be named com.example.Example1.conf.

borine:
> What do you think about renaming bluealsa.conf to org.bluealsa.conf?

I think that is the "correct" name. The only backwards compatibility issue I can see with changing the name is when an update results in two copies under different names (and also cross-release documentation such as the wiki, but that can be dealt with). So in principle I am in favour of such a change.

arkq:
> The only backwards compatibility issue I can see with changing the name is when an update results in two copies under different names

I think that in the next release we can go with exec renames (I'll submit PR proposal in few days maybe), so we can do more "destructive" stuff as well :D

PR #712, opened 2024-05-26
Rename bluealsa and bluealsa-cli execs

### LINKED ISSUE #587: [Question] Rename bluealsa and bluealsa-cli execs
It seems that in the Linux world it is very natural to name applications which "control" things as CTLs, e.g. `bluetoothctl`, `systemctl`. In the original PR for "bluealsa-client" https://github.com/Arkq/bluez-alsa/pull/413, there was some discussion how to name this tool and I opted for `bluealsa-cli`. Right now I can see that maybe that was a bad idea. So, I'm wondering how good/bad for community would be to change exec names....

I'm thinking about changing names for bluealsa service and bluealsa-cli (which right now seems more like "control" than "client"):

* `bluealsa` -> `bluealsad` (this will alight with `bluetoothd`, `ofonod`, and other system daemons)
* `bluealsa-cli` -> `bluealsactl` (this might be more nature for people who are familiar with systemctl and other CTLs)

I'd like to hear some opinion from anyone who uses this project in some way and from package maintainers (maybe there were some rename cases like that in other projects and what were the outcomes of such decisions).

arkq:
Closes #587


arkq:
@borine I think that I've finished all renames. Now, I need to figure out how to communicate this in a clear way, so everyone will be notified.

I've been thinking about writing a migration guide wiki page and explain in it what to do in order to migrate from v4.2.0 to v5.0.0. Also, maybe such migration guide should be linked on the project's frontpage (README), e.g. just on the top (maybe even before the header) in a "yellow warning frame" :D Also, I'm not sure how to migrate wiki. Should we add compatibility patch to all places where `bluealsa` system service is mentioned (and `bluealsa-cli`), or simply rename everything accordingly and add some asterisk with a footnote link to the migration guide.

What do you think about that?

I will wait a bit (maybe few weeks) before merging these changes into the master, so I will have a time to think about it carefully.

arkq:
> I suggest something like:

Thanks, that looks excellent!

borine:
Having read through the wiki pages, I think that some are easy to adapt and can be done as soon as this PR is merged, whereas others (the ones with scripts or other parts that will need testing) may need a little more time. So I suggest that the "easy" pages are modified off-line with something like this at the top of the page:
```markdown
> [!IMPORTANT]
> This document refers to BlueALSA components by the names used in the latest sources. For release v4.2.0 or earlier please note that:
> * The `bluealsad` daemon was called `bluealsa`
> * The `bluealsactl` utility was called `bluealsa-cli`
```
This renders as:

---------------------------------------------------------------------------

> [!IMPORTANT]
> This document refers to BlueALSA components by the names used in the latest sources. For release v4.2.0 or earlier please note that:
> * The `bluealsad` daemon was called `bluealsa`
> * The `bluealsactl` utility was called `bluealsa-cli`

---------------------------------------------------------------------------

For the other pages, the admonition could be:
```markdown
> [!IMPORTANT]
> This document refers to BlueALSA components by the names used in release v4.2.0 and earlier. For the current BlueALSA sources, note that:
> * The `bluealsa` daemon is now called `bluealsad`
> * The `bluealsa-cli` utility is now called `bluealsactl`
>
> Some elements have not been tested with BlueALSA more recent than release v4.2.0
```
rendering as:

-----------------------------------------------------------------------------
> [!IMPORTANT]
> This document refers to BlueALSA components by the names used in release v4.2.0 and earlier. For the current BlueALSA sources, note that:
> * The `bluealsa` daemon is now called `bluealsad`
> * The `bluealsa-cli` utility is now called `bluealsactl`
>
> Some elements have not been tested with BlueALSA more recent than release v4.2.0

# Using BlueALSA and PulseAudio Together

PulseAudio (https://www.freedesktop.org/wiki/Software/PulseAudio/) is the de-facto Linux standard audio server, particularly for desktop installations. Many linux distributions' desktops require PulseAudio in order to provide full functionality.

-----------------------------------------------------------------------------

Once the changes are agreed, I can apply them all in git on my local machine, then push them all together soon after the PR is merged.

arkq:
OK, it seems great. However, I think that it will have to be another release before this change. It would be a fixup release but since the opus support is already in master, I will tag version 4.3.0. In current form the release 4.2.0 fails to be installed on gentoo when tests are enabled. I'm trying to fix that now.

borine:
Just to complete this thread - I've updated (all but one of) the affected wiki pages to use the new file names, with an information panel to alert readers using 4.3.1 and earlier. I have not changed the yocto-oe page as I have never used that system and am I not sure what the correct changes should be.

arkq:
I think that yocto can be moved to archive or dropped completely. Bluealsa is available in the official openembeded index: https://layers.openembedded.org/layerindex/recipe/189320/

borine:
OK, i've deleted the yocto/oe page.

PR #726, opened 2024-08-25
Use basename implementation from glib-2.0

### LINKED ISSUE #123: Unable to get transport after 5e0d8a8
I am using BlueZ 5.49, alongside with alsa-lib 1.1.4.1 and bluez-alsa (for which I can't thank you enough for maintaining!!).
Both audio source and sink scenarios have been realised, and I am working to improve stability and audio delay regarding some remote devices. Everything works fine until commit 5e0d8a8, which sadly breaks the audio sink functionality, because the lib I'm writing suddenly can't seem to get the transport anymore, for e.g. reading the sample rate in order to set up the audio stream accordingly.

Is there anything you can recommend, may it be in general or specifically?
Let me know if you require additional information.
Kind regards!

kraj:
This is portable across various platforms and system C libraries e.g. basename differs between glibc and musl, where glibc provides a GNU implementation and POSIX version, and musl provides posix version only and Newer version of musl have removed prototype for basename in string.h [1] which caused build failures with gcc-14 on musl systems.

g_path_get_basename() is implemented in glib-2.0 and is portable across these systems. It allocates a new string for the returned value which should be free'd by user, however, all our usecases here are in main() functions, so technically we are not going to leak any memory.

[1] https://git.musl-libc.org/cgit/musl/commit/?id=725e17ed6dff4d0cd22487bb64470881e86a92e7

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -276,8 +276,7 @@ int main(int argc, char **argv) {
 			syslog = true;
 			break;
 		}
-
-	log_open(basename(argv[0]), syslog);
+	log_open(g_path_get_basename(argv[0]), syslog);
 
 	if (ba_config_init() != 0) {
 		error("Couldn't initialize bluealsa config");
--- a/test/mock/mock.c
+++ b/test/mock/mock.c
@@ -256,7 +256,7 @@ int main(int argc, char *argv[]) {
 			return EXIT_FAILURE;
 		}
 
-	log_open(basename(argv[0]), false);
+	log_open(g_path_get_basename(argv[0]), false);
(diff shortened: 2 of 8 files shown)


arkq:
Could you please verify whether simple solution like that will work: https://github.com/arkq/bluez-alsa/tree/basename ?

If possible I would like to avoid pulling in glib-2.0 dependency on bluealsa clients.

arkq:
> although posix basename does not guarantee that input string will be unmodified

With the bluez-alsa usage of `basename` it's guaranteed that the argument will not be modified.

Many thanks for reporting the issue!

PR #734, opened 2024-10-29
Enable dbus delay signals for internal processing delay.

borine:
I've raised this as a draft PR for discussion because:

1. I'm not sure if there is a reason why D-Bus signalling was omitted originally.
2. I'm not sure of the GIO dbus thread handling. This solution emits dbus signals from the transport I/O thread, but without any locking. Is that safe?
3. My choice of a 10ms difference threshold to limit the rate that signals can be emitted is entirely arbitrary with no specific evidence to justify it.

### DIFF
--- a/src/a2dp-aac.c
+++ b/src/a2dp-aac.c
@@ -30,6 +30,7 @@
 #include "ba-config.h"
 #include "ba-transport.h"
 #include "ba-transport-pcm.h"
+#include "bluealsa-dbus.h"
 #include "io.h"
 #include "rtp.h"
 #include "utils.h"
@@ -290,6 +291,7 @@ void *a2dp_aac_enc_thread(struct ba_transport_pcm *t_pcm) {
 
 	/* Get the delay introduced by the encoder. */
 	t_pcm->codec_delay_dms = info.nDelay * 10000 / rate;
+	ba_transport_pcm_delay_sync(t_pcm, BA_DBUS_PCM_UPDATE_DELAY);
 
 	rtp_header_t *rtp_header;
 	/* initialize RTP header and get anchor for payload */
@@ -377,6 +379,7 @@ void *a2dp_aac_enc_thread(struct ba_transport_pcm *t_pcm) {
 					if (!io.initiated) {
 						/* Get the delay due to codec processing. */
 						t_pcm->processing_delay_dms = asrsync_get_dms_since_last_sync(&io.asrs);
+						ba_transport_pcm_delay_sync(t_pcm, BA_DBUS_PCM_UPDATE_DELAY);
 						io.initiated = true;
 					}
 
@@ -556,6 +559,7 @@ void *a2dp_aac_dec_thread(struct ba_transport_pcm *t_pcm) {
 
 			/* Update the delay introduced by the decoder. */
 			t_pcm->codec_delay_dms = info->outputDelay * 10000 / rate;
+			ba_transport_pcm_delay_sync(t_pcm, BA_DBUS_PCM_UPDATE_DELAY);
 
 			/* update local state with decoded PCM frames */
 			rtp_state_update(&rtp, info->frameSize);
--- a/src/a2dp-aac.c
+++ b/src/a2dp-aac.c
@@ -1,6 +1,6 @@
 /*
  * BlueALSA - a2dp-aac.c
- * Copyright (c) 2016-2024 Arkadiusz Bokowy
+ * Copyright (c) 2016-2025 Arkadiusz Bokowy
  *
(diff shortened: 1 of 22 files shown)


arkq:
> I'm not sure if there is a reason why D-Bus signalling was omitted originally.

It was omitted due to lack of time to implement that :D Many thanks for that!

> I'm not sure of the GIO dbus thread handling. This solution emits dbus signals from the transport I/O thread, but without any locking. Is that safe?

Yes, it should be safe. Internally signal is just a D-Bus message, and messages are sent by `g_dbus_connection_send_message` which is thread safe.

> My choice of a 10ms difference threshold to limit the rate that signals can be emitted is entirely arbitrary with no specific evidence to justify it.

I'm going to use such delay for the client-delay signaling as well and it's also arbitrary decision (based on commersial headset delay report updates). However, during testing, I've also added time-rate limiting logic (https://github.com/arkq/bluez-alsa/pull/516/files#diff-8bfb2124e7db1e5681c384f69b807d2a22821e0db287cab13d9ea0eb000cff81R827) because during playback start I could unnecessarily flood other device with lots of delay reports (the 1s rate limit is also based on some commercial headset). It's implemented on the client side, but maybe it should be moved to the bluealsad... I'll have to check how this delay report signaling will work with your PR. I hope that the client-delay will be merged soon (after support on BlueZ side will fully land in upstream).

src/a2dp-aac.c
```diff
@@ -419,7 +419,7 @@ void *a2dp_aac_enc_thread(struct ba_transport_pcm *t_pcm) {
 			rtp_state_update(&rtp, pcm_frames);
 
 			/* update busy delay (encoding overhead) */
-			t_pcm->processing_delay_dms = asrsync_get_busy_usec(&io.asrs) / 100;
+			ba_transport_pcm_update_processing_delay(t_pcm, asrsync_get_busy_usec(&io.asrs) / 100);
```
arkq, line 422: Client delay support landed in master (if you are using BlueALSA as a speaker system for watching videos, maybe you can test it as well; I've tested it to some extend and it seems to work quite nice with BlueALSA<->BlueALSA setup, but I wonder how it will work with smartphones or other devices, unfortunately my Android does not support delay reporting... :/), so here instead of `ba_transport_pcm_update_processing_delay` I would use
```c
t_pcm->processing_delay_dms = asrsync_get_busy_usec(&io.asrs) / 100;
ba_transport_pcm_delay_sync(t_pcm, BA_DBUS_PCM_UPDATE_DELAY);
```
so in case of decoder thread, the remote Bluetooth device will be notified (if it supports delay reporting). And of course move threshold logic to `ba_transport_pcm_delay_sync()`. I hope that it will work that way.

borine: I've rebuilt bluez and bluez-alsa on my test pi zero w, both from their master branches, and am testing with bluez-alsa master on my laptop with firefox, and the audio sync is really excellent - no bluealsa delay adjustment applied on either device.

I will build bluez-alsa again from this branch and try again, probably at the weekend, although I think this branch is only useful when the sink does not support delay reporting, because otherwise Delay change signals from the sink will cause the source client to pick up the internal processing delay anyway.

arkq:
I can see that you've used `ba_transport_pcm_delay_sync` to notify everyone about delay change, but the `ba_transport_pcm_delay_sync` function does not have any hysteresis/throttling for sending D-Bus signals (it has that for updating BlueZ Delay property). I think that at least some kind of hysteresis might be required there otherwise we will DDoS D-Bus daemon :D

arkq (dismissed):

arkq:
During testing I've discovered one issue (thanks to this PR). The processing delay calculated in the encoder loop takes into account the BT write time, which in case of BT connection instability might increase much (up to BT connection timeout). So, before pushing this into the master I will have to think about changing the place where the delay is calculated (maybe it should be calculated before BT write, to account codec processing only, since it's named `processing_delay_dms` :D). The case is, that right now, when the BT socket is almost full, the "processing_delay_dms" value jitters beyond 10ms which still floods D-Bus. However, the logic in this PR is OK for me as it is.

arkq:
After some thinking about the encoder processing delay, I've ended up with a redesign of what it actually is. The summary of that is here: https://github.com/arkq/bluez-alsa/blob/b4cc4f18ecda69fa9d3b3f4cf58ef7d2cbfbaf08/src/a2dp-sbc.c#L263

I've rebased your PR on top of that.

arkq:
Due to redesign of processing delay calculation, I've simplified your changes a little bit. I think that now it's ready to go to master.

PR #745, opened 2024-12-20
Improve ALSA PCM plugin compatibility on a2dp-sink, hfp-hf and hsp-hs

### LINKED ISSUE #743: Force silent frames on PCM device when no audio is playing instead of closing it (to fix Jack + zita-a2j behavior)
### Feature description

As far as I understand, starting `bluealsad` with the "keep-alive" option like this ...

```sh
bluealsad -S --keep-alive=-1 -p a2dp-sink
```

... only keeps the Bluetooth transport active. If there's no audio playing from the external device, the PCM is closed and not even silent frames are written. This is a bit like an analog capture card which only starts working when you plug in an audio cable.

Most of the time this is totally fine and maybe it is even the desired behavior (e.g. efficiency, power consumption in embedded devices)?

However, there are scenarios where it is desired to have silent frames output. Currently, it's not possible to import `bluealsa` into Jack with `zita-a2j` reliably.

## Additional context

When there is no audio playing and you import `bluealsa` into Jack with `zita-a2j`, then once you start playing something, the audio is muffled and distorted.

Steps to reproduce:

1. Run bluealsa as a2dp-sink on the host device:

```sh
/usr/bin/bluealsa -S --keep-alive=5 -p a2dp-sink
```

2. Make the host device discoverable:

```sh
$ bluetoothctl --agent=NoInputNoOutput
system-alias TestDevice
discoverable on
pairable on
default-agent
```

3. Connect phone to host device called "TestDevice".

4. Start Jack server (the order for this doesn't matter, you can start Jack server before as well):

```sh
jackd -d alsa -P hw:1 -C hw:2 -i 2 -o 2 -r 44100 -p 64 -n 3 -D -S
```

5. Assuming no audio is playing from the phone, start `zita-a2j` and connect outputs:

```sh
zita-a2j -v -j bluealsa -d bluealsa -c 2 -r 44100 -p 4096 -n 3
jack_connect bluealsa:capture_1 system:playback_1 && jack_connect bluealsa:capture_2 system:playback_2
```

6. Start playing some music from the phone. The audio is muffled, distorted and you can barely hear anything.

However, if you start to play music **before** you start `zita-a2j`, then it is fine. BUT, if you pause the audio and resume after a few seconds, then you have the same problem.

You could argue that `zita-a2j` should also be aware of the PCM state, but I still think it's counterintuitive (at least for me), to not have silent frames when a device is connected but no audio is playing.

Anyone willing to provide ideas and insights on how this could be patched in bluez-alsa?

borine:
Permit a BlueALSA PCM to be used more like a `hw` PCM on a2dp-sink, hfp-hf and hsp-hs nodes by making the PCM capture delay value more useful, and by offering a choice of operating modes to suit different application requirements when a PCM exists but the stream does not offer any samples.

Closes #743


test/test-alsa-pcm.c
```diff
@@ -1200,34 +1222,36 @@ int main(int argc, char *argv[]) {
...
+
 	if (run_capture) {
 		TCase *tc = tcase_create("capture");
-		tcase_add_test(tc, dump_capture);
-		tcase_add_test(tc, test_capture_start);
-		tcase_add_test(tc, test_capture_drain);
-		tcase_add_test(tc, test_capture_pause);
-		tcase_add_test(tc, test_capture_overrun);
-		tcase_add_test(tc, test_capture_poll);
+		tcase_add_loop_test(tc, dump_capture, 0, hwcompat_count);
```
arkq, line 1229: I'm not sure how feasible it would be, but how about adding command line option to specify the `hwcompat` used for the test and instead of running tests in a loop, create bash script wrappers for running `test-alsa-pcm` with all values, i.e. `test-alsa-pcm-hwcompat-busy`, `test-alsa-pcm-hwcompat-silence`. The default would be running tests without hwcompat specified (built-in default). Then add these scripts to `TESTS` in `Makefile.am`. This should allow better parallelization, because right now this test runs for ~30s. Also, iterating over all possible values by default does not scale well (e.g. in cases we will have anther option like that in the future).

borine: I've tried to implement your suggestions here, but I'm not sure that I have correctly defined the dependency between the scripts and the test-alsa-pcm executable. It seems to build and pass the tests on my test machine though.

arkq: > but I'm not sure that I have correctly defined the dependency between the scripts and the test-alsa-pcm executable

I don't think that any dependency is required here at all... Scripts do not need to be "rebuilt" after `test-alsa-pcm` changes. But maybe I'm wrong...

borine: The scripts are actually put in place as links by the configure script; I was thinking about ensuring that the test-alsa-pcm executable is build before the scripts are executed. Does automake ensure that all tests are built before the first test is run ?

arkq: > Does automake ensure that all tests are built before the first test is run ?

To be honest I'm not sure. Please that a look at: https://www.gnu.org/software/automake/manual/html_node/Scripts_002dbased-Testsuites.html

I think that current setup "ensures" that programs are built beforehand. However, it's not required to make links in the configure stage. The TESTS variable runs scripts from `srcdir` as well.

arkq: I've just checked the Makefile generated by autotools and the `check-TESTS` target depends on `check_PROGRAMS`, so it is guaranteed that PROGRAMS are built before tests are executed. So, we should be fine without defining any explicit dependencies for test scripts.

borine: OK, thanks for checking that. I will re-work the autotools config for the scripts as soon as I can, some time this week.

arkq: Don't worry, there is not much to rework, simply remove unnecessary bits. I will do that during review :) However, I have not reviewed the main part yet...

arkq: The alsa-pcm test does not pass sanitizer test, hence, it is in the XFAIL group, but newly added scripts are not. It's like a one line fix I guess.

However, speaking about the overall design. Long time ago (at the beginning of that project), the PCM plugin was based on the `transfer` snd_pcm_ioplug_callback_t callback. However, due to some issues with playback, I've changed the design to a dedicated thread responsible for data transfer. Such design is more complex and to be honest I'm not sure that it is ideal. Also, as far as I remember other projects (PulseAudio for instance) use `transfers` successfully (I guess). I remember that the capture was much easier to handle in case of transfer callback (at least in case where we do not care about timing - transfer was driven by FIFO polling). So, my question is, whether current design is good in your opinion (considering that you've been tinkering around ALSA a lot recently :D)? I know that you've spent a lot of time improving current design (I do not want to waste that, so all that will land in master for sure - I will just have to read it carefully to understand what is going on :D) but I still have some worries in my mind that we might be following the dead-end road which adds unnecessary complexity to the system...

borine: I have an implementation that uses the transfer call to get notification of writes and reads from the application, but still uses a dedicated i/o thread so that the period timing is independent of the application. I found that this approach enabled implementation of the `silence_threshold` and `silence_size`.

That code is on the [sw-params](https://github.com/borine/bluez-alsa/blob/sw-params/src/asound/bluealsa-pcm.c) branch of my fork if you would like to see it. That branch is not compatible with this PR, but can be rebased if you would like to use it in master.

As this PR is not urgent, perhaps we should  postpone work on it, investigate the use of `transfer()` first, and then come back to this PR once a decision on that is made?

If you want to do that, I can either:
- raise a (draft) PR from the `sw-params` branch, or
- create a new branch to extract only the `transfer()` changes and leave the rest of the sw-params code for some potential uture update.

arkq: I've just check your sw-params branch and to be honest I'm not sure.... I thought that using `transfer()` might (somehow magically :D) simplify the logic within the plugin, but it seems that it's not the case. So, for now I will stick with reviewing this PR. However, if you're OK, I will cherry-pick the stop-threshold feature from your sw-params. It seems to be simple enough that I will be able to review it with my currently limited availability :)

borine: The reason that pulse and pipewire seem simpler is that they use the client libraries of their respective servers, and all the thread management happens within those libraries. The advantage of the transfer() approach is, I suppose, that you get the chance to lock a mutex before writing to the ring buffer, and this appears to be enough to satisfy TSAN.

You are welcome to cherry-pick anything from my bluez-alsa fork. I have not yet found any application that uses **any** of the sw-params implemented there, so I have doubts over the usefulness of implementing them (other than just academic interest of course). Please let me know if you have found a "real" use case.

arkq: > The advantage of the transfer() approach is, I suppose, that you get the chance to lock a mutex before writing to the ring buffer, and this appears to be enough to satisfy TSAN.

Good point! OK, so maybe I will get that as well.

> Please let me know if you have found a "real" use case.

To be honest, most work that I'm doing here (from my point of view) is in a scope of academic research. My real use case was covered long time ago - streaming audio to one or two headsets (with `multi` plugin) over A2DP with `mpv`. Another real use case will be when I will buy a headset with LE audio. However, for now I'm using my old-and-still-working-fine Jabra headsets :D Anyway, if someone has an idea which might improve the project, and such improvement is within my maintenance capabilities, I'm glad to include such feature (even work on it to some extent if I'm able to setup testing environment). The stop-threshold feature seems to be a nice one and simple enough to include it here.

src/asound/bluealsa-pcm.c
```diff
@@ -1715,20 +1994,34 @@ SND_PCM_PLUGIN_DEFINE_FUNC(bluealsa) {
...
+	/* If the BT transport codec is not known (which means that PCM sampling
+	 * rate is also not known), we cannot construct useful constraints. */
 	if (pcm->ba_pcm.rate == 0) {
 		ret = -EAGAIN;
 		goto fail;
 	}
 
-	if (stream == SND_PCM_STREAM_CAPTURE)
+	/* HW compatible mode applies only to a2dp-sink, hfp-hf and hsp-hs. */
+	if (pcm->ba_pcm.transport & (BA_PCM_TRANSPORT_A2DP_SOURCE | BA_PCM_TRANSPORT_MASK_AG))
```
arkq, line 2005: I've just started reviewing the HW compat code and I have a question. Is this line correct? The documentation says that compatibility mode is for capture and playback as well...

> silence - Inserts silence for capture streams, or simply drops frames for playback streams [...]

borine: I've modified the line in question to allow silence mode on all roles, and tested with `alsaloop` on a HFP-AG node. Seems to be working well.

arkq (approved):

arkq:
Sorry for the veeeryyy long delay with this PR, but I had a lot of real life related chores lately...

Once again, many thanks for your contributions!

PR #749, opened 2025-01-17
Document the release policy

borine:
Help users to understand how development is organized.

### DIFF
--- a/README.md
+++ b/README.md
@@ -233,11 +233,33 @@ wrapper as follows:
 bluealsa-aplay -L
 ```
 
+## Release Policy
+
+The bluez-alsa project does not create nor distribute release software
+packages. However, at irregular intervals, when an important new feature or fix
+is considered stable then the `master` branch is tagged with a release version
+number. These tags are included only to mark reference baselines; their main
+purpose is to help Linux distribution package maintainers to relate their own
+package versioning schemes to the original codebase.
+
+A release tag is formed as `vN.N.N` where the first number indicates a change
+that is **not** backwards compatible (often a change to the D-Bus API), the
+second a feature addition that **is** backwards compatible, and the third a
+backwards compatible bug-fix or internal change. This scheme complies with the
+rules of [Semantic Versioning]. The most significant changes included in each
+tag are listed in the [NEWS] file. For more detail on the changes please
+consult the [commit history].
+
+All changes are applied only to the `master` branch. There are no release
+branches and no support for old release tags. Distributions are expected to
+maintain their own patches if they need to support a package based on a release
+tag without upgrading to the latest `master` branch source.
+
 ## Contributing
 
 This project welcomes contributions of code, documentation and testing.
 
-Please see the [CONTRIBUTING](CONTRIBUTING.md) guide for details.
+Please see the [CONTRIBUTING] guide for details.
 
 ## Bug reports, feature requests, and requests for help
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -233,11 +233,32 @@ wrapper as follows:
...
+## Release Policy
+
+The bluez-alsa project does not create nor distribute release software
+packages. However, at irregular intervals, when an important new feature is
+considered stable then the `master` branch is tagged with a release version
+number. These tags are included only to mark reference baselines; their main
+purpose is to help Linux Distribution package maintainers to relate their own
+package versioning schemes to the original codebase.
+
+A release tag is formed as `vN.N.N` where the first number indicates a change
```
arkq, line 245: If possible I would add a link to https://semver.org/ somewhere.

arkq (approved):
That's awesome!

README.md
borine: done

README.md
```diff
@@ -253,13 +275,22 @@ issue will not discover the information in there.
 
 If reporting a problem as a new issue, please use the appropriate
 [bluez-alsa GitHub issue reporting template][] and complete each section of
-the template as fully as possible.
-
+the template as fully as possible. Note that the `Bug report` template should
+be used for all questions relating to a specific installation of BlueALSA, even
+if you are not sure that it is a bug (the name of the template is set by
```
arkq, line 280: I don't think that it's set by GitHub... It's here: https://raw.githubusercontent.com/arkq/bluez-alsa/refs/heads/master/.github/ISSUE_TEMPLATE/1-bug-report.md and we can change it (and the name of the file itself). If you feel that `Problem report` is better wording (I'm not a native speaker, so for me bug/problem/issue is all the same more or less :D) you can update the template.

README.md
```diff
@@ -274,3 +305,4 @@ file](LICENSE) for details.
 1. [RTP Payload Format for MPEG-4](https://tools.ietf.org/html/rfc6416)
 1. [Coding of MPEG-4 Audio](https://www.iso.org/standard/42739.html)
 1. [ALSA project library reference](https://www.alsa-project.org/alsa-doc/alsa-lib/index.html)
+1. [Semantic Versioning]
```
arkq, line 308: I don't think that it's required here. I've been trying to keep resources related to the functionality of the project here. Semantic versioning is not specific to bluez-alsa, IMHO all projects should comply with it! :)

README.md
borine: I've had a look around some other projects and it seems that "Bug report" is very common, so perhaps that is best left as it is. I've removed the comment about github.

I'm thinking maybe the blank template should be removed, since so many issues are created with this when really the bug report template would have been more useful?

README.md
borine: done.

README.md
arkq: I'm not sure whether the blank one can be removed. I don't think that it's configured anywhere...

borine: After some more thought, perhaps the blank template should remain, at least until there has been time to assess whether the new README text is helping. If it is ever decided to remove it, then I think there must be a way, because there are some other projects that have done so.

arkq: Indeed it's possible to disable that blank template: https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository#configuring-the-template-chooser

arkq (approved):

arkq:
I've just squashed all the commits.

Many thanks for that!

PR #750, opened 2025-02-02
Aplay resampler

borine:
This is my attempt to address the topics previously discussed in PR #459

The libsamplerate resampler causes high CPU usage. Its not so bad on x86_64, but really very high on armhf. I have not tried with arm64. Having said that, I have tested on an old pi zero W (32bit armv6) with the "SINC_FASTEST" converter and got very stable audio with accurate delay reporting watching a 4 hour video stream. `top` showed bluealsa-aplay consuming 28% of the (single core) CPU but the pi continued to run smoothly. So anyway this is left as an option and not included in the build by default.

### DIFF
--- a/utils/aplay/alsa-pcm.c
+++ b/utils/aplay/alsa-pcm.c
@@ -11,6 +11,7 @@
 #include "alsa-pcm.h"
 
 #include <errno.h>
+#include <stdbool.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
@@ -19,9 +20,12 @@
 
 static int alsa_pcm_set_hw_params(
 		struct alsa_pcm *pcm,
-		snd_pcm_format_t format,
+		snd_pcm_format_t format_1,
+		snd_pcm_format_t format_2,
+		snd_pcm_format_t *format,
 		unsigned int channels,
-		unsigned int rate,
+		unsigned int *rate,
+		bool exact_rate,
 		unsigned int *buffer_time,
 		unsigned int *period_time,
 		char **msg) {
@@ -34,6 +38,7 @@ static int alsa_pcm_set_hw_params(
 	int err;
 
 	snd_pcm_hw_params_alloca(&params);
+	*format = SND_PCM_FORMAT_UNKNOWN;
 
 	if ((err = snd_pcm_hw_params_any(snd_pcm, params)) < 0) {
 		snprintf(buf, sizeof(buf), "Set all possible ranges: %s", snd_strerror(err));
@@ -41,12 +46,27 @@ static int alsa_pcm_set_hw_params(
 	}
 
 	if ((err = snd_pcm_hw_params_set_access(snd_pcm, params, access)) != 0) {
-		snprintf(buf, sizeof(buf), "Set assess type: %s: %s", snd_strerror(err), snd_pcm_access_name(access));
+		snprintf(buf, sizeof(buf), "Set assess type: %s: %s",
+				snd_strerror(err), snd_pcm_access_name(access));
 		goto fail;
 	}
 
-	if ((err = snd_pcm_hw_params_set_format(snd_pcm, params, format)) != 0) {
-		snprintf(buf, sizeof(buf), "Set format: %s: %s", snd_strerror(err), snd_pcm_format_name(format));
+	/* Prefer the first format if it is supported by the device. */
+	if ((err = snd_pcm_hw_params_set_format(snd_pcm, params, format_1)) == 0)
+		*format = format_1;
+	/* Otherwise try second format. */
+	else if (format_2 != SND_PCM_FORMAT_UNKNOWN) {
+		if ((err = snd_pcm_hw_params_set_format(snd_pcm, params, format_2)) == 0)
+			*format = format_2;
+		else {
+			snprintf(buf, sizeof(buf), "Set format: %s: %s and %s",
+					snd_strerror(err), snd_pcm_format_name(format_1), snd_pcm_format_name(format_2));
+			goto fail;
+		}
+	}
+	else {
+		snprintf(buf, sizeof(buf), "Set format: %s: %s",
+				snd_strerror(err), snd_pcm_format_name(format_1));
 		goto fail;
 	}
 
@@ -55,8 +75,13 @@ static int alsa_pcm_set_hw_params(
 		goto fail;
 	}
 
-	if ((err = snd_pcm_hw_params_set_rate(snd_pcm, params, rate, 0)) != 0) {
-		snprintf(buf, sizeof(buf), "Set sample rate: %s: %d", snd_strerror(err), rate);
+	dir = 0;
+	if (exact_rate)
+		err = snd_pcm_hw_params_set_rate(snd_pcm, params, *rate, dir);
+	else
+		err = snd_pcm_hw_params_set_rate_near(snd_pcm, params, rate, &dir);
+	if (err != 0) {
+		snprintf(buf, sizeof(buf), "Set sample rate: %s: %u", snd_strerror(err), *rate);
 		goto fail;
 	}
 
@@ -87,8 +112,7 @@ fail:
 
 static int alsa_pcm_set_sw_params(
 		struct alsa_pcm *pcm,
-		snd_pcm_uframes_t buffer_size,
-		snd_pcm_uframes_t period_size,
+		snd_pcm_uframes_t start_threshold,
 		char **msg) {
 
 	snd_pcm_t *snd_pcm = pcm->pcm;
@@ -103,16 +127,8 @@ static int alsa_pcm_set_sw_params(
 		goto fail;
 	}
 
-	/* Start the transfer when three periods have been written (or when the
-	 * buffer is full if it holds less than three periods. */
-	snd_pcm_uframes_t threshold = period_size * 3;
-	if (threshold > buffer_size)
-		threshold = buffer_size;
-
-	pcm->start_threshold = threshold;
-
-	if ((err = snd_pcm_sw_params_set_start_threshold(snd_pcm, params, threshold)) != 0) {
-		snprintf(buf, sizeof(buf), "Set start threshold: %s: %lu", snd_strerror(err), threshold);
+	if ((err = snd_pcm_sw_params_set_start_threshold(snd_pcm, params, start_threshold)) != 0) {
+		snprintf(buf, sizeof(buf), "Set start threshold: %s: %lu", snd_strerror(err), start_threshold);
 		goto fail;
 	}
 
@@ -136,7 +152,8 @@ void alsa_pcm_init(struct alsa_pcm *pcm) {
 int alsa_pcm_open(
 		struct alsa_pcm *pcm,
 		const char *name,
-		snd_pcm_format_t format,
+		snd_pcm_format_t format_1,
+		snd_pcm_format_t format_2,
 		unsigned int channels,
 		unsigned int rate,
 		unsigned int buffer_time,
@@ -146,6 +163,10 @@ int alsa_pcm_open(
 
 	char *tmp = NULL;
 	char buf[256];
+	unsigned int actual_buffer_time = buffer_time;
+	unsigned int actual_period_time = period_time;
+	unsigned int actual_rate = rate;
+	const bool exact_rate = !(flags & SND_PCM_NO_AUTO_RESAMPLE);
 	int err;
 
 	if ((err = snd_pcm_open(&pcm->pcm, name, SND_PCM_STREAM_PLAYBACK, flags)) != 0) {
@@ -153,9 +174,8 @@ int alsa_pcm_open(
 		goto fail;
 	}
 
-	unsigned int actual_buffer_time = buffer_time;
-	unsigned int actual_period_time = period_time;
-	if ((err = alsa_pcm_set_hw_params(pcm, format, channels, rate,
+	if ((err = alsa_pcm_set_hw_params(pcm, format_1, format_2,
+				&pcm->format, channels, &actual_rate, exact_rate,
 				&actual_buffer_time, &actual_period_time, &tmp)) != 0) {
 		snprintf(buf, sizeof(buf), "Set HW params: %s", tmp);
 		goto fail;
@@ -167,7 +187,13 @@ int alsa_pcm_open(
 		goto fail;
 	}
 
-	if ((err = alsa_pcm_set_sw_params(pcm, buffer_size, period_size, &tmp)) != 0) {
+	/* Start the transfer when three requested periods have been written (or
+	 * when the buffer is full if it holds less than three requested periods. */
+	snd_pcm_uframes_t start_threshold = ((snd_pcm_uframes_t)period_time * 3 / 1000) * (rate / 1000);
+	if (start_threshold > buffer_size)
+		start_threshold = buffer_size;
+
+	if ((err = alsa_pcm_set_sw_params(pcm, start_threshold, &tmp)) != 0) {
 		snprintf(buf, sizeof(buf), "Set SW params: %s", tmp);
 		goto fail;
 	}
@@ -177,15 +203,15 @@ int alsa_pcm_open(
 		goto fail;
 	}
 
-	pcm->format = format;
 	pcm->channels = channels;
-	pcm->sample_size = snd_pcm_format_size(format, 1);
-	pcm->frame_size = snd_pcm_format_size(format, channels);
-	pcm->rate = rate;
+	pcm->sample_size = snd_pcm_format_size(pcm->format, 1);
+	pcm->frame_size = snd_pcm_format_size(pcm->format, channels);
+	pcm->rate = actual_rate;
 	pcm->buffer_time = actual_buffer_time;
 	pcm->period_time = actual_period_time;
 	pcm->buffer_frames = buffer_size;
 	pcm->period_frames = period_size;
+	pcm->start_threshold = start_threshold;
 	pcm->delay = 0;
 	pcm->hw_avail = 0;
 
@@ -231,7 +257,7 @@ int alsa_pcm_write(
 
 	snd_pcm_sframes_t frames = ffb_len_out(buffer) / pcm->channels;
 	snd_pcm_uframes_t hw_avail = pcm->buffer_frames - avail;
-	snd_pcm_sframes_t written_frames = 0;
+	snd_pcm_uframes_t written_frames = 0;
 
 	if (!drain) {
 
--- a/utils/aplay/aplay.c
+++ b/utils/aplay/aplay.c
@@ -43,6 +43,9 @@
 #include "alsa-pcm.h"
 #include "dbus.h"
 #include "delay-report.h"
+#if WITH_LIBSAMPLERATE
+# include "resampler.h"
+#endif
 
 /* Many devices cannot synchronize A/V with very high audio latency. To keep
  * the overall latency below 400ms we choose default ALSA parameters such that
@@ -94,6 +97,9 @@ static bdaddr_t *ba_addrs = NULL;
 static size_t ba_addrs_count = 0;
 static unsigned int pcm_buffer_time = 0;
 static unsigned int pcm_period_time = 0;
+#if WITH_LIBSAMPLERATE
+static enum resampler_converter_type resampler_method = RESAMPLER_CONV_NONE;
+#endif
 
 /* local PCM muted state for software mute */
 static bool pcm_muted = false;
@@ -412,6 +418,40 @@ static void io_worker_mixer_event_callback(void *data) {
 	io_worker_mixer_volume_sync_ba_pcm(worker, &worker->ba_pcm);
 }
 
+static snd_pcm_uframes_t io_worker_playback_delay(
+		const struct io_worker *w,
+#if WITH_LIBSAMPLERATE
+		const struct resampler *resampler,
+		const ffb_t *resampler_buffer,
+#endif
+		const ffb_t *read_buffer) {
+
+	snd_pcm_uframes_t delay = 0;
+	double rate_ratio = 1.0;
+
+	unsigned int ba_pcm_buffered = 0;
+	/* Get the delay due to BlueALSA PCM FIFO buffering. */
+	ioctl(w->ba_pcm_fd, FIONREAD, &ba_pcm_buffered);
+	delay += ba_pcm_buffered / read_buffer->size / w->ba_pcm.channels;
+
+	/* Get the delay accumulated in the read buffer. */
+	delay += ffb_len_out(read_buffer) / w->ba_pcm.channels;
+
+#if WITH_LIBSAMPLERATE
+	if (resampler != NULL) {
+		rate_ratio = resampler_current_rate_ratio(resampler);
+		/* Get the delay accumulated in the resampler. */
+		delay += ffb_len_out(resampler_buffer) / w->ba_pcm.channels / rate_ratio;
+	}
+#endif
+
+	/* Get the ALSA device delay. In case when resampler is used, convert
+	 * ALSA device sample rate to the BlueALSA PCM sample rate. */
+	delay += w->alsa_pcm.delay / rate_ratio;
+
+	return delay;
+}
+
 static void io_worker_routine_exit(struct io_worker *w) {
 
 	pthread_mutex_lock(&w->mutex);
@@ -438,20 +478,48 @@ static void *io_worker_routine(struct io_worker *w) {
 	const snd_pcm_format_t pcm_format = bluealsa_get_snd_pcm_format(&w->ba_pcm);
 	const ssize_t pcm_format_size = snd_pcm_format_size(pcm_format, 1);
 	const size_t pcm_1s_samples = w->ba_pcm.rate * w->ba_pcm.channels;
-	ffb_t buffer = { 0 };
+	/* Buffer for audio frames read from the BlueALSA server. */
+	ffb_t read_buffer = { 0 };
+	/* Buffer from which audio frames are written to the ALSA PCM. If the
+	 * resampler is used then this is the resampler output buffer. Otherwise
+	 * it is the same as the read buffer. */
+	ffb_t *write_buffer = &read_buffer;
+	/* Preferred format for the ALSA PCM. If not using the resampler then this
+	 * is the format of the incoming BlueALSA stream. */
+	snd_pcm_format_t format_1 = pcm_format;
+	/* Alternative format that can be generated internally by the resampler.
+	 * This is only used if the resampler is enabled. */
+	snd_pcm_format_t format_2 = SND_PCM_FORMAT_UNKNOWN;
+
+#if WITH_LIBSAMPLERATE
+	/* The resampler requires the native endian format for the input data. */
+	const snd_pcm_format_t resampler_pcm_format = resampler_native_endian_format(pcm_format);
+	struct resampler resampler = { 0 };
+	ffb_t resampled_buffer = { 0 };
+	/* For detecting when the ALSA device has auto-started after reaching its
+	 * start threshold. */
+	bool alsa_pcm_started = false;
+	bool use_resampler = false;
+#endif
+
+	int pcm_flags = 0;
 
 	/* Cancellation should be possible only in the carefully selected place
 	 * in order to prevent memory leaks and resources not being released. */
 	pthread_setcancelstate(PTHREAD_CANCEL_DISABLE, NULL);
 
 	pthread_cleanup_push(PTHREAD_CLEANUP(io_worker_routine_exit), w);
-	pthread_cleanup_push(PTHREAD_CLEANUP(ffb_free), &buffer);
+	pthread_cleanup_push(PTHREAD_CLEANUP(ffb_free), &read_buffer);
+#if WITH_LIBSAMPLERATE
+	pthread_cleanup_push(PTHREAD_CLEANUP(resampler_free), &resampler);
+	pthread_cleanup_push(PTHREAD_CLEANUP(ffb_free), &resampled_buffer);
+#endif
 
-	/* Create a buffer big enough to hold enough PCM data for half the
-	 * requested PCM buffer time. This will be revised to match the actual
-	 * ALSA start threshold when the ALSA PCM is opened. */
-	const size_t nmemb = pcm_buffer_time * pcm_1s_samples / 1000000 / 2;
-	if (ffb_init(&buffer, nmemb, pcm_format_size) == -1) {
+	/* Create a buffer big enough to hold enough PCM data for three periods.
+	 * This will be later be revised if necessary to match the actual ALSA
+	 * start threshold when the ALSA PCM is opened. */
+	const size_t nmemb = ((size_t)pcm_period_time * 3 / 1000) * (pcm_1s_samples / 1000);
+	if (ffb_init(&read_buffer, nmemb, pcm_format_size) == -1) {
 		error("Couldn't create PCM buffer: %s", strerror(errno));
 		goto fail;
 	}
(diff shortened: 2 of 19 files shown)


arkq:
I'm currently in the review process. If you agree, I will push the refactoring bits to the master branch and keep only the resampling-related changes here. Please give me a day or two to examine the refactoring, then I will look at the resampler. Thank you very very much for this piece of work!

borine:
> If you agree, I will push the refactoring bits

Yes, please just take any pieces you feel would be useful.

arkq:
I've already took first 4 commits. I've applied some polishing to them, so unfortunately, your other commits will not apply cleanly on master. But please, do not worry, I will rebase this PR on top of master once I will review the "aplay: improve overrun/underrun handling" commit.

I've updated the delay report API a little bit. Now, it gets all delay components internally (ALSA PCM delay as well). I'm not sure whether it will be OK with the rest of your work, but from this single commit point of view it is cleaner that way. The rest is not changed other than some comments updates or code style updates.

arkq:
I just reviewed the "improve overrun/underrun handling" commit and it also looks OK.
I've pushed it here: https://github.com/arkq/bluez-alsa/tree/aplay-sync

I've made some minor changes, though:

1. I've removed `delay_report_reset(&dr);` from the "if" where samples are removed from the internal ffb buffer. The case is, that removing all accumulated delay values will cause delay report to be ~0 and then again accumulate to real value. I'm not sure that we want that.
2. I've changed the `debug("Read buffer overrun")` to warning, because I think that drops should be visible to user to indicate that something is not right.
3. The `worker->ba_pcm.running = pcm->running;` is done in `supervise_io_worker_start()` since this function is responsible for creating worker, so it can also maintain already created one.
4. I think that here was an error, so I've fixed it:
    ```c
    /* Pad the buffer with enough silence to restore it to the underrun
     * threshold. */
    size_t padding = (pcm->underrun_threshold - frames) * pcm->channels;
    if (verbose)
       info("Underrun imminent: inserting %zu silence frames", padding / pcm->channels);
    snd_pcm_format_set_silence(pcm->format, buffer->tail, padding);
    ffb_seek(buffer, padding);
    frames += padding;      // <---- the padding is in samples, not frames
    ```

Other than that, I've just rebased it on top of master.

I will test it tomorrow and if everything is OK I will push that to master (if you do not have any objections/remarks).

borine:
> removing all accumulated delay values will cause delay report to be ~0 

yes, not good from the client view. The same is also true of the reset after an ALSA underrun. When I put those lines in I was thinking only of the resampler, needing some way to indicate that the delay report needed time to settle on a new average before using to calculate a rate adjustment. I agree that clients will benefit by removing the reset() calls, and the resampler management needs a bit more work.

arkq:
I've rebased the resampler on top of master. However, I'm not sure whether I have not broken anything along the way... Also, I have not reviewed changes yet, it's only a bling rebase attempt. Over the weekend I will try to do some preliminary review of all these changes.

borine:
I found a bug in the already-merged alsa-pcm code: the written frames are counted twice when calculating the delay. commit 082f1bb fixes that, and I recommend it be merged before the resampler code.

commit 8670572 is also worth considering for immediate merge as it reduces the CPU load (slightly) and helps keep the delay report more stable.

There is at least one bug in the resampler (softvol muting may use the wrong sample format), and I hope to have a commit for that ready later today. Other than that the resampler is now looking good in my testing with iphone and bluealsa sources.

utils/aplay/aplay.c
```diff
@@ -681,20 +796,61 @@ static void *io_worker_routine(struct io_worker *w) {
...
+
+		if (alsa_pcm_write(&w->alsa_pcm, write_buffer, !w->ba_pcm.running, verbose) < 0)
 			goto close_alsa;
 
+		/* Set poll() timeout such that this thread is always woken before an
+		 * ALSA underrun can occur */
+		timeout = 1000 * w->alsa_pcm.hw_avail / w->alsa_pcm.rate;
+		/* poll() timeouts may be late by up to 2ms depending on the scheduler
+		 * and workload. So we allow for this when setting the timeout value. */
+		timeout -= 2;
```
arkq, line 820: Is it guaranteed that the timeout calculated a line above will be >= 2 (otherwise we will go into negative values)? Also, the `w->alsa_pcm.hw_avail` is updated after the `if (drain) { ... }` block and 0 is returned. In such case the `hw_avail` will not bw updated (jump to `close_alsa` is done only if `alsa_pcm_write` returns < 0).

borine: > Is it guaranteed that the timeout ... will be >= 2 ?

No, a fix for that is on its way.

> In such case the hw_avail will not be updated

I will set both the delay and hw_avail values to 0 after a drain, since they do not have any meaningful value when pcm state is SND_PCM_STATE_SETUP. In any case, we only drain when the ba_pcm becomes idle, so I will move the ba_pcm.running check before the timeout setting. New commit hopefully later today after testing.

arkq:
I've squashed all the commits (to simplify review outside of github) and rebased it on top of master.

borine:
Testing with a bigger range of source applications has shown that there is (at least) a couple of issues with this as it stands:
1. Inserting silence to prevent underrun can result in large delays if the source has regular breaks in the stream. I'm experimenting with using snd_pcm_drain() to stop the alsa device cleanly when there are no frames to play, and so far this is giving a better listener experience than padding the buffer with silence.
2. The resampler can become unstable when the source stream is unsteady, resulting in very wild swings in the rate adjustment. I'm experimenting with enforcing limits on the magnitude of the adjustment to control this.

I hope to have some updates next week if these new ideas prove useful.

Incidentally, one of my test sources is my multi-client branch of bluealsa with chromium, and this has proven to be particularly troublesome. I was planning to submit that as a PR soon, but it will have to wait longer as clearly there is (yet another) problem which must be solved first.

arkq:
I've renamed the configuration option to `--with-libsamplerate`. The case is that libsamplerate does not add any new functionality to the bluealsa-aplay (IMHO). By default there is a dummy approach of dropping/inserting frames. Also, many other projects use similar configuration option.

arkq:
#### Setup

Debian GNU/Linux 12 (bookworm)
libsamplerate0/stable,now 0.2.2-3 arm64

ALSA PCM sample rate: 48000 Hz
ALSA PCM channels: 2

#### CPU usage on RPi4 arm64

| METHOD | CPU |
|--------|--------|
| zero-hold | ~2 % |
| linear | ~2.2 % |
| none | ~6 % |
| fastest | ~6.0 % |
| medium | ~12.2 % |
| best* | ~59 %  |

* with load ~60% the connection quality has to be very good (short distance between devices, otherwise there are PCM underruns, then PCM drops)

Also, it's strange that `none` uses 6% CPU where `linear` uses only 2%....

And there is a segfault in case of PCM drop (not every time though):
```
bluealsa-aplay: 128.056381: [65653] W: ../../../utils/aplay/alsa-pcm.c:252: ALSA playback PCM underrun
bluealsa-aplay: 128.831193: [65653] W: ../../../utils/aplay/alsa-pcm.c:252: ALSA playback PCM underrun
bluealsa-aplay: 129.148652: [65653] W: ../../../utils/aplay/aplay.c:600: Dropping PCM frames: 6256

Thread 2 "bluealsa-aplay" received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7ff759f0c0 (LWP 65653)]
0x0000007ff7bd1b30 in src_float_to_short_array () from /lib/aarch64-linux-gnu/libsamplerate.so.0
(gdb) bt
#0  0x0000007ff7bd1b30 in src_float_to_short_array () from /lib/aarch64-linux-gnu/libsamplerate.so.0
#1  0x000000555555bdc0 in resampler_process (resampler=resampler@entry=0x7ff0047b40, in=in@entry=0x7ff759e110, out=out@entry=0x7ff759e130) at ../../../utils/aplay/resampler.c:188
#2  0x000000555555a860 in io_worker_routine (w=0x5555575620) at ../../../utils/aplay/aplay.c:807
#3  0x0000007ff79fee90 in start_thread (arg=0x7fffffe777) at ./nptl/pthread_create.c:442
#4  0x0000007ff7a67b1c in thread_start () at ../sysdeps/unix/sysv/linux/aarch64/clone.S:79
```

arkq:
> Inserting silence to prevent underrun can result in large delays if the source has regular breaks in the stream.

I'm not using `bluealsa-aplay` on regular basis, so I did not catch that earlier, but after same testing I can confirm that from the user experience point of view the logic prior to f0f9427da7 seems to be smoother in case of not stable BT connection.

borine:
> there is a segfault in case of PCM drop

It looks like I messed up big time there; I'll look into it at the weekend.

> Maybe bluealsa-aplay should behave in a similar way adaptive bit rate works. In case of not stable connection the delay (buffering) should be increased?

OK, so then we accept that larger delay is inevitable with poor connections. I had not considered that, but will think it over in the coming days. I the meantime I have an alternative which is to just stop the ALSA PCM when  the buffer runs empty and let it auto start again when the start threshold is achieved. I'll push a commit with that, hopefully tomorrow, to see what you think.

arkq:
> OK, so then we accept that larger delay is inevitable with poor connections.

IMHO, it will be better than constant underruns then drops.

There are two components here: a) clocks synchronization (long term drift and rather monotonous in one direction); b) connection instability causing delays and bursts in incoming PCM. So, now the question is for which component the resampler should help. IMHO, only for the case A. For case B, it might make things even worse (in extreme conditions). I do not know the features of libsamplerate, but it seems that we need to "guide" samplerate not to resample too fast. I do not know haw to call it, so I will give an example. Let say that we have 200ms of audio, but in order to be on-time, we have to shrink it to 100ms. So, such resampling should not be allowed, because it is a fix for case B, not for A. Of course, to some extent maybe we can fine-tune the sampling frequency even for local fluctuations, but if the problem is with delay/burst we will end up with lots of distortions. That's why I think that for case B, the proper solution would be to adjust buffer size not sampling frequency.

borine:
I've just pushed ~~2~~ 3 commits that were were written before your last comments.

[7e98498](https://github.com/arkq/bluez-alsa/pull/750/commits/7e98498aa6bce52bbf05bdcdcbf52d990d169871) uses drain instead of silence to deal with empty input buffer.

[6119a82](https://github.com/arkq/bluez-alsa/pull/750/commits/6119a82c41e3ce7e6af98977d0928d276a0789ba) limits the size of rate changes, and may help with filtering out delay changes due to lost audio frames and jitter introduced by the A2DP decoder.

The segfault is most likely still there, I haven't looked for that yet, and I will think about your other comments next week.

borine:
> There are two components here: a) clocks synchronization ...; b) connection instability ...

I think there is a third component: clock jitter, mostly caused by the A2DP encoding/transfer/decoding process.

In any case, I think we agree that a broken connection will inevitably result in broken audio. The only responsibility for bluealsa-aplay here is to not fall over and to be in a position to recover should the incoming stream later stabilize. So I think simply discarding frames or stopping the playback device is good enough, I do not see any need for dynamic buffer size adjustment. We do need to decide what is the appropriate condition for return to normal playback. I am now thinking that perhaps we should try to get the delay close to the target (i.e. near to the start threshold) before resuming normal behaviour, but currently don't know how that should be done.

So the remaining problem is how to identify the "drift" component when it is obfuscated by the (large) "jitter" component in the timing of the incoming data chunks. My current approach has no mathematical basis; the code simply steps the resampling rate up or down by an arbitrary amount when the reported average delay decreases or increases. There is no guarantee, or expectation, that the rate will ever stabilize to a constant value that adjusts only for timer drift. So there must be room for improvement there, but my knowledge of these things is very limited. I will continue to read, and experiment, to see if there is a (low complexity) way to do this better.

> maybe we should not use resampling all the time

That's an interesting idea. I don't plan to work on it right now, but perhaps you (or someone else) could develop it?

> a maintenance cost which should be taken into account

I absolutely agree. The reality is that bluealsa-aplay works well enough without any resampling in the majority of use cases. The only significant failure is when using dmix with the alsa-lib rate plugin internal resampler. But that is an issue for all applications, and a work-around is already documented in the bluealsa-aplay manual page. Other than that, a stream must run continuously for several hours before timer drift causes xruns and it appears that not many users do that. Perhaps now that delay reporting is available it might become more common to use bluealsa-aplay when watching movies or other long streams on a phone or tablet, but unless and until that happens we are unlikely to see many issues raised regarding xruns.

borine:
> with load ~60% the connection quality has to be very good (short distance between devices, otherwise there are PCM underruns, then PCM drops)

The drops, I believe, were caused by inserting silence even though (some of) the incoming frames were just late, and not completely lost. commit 7e98498 fixes that.

> Also, it's strange that none uses 6% CPU where linear uses only 2%....

I can only assume that resampling is happening in alsa-lib. Are you using the rate plugin internal converter, or the alsa-plugins speexrate or libsamplerate converters? (If the libasound2-plugins package is installed then by default the rate plugin uses the speexrate SRC_SINC_FASTEST converter).

> And there is a segfault in case of PCM drop (not every time though)

That is now fixed by commit b63c2db

utils/aplay/alsa-pcm.c
```diff
@@ -236,23 +261,23 @@ int alsa_pcm_write(
...
+			/* When the stream runs dry we drain the ALSA buffer and leave the
+			 * ALSA device stopped until fresh frames arrive from the server. */
+			debug("draining ALSA playback PCM to avoid underrun");
+			snd_pcm_drain(pcm->pcm);
+			snd_pcm_prepare(pcm->pcm);
+			pcm->delay = 0;
+			pcm->hw_avail = 0;
+			/* Flag an underrun to indicate that there has been a discontinuity
+			 * in the input stream. */
+			pcm->underrun = true;
```
arkq, line 276: Isn't `return 0;` missing here?

I've applied this commit on current master and I think I will push it to github. After some tests I'm not sure whether this approach is better or not, but it is definitely simpler (so if the effect is the same, simplicity wins). Also, during tests I'm seeing mostly debug log `ALSA playback PCM underrun` which is from the `snd_pcm_avail_delay()` check. So, it seems that the code responsible for inserting silent frames was rarely used anyway...

borine: > Isn't return 0; missing here?

Yes ! sorry :disappointed:

arkq: Nothing to be sorry, that's what for the review process is :D

Also, please check GitHub invitations. I'm not sure whether it will be of any use for you, but treat it like a gift for personal use.

borine: I'm now reducing my time spent on software development (due to health issues, I am an old man now). I will continue to contribute comments and documentation to bluez-alsa for as long as I am using it myself, but this PR and my multi-client branch will most likely be my last original code contributions (I also have some plugin code for ALSA swparams support, but there do not appear to be any applications that use stop_threshold, silence_threshold, etc so there is little benefit in maintaining it in master). So, thank you for your confidence in me to invite me to a new project, but unfortunately I have to decline.

arkq: > So, thank you for your confidence in me to invite me to a new project, but unfortunately I have to decline.

It's not a project which needs any development I think. You can use it with bluez-alsa as it is. But of course it's up to you.

Also, once again many thanks for your excellent contributions!

arkq:
I've just reviewed the code and it looks OK. Also, I've done some minor code polishing (mostly codestyle to fit with the rest of the project). I will try to test it for the next 2-3 days and if everything is all right I will merge it into master.

borine:
> I will try to test it for the next 2-3 days

I will also build with this latest code and install it today on my main music system, a raspberry pi 2B (4-core armv7 32-bit). I'll "live with it" for a few days real-life usage to see how things work out.

arkq:
Great. After my polishing the logic is (almost) exactly the same as in your original PR. I've changed minor things like:

- print warning if resampler will not be used: https://github.com/arkq/bluez-alsa/blob/2ba336ca4dcef1ea61fc0b03de176126b09f177e/utils/aplay/aplay.c#L498-L503
- use native format for silence only if resampler is used: https://github.com/arkq/bluez-alsa/blob/2ba336ca4dcef1ea61fc0b03de176126b09f177e/utils/aplay/aplay.c#L801-L807  (now I see that I should move the logic inside the `if (!w->alsa_mixer.has_mute_switch && pcm_muted)` if logic.
- use enums defined by libsamplerate: https://github.com/arkq/bluez-alsa/blob/2ba336ca4dcef1ea61fc0b03de176126b09f177e/utils/aplay/resampler.h#L23-L30
- use functions from `endian.h` for endian conversions: https://github.com/arkq/bluez-alsa/blob/2ba336ca4dcef1ea61fc0b03de176126b09f177e/utils/aplay/resampler.c#L426-L431
- use logic only for negative values (I've removed the `data[n] &= 0x00ffffff;` else logic, I think that it's not required, buuuuutttt I might be wrong..., that's the only part that I'm not 100% sure :D): https://github.com/arkq/bluez-alsa/blob/2ba336ca4dcef1ea61fc0b03de176126b09f177e/utils/aplay/resampler.c#L455-L459

arkq:
The input PCM comes from BlueALSA anyway, so for safety (interoperability with ALSA) BlueALSA should make sure that not used byte in S24_32 is always zeroed. We should be safe without this double check on bluealsa-aplay side.

arkq:
I've added some minor tweaks for logging ("Resetting resampler: _reason_"). Also, I moved the "Reset the resampler whenever the delay exceeds the limit" code block below the `if (resampler->target_delay == 0) { ... }` block, because in doesn't make sense to calculate `delay_diff = delay - resampler->target_delay;` if `target_delay` is not initialized.

However, there is still one issue. When there are PCM drops and then sudden influx of PCMs (due to connection instability), the resampler "locks" itself in a state like this:

```
aplay.c:602: Dropping PCM frames: 69
aplay.c:605: Resetting resampler: Dropped PCM frames
resampler.c:350: Adaptive resampling enabled: target delay = 172.9ms
aplay.c:861: Resampler rate ratio changed: delay=487.6ms ratio=1.088035
resampler.c:364: Resetting resampler: Delay difference limit exceeded: 13884 > 441
aplay.c:861: Resampler rate ratio changed: delay=487.7ms ratio=1.088435
resampler.c:350: Adaptive resampling enabled: target delay = 172.9ms
aplay.c:861: Resampler rate ratio changed: delay=487.2ms ratio=1.088035
resampler.c:364: Resetting resampler: Delay difference limit exceeded: 13872 > 441
aplay.c:861: Resampler rate ratio changed: delay=487.4ms ratio=1.088435
resampler.c:350: Adaptive resampling enabled: target delay = 172.9ms
aplay.c:861: Resampler rate ratio changed: delay=487.4ms ratio=1.088035
resampler.c:364: Resetting resampler: Delay difference limit exceeded: 13906 > 441
```

The delay target is set to 172.9ms where current delay is much higher, so the resampler resets itself in a loop. Since you understand the logic behind this feature better, maybe you will spot the issue faster then me :D (it's quite possible that I've broken something during "cleanups"....).

Also, sometimes when hitting `ctrl-c` to terminate bluealsa-aplay I can see some aborts (due to `free()` corruption), I'm not sure whether it's related to recent changes in bluealsa-aplay or maybe the bug was already there... I will check that with ASAN maybe later today.

borine:
> resampler resets itself in a loop

Yes, that's an error in my code. I have a fix which I will upload today

> some aborts (due to free() corruption)

I will look for them tomorrow

utils/aplay/resampler.c
```diff
@@ -0,0 +1,499 @@
...
+	snd_pcm_sframes_t delay_diff = delay - resampler->target_delay;
+	snd_pcm_uframes_t delay_diff_abs = labs(delay_diff);
+
+	/* Reset the resampler whenever the delay exceeds the limit. */
+	if (delay_diff_abs > resampler->max_delay_diff) {
+		/* Reset the resampler if not already done. */
+		if (is_timespec_zero(&resampler->reset_ts)) {
+#if DEBUG
+			if (resampler->target_delay != 0)
+				debug("Resetting resampler: Delay difference limit exceeded: %zu > %zu",
```
arkq, line 366: Why only the debug is under the `if`, but the actual `resampler_reset()` is not? Is that a typo or intended logic?

borine: The intention is to silence the debug message that would otherwise appear when the ALSA PCM first starts (ie when its start threshold is reached); we still need to reset the resampler there (to set the initial target delay), but I think that the "Delay difference limit exceeded" message is misleading in that special case. Perhaps an explanatory comment in the code is needed to emphasise that this code path applies both to the (normal) adaptive resampling initial start and the (exception) delay limit exceeded condition.

arkq:
With your last fix everything seems to be OK. To be honest I'm really impressed how well it works! :D

I'm going to test it a bit more in a quasi stable BT connection condition to check for any potential issues, but I think that the logic itself is solid! After some drops the target delay is reached within 1-2 minutes, so there are no audible audio distortions.

borine:
I have not managed to reproduce the abort(), so I have not changed anything to fix it. Are you sure it is not still hiding in there somewhere? Other than that, I am not yet 100% confident in the resampler start logic. I will be busy for the first half of next week, but then I want to work through as many code paths as possible just to try to identify any "corner cases" that may not be handled correctly.

arkq:
> I have not managed to reproduce the abort(), so I have not changed anything to fix it. Are you sure it is not still hiding in there somewhere?

I think that it is still there. Also, there is a leak because of `pthread_cleanup_push(PTHREAD_CLEANUP(resampler_destroy), resampler);`. When this is called, the `resampler` is still NULL, so the `resampler_destroy` is called with NULL. I will fix that.

arkq:
I've fixed the leak by keeping the resampler structure on the stack (it's not as elegant as before, but at least there will be one allocation less).

As for the free corruption, it seems that something is not right with the libsamplerate library, or with SRC_DATA passed to `src_process()` because it writes beyond `resampler->out_buffer` buffer... I'm using valgrind to detect it.

arkq:
OK, I've found it. The resampler was initialized with `read_buffer.nmemb` but it has to be initialized with `resampled_buffer.nmemb`.

arkq:
I've rewritten the internal processing buffer to be internally managed. I have not tested the FLOAT path (when the buffer is not used) - it seems that my sound card does not support floats....

But it seems that I've messed something up with the initialization... will push a fix in a moment.

arkq:
I've been testing it the whole day today and I'm happy how it works. I think that code "polishing" on my side is done. As for me, the feature can be merged to master. However, I will wait for you final approval (after checking corner cases, because I've been testing it only with Android source with various apps and RPi sink with linear converter).

borine:
I'm getting a segfault in libsamplerate src_process() when using 24-bit samples (aptx-hd). Haven't had time to debug this yet, but we are not yet ready for merge to master.

utils/aplay/aplay.c
```diff
@@ -555,7 +611,11 @@ static void *io_worker_routine(struct io_worker *w) {
 			if (ret % pcm_format_size != 0)
 				warn("Invalid read from BlueALSA source PCM: %zd %% %zd != 0", ret, pcm_format_size);
 
-			ffb_seek(&buffer, read_samples);
+#if WITH_LIBSAMPLERATE
+			if (use_resampler)
+				resampler_convert_to_native_endian_format(&read_buffer.tail, read_samples, pcm_format);
```
arkq, line 616:
```suggestion
				resampler_convert_to_native_endian_format(read_buffer.tail, read_samples, pcm_format);
```

This was causing segfault :D

arkq: Anyway there is still a bug with aptx-HD because the `/* convert to S32 */` is wrong. We need to shift `data[n] <<= 8;` otherwise audio is VEEEERRYY quiet. I will push these changes later.

borine:
commit 8d3de0a is not related to the aplay resampler, but fixes a long-standing memory management bug. To reproduce the error:
1. start bluealsa-aplay
2. connect device A
3. connect device B
4. start playback on device **B**
5. disconnect device A
6. re-connect device A

At this point the io_worker thread for device B is broken. Most often it remains blocked in poll(), sometimes it stops with  error "Bad file descriptor", and on rare occasions the whole process is terminated by an abort() in libsasound.

The error is caused by the worker state memory of device B being trashed when device A reconnects.

arkq:
>  but fixes a long-standing memory management bug

hmmm... the `static struct io_worker *workers = NULL` seems to be a relict from the very beginnings of this app...

Anyway, I think that we can (and should) simplify this even more (so it will be easier to read).
How about `static struct io_worker *workers[16] = { 0 }`, so every worker will be allocated separately?
The max number of supported simultaneous connections is 7 for BR/EDR anyway, and for now BlueALSA does not support BLE audio. With BLE this number might be bigger but realistically it will be rather low number (hence the number of workers can be statically allocated).

The current approach is still broken (even with your fix). The `if ((workers = realloc(workers, sizeof(*workers) * workers_size)) == NULL)` might reallocated the memory chunk, so the pointer passed to `io_worker_routine()` will point to deallocated memory...

borine:
> ` static struct io_worker *workers[16] = { 0 }`

Yes, or if dynamic allocation should ever be required, use a hash table or linked list so each worker can still have a fixed address. I think the static array is the best choice for an initial fix, at least until BLE audio is supported.

Would you like me to try to write a static array implementation next week, or do you already have some code in place?

borine:
I've just pushed a global static solution, no allocations.

arkq:
> I've just pushed a global static solution, no allocations.

Great! However, could you create a separate PR for that (not to clobber the resampler PR with not-related bug).

I've just checked your commit and I think that we need to "complicate" it a little bit :D Instead of having entire structure statically allocated, how about keeping only the list of pointers on the statically allocated array.
```c
static struct io_worker * workers[16] = { NULL };
```

The case is that during testing I've found another issue. TSAN reports data race when doing `workers[i].ba_pcm.running = ba_pcm->running;` which later is accessed in the IO worker. To fix that we need a mutex which will guard IO worker structure. And the initialization will be "nicer" if the structure is allocated on the heap in the `supervise_io_worker_start()` function. Otherwise the `workers` initialization will have to look like:
```
static struct io_worker workers[16] = {
    { .mutex = PTHREAD_MUTEX_INITIALIZER },
    { .mutex = PTHREAD_MUTEX_INITIALIZER },
    ....
    { .mutex = PTHREAD_MUTEX_INITIALIZER },
};
```

Also such approach will be easier to refactor in the future to support more connections.

borine:
OK, I will have a go at that next week.

arkq:
I've done some tests with LDAC (samples S32, rate 96000 Hz) and I've got some "glitch". The logs from `bluealsa-aplay` looks like this:
```
346.393377: D: aplay.c:864: PCM sample rate conversion: 96000 Hz -> 48006.53 Hz
346.594583: D: aplay.c:864: PCM sample rate conversion: 96000 Hz -> 48006.91 Hz
<--- here I've separated BT devices far apart to cause connection instability --->
346.839267: W: alsa-pcm.c:246: ALSA playback PCM underrun
346.839416: D: aplay.c:856: Resetting resampler: ALSA underrun detected
346.839463: D: aplay.c:864: PCM sample rate conversion: 96000 Hz -> 48000.00 Hz
351.930630: D: resampler.c:309: Adaptive resampling target delay: 106.7 ms
351.930697: D: aplay.c:864: PCM sample rate conversion: 96000 Hz -> 47961.60 Hz
```

In the `bluealsactl monitor -p` I can see that the client delay was set to ~340 and is lowered by the resampler at the max allowed rate ( 96000 Hz -> 47961.60 Hz). However, I would suspect that after an underrun I will go to a state where the delay is low, not high... I had no time to check the code yet, I'm just reporting my latest observation.

Other than that everything seems to be OK. Even the CPU load with the `--resampler=best` seems to be lower now - ~40% (versus ~60% when I've been checking it two weeks ago).

borine:
> I would suspect that after an underrun I will go to a state where the delay is low,

Yes, at the moment that the underrun occurs, the **instantaneous** delay is (close to) zero. But the delay value we are reporting is the average over the previous 16 loop iterations. In each iteration the instantaneous delay measurement is taken at the time a chunk of audio is read from the server, so is never less than the chunk size. (The underrun would have occurred earlier, while the thread was blocked in poll())

> I can see that the client delay was set to ~340

This appears to be due to the bluetooth controller. I see a similar result with SBC, AAC and AptX. It seems that when transport frames are lost in the air, then the controller can re-send them. There appears to be a buffer of 300 - 500 ms audio which can build up in poor radio conditions, and then sent in a very short interval once reception improves. The bluealsad decodes these and dumps them to the fifo as fast as possible.

So the ALSA buffer is filled immediately that radio reception re-commences and from that point each loop iteration writes only a few frames and reads only a few frames so that the delay reporting buffer is filled with 16 very large instantaneous delay values.

This large delay reported is "accurate" in the sense that A/V remains synchronized on my phone, after a momentary glitch when radio reception loss interrupts the audio stream.

borine:
> 346.839267: W: alsa-pcm.c:246: ALSA playback PCM underrun

hopefully commit 13282fd will trigger a drain() before this underrun can occur

borine:
> hopefully commit https://github.com/arkq/bluez-alsa/commit/13282fdecee49f33e011a0f0a8c38727206b040c will trigger a drain() before this underrun can occur

Ugh - scratch that - underrun prediction can be improved, but that commit is not it!!!

borine:
commit 4cd792e is hopefully correct, and really does improve underrun prediction. It also restores the 5 second period after alsa starts before adaptive resampling chooses a target delay; this period must have been accidentally removed by a previous commit.

arkq:
> This appears to be due to the bluetooth controller. I see a similar result with SBC, AAC and AptX. It seems that when transport frames are lost in the air, then the controller can re-send them.

Yes, that was my thought as well, but I had no time to check that.

Also, when that happens I can hear "clicks" in the audio. Also I'm not sure about the cause, maybe that's my phone.... anyway, when the delay is reduced to the target value clicks disappear. I will have to test it a little bit more (different codes, different sampling rates, different resampler converters).

And one thing about logging. Now in the logs we have:
```
alsa-pcm.c:268: Draining ALSA playback PCM to avoid underrun
aplay.c:858: Resetting resampler: ALSA underrun detected
```

Do you have any ideas how to reword it to make these logs less confusing ("avoid underrun" -> "underrun detected")? :D

borine:
> how to reword it to make these logs less confusing

How about simply dropping the underrun reference from the aplay.c message? Since the (potential/actual) underrun event has already been logged by alsa-pcm.c, there is no need to repeat it in aplay.c. So just:
```console
aplay.c:858: Resetting resampler
```
is good enough?

borine:
>  "clicks" in the audio

That could be either lost frames in the bluetooth link, or underrun in the DAC. For the DAC the "click" should only happen when it stops (either at underrun or drain), so repeated clicks are more likely caused by dropped frames within the bluetooth chain (eg bluealsad dropping frames when the fifo is full or the AVDTP transport unable to resend all lost frames)

arkq:
> eg bluealsad dropping frames when the fifo is full

That's not the cases because bluealsad prints log in such case, and there are no such logs printed.
I will debug that (I will dump PCM in various places to see since where audio is corrupted).

Also, I've just discovered one thing. I think that something is not right with delay calculation. Right now the delay is calculated by adding frames from various places and dividing it by sampling rate of input PCM. But, with resampler we might have two different rates... Delay reported by ALSA should be divided by out-rate, where internal buffering should be divided by in-rate.

borine:
> something is not right with delay calculation.

oops! I've just looked at delay_report_update() and it does indeed get some details wrong:
- frame size for input fifo is the ba format size, but delay report incorrectly uses alsa format size
- similarly the ba and alsa rates are different.

I think the delay_report_update() function interface needs to change; will look at that tomorrow/friday.

borine:
> why min_target_delay and max_target_delay are different for different Bluetooth codecs

Hmm, the minimum target is the alsa start threshold (because the delay must be at least that much before the alsa device can start to consume frames); the max target is one alsa period bigger that that (simply to have some upper limit, we could choose some other value if necessary). The actual values depend on the capabilities of the alsa device, and may well be different for different rates or formats; the requested values are 150ms for start threshold and 50ms for period, but the device may set its own constraints and choose some other values to comply with those constraints. So if two codecs have the same rate and format, then they should also result in the same start threshold and period. If not then something very unexpected is happening.

borine:
commit d5af218 fixes a log message as suggested in [this comment](https://github.com/arkq/bluez-alsa/pull/750#issuecomment-2755546165)
commit 86e3b76 fixes the resampling delay report issue mentioned in [this comment](https://github.com/arkq/bluez-alsa/pull/750#issuecomment-2755625503)
commit 81aba81 may help prevent "clicks" when the transport re-starts after being idle.

borine:
I've been reading up on sample rate conversion, and now realize that the manual page description of "linear" and "zero-hold" is wrong. "linear" is in fact a synonym for "first-order-hold" and is a higher complexity than "zero-order-hold", generally giving better results but with higher cpu load. (zero-order interpolates between two consecutive samples simply by using the value of the first of the two; first-order interpolates by assuming the wave-form follows a straight line joining the two samples.) The three SINC converters are nth-order-hold, giving much smoother outputs at the cost of much higher mathematical complexity.

Also, I think that the name "zero-order" might be a better choice than "zero-hold" for this option, or perhaps we should use its full name "zero-order-hold". What do you think?

arkq:
> or perhaps we should use its full name "zero-order-hold"

To be honest I'm not sure... It depends how mathematical we would like to be versus the final result heard by the user. Now we could have: "best", "medium", "fastest", "linear", "zero-order-hold" and "none". And the question might be why the "fastest" algorithm is not the fastest after all... (it consumes more CPU then "zero-order-hold"). So, in my opinion either we will grade algorithms by CPU usage (skipping implementation details) like "best", "medium", "good", "fastest", etc. (I think that you will know better which words will suit the best here). Or we will stick with the math nomenclature - use full names for sinc-based ones: "sinc-best", "sinc-fastest", etc. Personally I would stick with the math because the names are already there (provided by libsamplerate).

arkq (approved):

arkq:
Many many thanks for that feature! Today I've tested your latest changes a bit and I think that there are no reasons to keep it as a PR any more :D

Issue #754, opened 2025-02-14
pkg-config is not able to complete configuration


sergey-suloev:
> Please read the [troubleshooting guide](../blob/master/TROUBLESHOOTING.md)
> before raising a new issue.

### Problem

pkg-config is not able to make the full configuration. In order to complete configuration I have to specify a lot of dependencies.

### Reproduction steps

cross-compile for armhf/aarch64

### Setup

> - ubuntu noble

### Additional context

    CFLAGS="-I${R}/usr/include"
    LDFLAGS="-L${R}/usr/lib"

    PKG_CONFIG=${BB_PKG_CONFIG} \
    CC=${BB_GCC} \
    NM=${BB_NM} STRIP=${BB_STRIP} RANLIB=${BB_RANLIB} OBJCOPY=${BB_OBJCOPY} OBJDUMP=${BB_OBJDUMP} AR=${BB_AR} \
    CFLAGS="${CFLAGS}" \
    LDFLAGS="${LDFLAGS}" LIBS="-ltinfo" \
    GIO2_LIBS="-lgio-2.0 -lgobject-2.0 -lgmodule-2.0 -lffi -lz  ${LDFLAGS}" \
    BLUEZ_LIBS="-lbluetooth -lncurses -ltinfo  ${LDFLAGS}" \
    LIBBSD_LIBS="-lbsd -lmd  ${LDFLAGS}" \
    MAKEINFO=/bin/true \
        ./configure --prefix=/usr \
                --enable-shared \
                --host=${BB_PLATFORM} \
                --with-sysroot=${R} \
                --with-dbusconfdir="/usr/share/dbus-1/system.d" \
                --with-alsaplugindir="/usr/lib/alsa-lib" \
                --with-alsaconfdir="/usr/etc/alsa/conf.d" \
                --enable-upower \
                ${BLUEZ_ALSA_EXTRA_PARAMS}

sergey-suloev:
#!/bin/sh
R=/path/to/sysroot
export PKG_CONFIG_SYSROOT_DIR="${R}"
export PKG_CONFIG_LIBDIR="${R}/lib/pkgconfig:${R}/usr/lib/pkgconfig:${R}/usr/share/pkgconfig"
export PKG_CONFIG_PATH=
exec /usr/bin/pkg-config "$@"

sergey-suloev:
user@host: ls -l /path/to/sysroot/usr/lib/pkgconfig/
total 216
[...]

PR #755, opened 2025-02-27
use pkg-config to configure readline for bluealsa-rfcomm

borine:
see #754


arkq:
You need to modify `utils/rfcomm/Makefile.am` as well :)

arkq (approved):

PR #758, opened 2025-03-25
aplay: allocate io_workers dynamically

borine:
fixes a bug whereby a disconnecting device would cause pointers held by io_workers of other devices to become invalid.

To reproduce the error on current sources:
1.start bluealsa-aplay
2. connect device A
3. connect device B
4. start playback on device B
5. disconnect device A
6. re-connect device A

At this point the io_worker thread for device B is broken. Most often it remains blocked in poll(), sometimes it stops with error "Bad file descriptor", and on rare occasions the whole process is terminated by an abort() in libsasound.

### DIFF
--- a/test/test-utils-aplay.c
+++ b/test/test-utils-aplay.c
@@ -168,11 +168,12 @@ CK_START_TEST(test_play_all) {
 	ck_assert_int_ne(spawn_bluealsa_aplay(&sp_ba_aplay,
 				"--profile-a2dp",
 				"--pcm=null",
+				"--volume=none",
 				"-v", "-v",
 				NULL), -1);
 	spawn_terminate(&sp_ba_aplay, 500);
 
-	char output[8192] = "";
+	char output[16384] = "";
 	ck_assert_int_gt(spawn_read(&sp_ba_aplay, NULL, 0, output, sizeof(output)), 0);
 
 	/* check if playback was started from both devices */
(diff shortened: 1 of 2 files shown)


borine:
Instead of introducing a mutex to protect worker->ba_pcm.running as is done here, perhaps it would be more efficient to introduce a new `atomic_bool running` member into the worker struct, initialised to the value of `ba_pcm.running` when the worker is initialised? I have tested some code for this and it appears to work well.

arkq:
For running flag only it would work, however, TSAN reports more data races (e.g. for mixer sync), so it will be easier to suppress/fix them with mutex. Yesterday I started to do some cleanups/fixes to this branch of yours, but unfortunately I've messed up my RPi beyond repair so I had to reflash SD card... I will continue with TSAN fixes today. If you are OK I will push my changes here.

borine:
> If you are OK I will push my changes here

That's fine. I will leave this branch alone until you have your updates ready.

arkq:
I was not able to test the mixer locking properly today, but hopefully I will do it tomorrow.

arkq:
I've checked mixer and it seems that TSAN is happy now. However, I've discovered some issue with volume update, but it's not related to use after free. If you are OK I will push that change to master.

borine:
>  If you are OK I will push that change to master.

commit b0d6714 is working fine for me, so I am OK to merge it.

> I've discovered some issue with volume update

I have a small change to support devices that do not have dB scale info, but will hold that back until you are happy with the existing volume update code.

Issue #757, opened 2025-03-17
In bluealsa-aplay the mixer device is not set by default same as the output device


vale-max:
Hi all,
I finally moved from an old version of bluez-alsa to the latest from source, however I had quite an hard time setting up the hardware mixer in bluealsa-aplay

I have a Raspberry Pi 2 with an USB DAC, I thus have three audio output device.

With the older version I used the following startup command:
`/usr/bin/bluealsa-aplay -S -D plughw:CARD=AUDIO,DEV=0 --mixer-name="SMSL USB AUDIO "`

this worked just fine, however with latest code it failed to open the mixer:
```
/usr/bin/bluealsa-aplay -D plughw:CARD=AUDIO,DEV=0 --mixer-name='SMSL USB AUDIO '

bluealsa-aplay: [2717] D: ../../../utils/aplay/aplay.c:797: Creating IO worker 78:D8:40:82:76:6E
bluealsa-aplay: [2717] D: ../../../utils/aplay/aplay.c:1223: Starting main loop
bluealsa-aplay: [2718] D: ../../../utils/aplay/aplay.c:460: Opening BlueALSA source PCM: /org/bluealsa/hci0/dev_78_D8_40_82_76_6E/a2dpsnk/source
bluealsa-aplay: [2718] D: ../../../utils/aplay/aplay.c:484: Starting IO loop
bluealsa-aplay: [2718] D: ../../../utils/aplay/aplay.c:613: Opening ALSA playback PCM: name=plughw:CARD=AUDIO,DEV=0 channels=2 rate=96000
bluealsa-aplay: [2718] D: ../../../utils/aplay/aplay.c:634: Opening ALSA mixer: name=default elem=SMSL USB AUDIO  index=0
bluealsa-aplay: [2718] W: ../../../utils/aplay/aplay.c:640: Couldn't open ALSA mixer: Mixer element not found
```

After some trial and error I found that this syntax works:
`/usr/bin/bluealsa-aplay -D plughw:CARD=AUDIO,DEV=0 --mixer-name='SMSL USB AUDIO ' --mixer-device='default:CARD=AUDIO'
`
Basically bluealsa-aplay was selecting by default as mixer device the internal audio board rather than the same device used for the output.

While not technically an error I think is a wrong behavior. If it's not specified, is logical to suppose that the device for which we want to change the volume is the same where we output the sound.

arkq:
[...]

PR #760, opened 2025-04-26
Aplay changes

borine:
Changes aimed at clarifying some points made in issue #757

I am not convinced that "guessing" the appropriate mixer or control element would really improve the general user experience. Instead these changes try to explain more clearly what happens when explicit options for mixer device or control name are not given on the command line.

Devices that do not give dB scale TLV data are supported by assuming a simple linear scale; this is the same approach as is used by alsamixer, so should give a familiar experience for users of such devices.

### DIFF
--- a/doc/bluealsa-aplay.1.rst
+++ b/doc/bluealsa-aplay.1.rst
@@ -6,7 +6,7 @@ bluealsa-aplay
 a simple BlueALSA player
 ------------------------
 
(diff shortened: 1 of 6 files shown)


arkq (approved):

arkq:
Thanks! Changes applied with some minor typo fixes (some missing dots).

PR #762, opened 2025-05-08
Multi client support

borine:
Permit more than one client to open each PCM. For playback clients, the streams from each client are mixed together before encoding. For capture clients each client receives a copy of the decoded stream.

Stream mixing is done in C code, there is no explicit SIMD code. However, compiling with `-O3` does appear to allow `gcc` to optimize the mix loop very well on x86_64. I have not tested optimization with aarch64 (my old RPi machines are all 32-bit with no SIMD); but even without optimization I can run 3 playback clients simultaneously on a RPi zero W.

There was some interest in this feature three years ago, but there have been only a few enquiries more recently. I think this is pushing the scope boundary a little, as for most systems requiring multiple simultaneous audio applications there are also other requirements which mean that `pipewire` is probably a better fit. However I do have one genuine use case which is a "kiosk" system running only `chromium` which requires each chromium tab to independently open the PCM; and for this system BlueALSA with multi-client support is ideal.

### DIFF
--- a/.github/iwyu.imp
+++ b/.github/iwyu.imp
@@ -53,4 +53,6 @@
 	{ include: [ '<spandsp/g722.h>', private, '<spandsp.h>', public ] },
 	{ include: [ '<spandsp/plc.h>', private, '<spandsp.h>', public ] },
 
+	{ include: [ '<bits/types/struct_itimerspec.h>', private, '<sys/timerfd.h>', public ] },
+
 ]
--- a/.github/spellcheck-wordlist.txt
+++ b/.github/spellcheck-wordlist.txt
@@ -177,6 +177,8 @@ BTRH
 BTT
 bttransport
 btusb
+# aspell considers PIPE_BUF to be 2 words :(
+BUF
 CCE
 CIEV
 CIND
@@ -205,6 +207,7 @@ eoc
 EP
 EPIPE
 EPMR
+epoll
 EQMID
 errno
 FB
--- a/.github/spellcheck-wordlist.txt
+++ b/.github/spellcheck-wordlist.txt
@@ -179,6 +179,8 @@ BTRH
 BTT
 bttransport
 btusb
+# aspell considers PIPE_BUF to be 2 words :(
+BUF
 CCE
 CIEV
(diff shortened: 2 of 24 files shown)


arkq:
I've tried to cherry-pick the `Ensure encoder buffers are flushed when draining` commit to master (the code with some cleanups is here: https://github.com/arkq/bluez-alsa/tree/drain), but when testing I've spotted that in case of `mpg123` and `cmus` (and maybe some others) the final effect is not satisfactory. The problem is that these players, when stoping, trigger `drop` -> `resume` -> `drain` -> `drop` actions. And because of that my headset plays a click when silence is received. That happens with SBC and AAC codecs (I have not tested other codes yet, but I guess that the effect will be the same). Have you ever spotted some "clicks" when stopping playback (when using this drain flushing code)? Later today I will check with some other sinks. The problem does not occur with `aplay`, because it does not drop PCM before drain.

arkq:
I'm testing on Debian 12:
```console
$ mpg123 --version
mpg123 1.31.2
```

I've checked with [upstream master](https://github.com/madebr/mpg123) version of mpg123:
```console
$ src/mpg123 --version
mpg123 1.33.0-dev
```

and the outcome is the same - click at the end.

Here is the [calltrace](https://github.com/arkq/etrace) from the termination:
```
| safe_exit
| | dump_close
| | | play_prebuffer
| | out123_drop
| | audio_cleanup
| | out123_del
| | | out123_close
| | | | out123_drain
| | | | | out123_pause
| | | | out123_stop
| | | out123_set_buffer
| | | | out123_close
| | | | | out123_drain
| | | | | out123_stop
| | mpg123_delete
| | | mpg123_close
| | mpg123_exit
| | stream_close
| | split_dir_file
| | term_exit
| | | term_have_fun
| | | term_restore
```

So, at least based on function names, it seems that `mpg123` calls drop then it closes the output (which calls drain, then stop). 

I think that a workaround would to to track whether we have some samples after the drop, and if there are no new samples, drain is not required. I'll try to add that to the IO code and will think about it whether it will not cause some undesired effects.

PR #766, opened 2025-05-18
ALSA PCM plugin: use transfer() callback

borine:
As discussed in #745, using the ALSA PCM transfer() callback instead of allowing the ioplug to create and manage the ring buffer fixes the TSAN test failure.

Contrary to my supposition in #745, locking the pcm mutex within the transfer() function is not necessary, simply the fact that the bluealsa plugin now manages all access to the ring buffer means that thread synchronization is correct.


borine:
oops, sorry, too hasty - passes test suite but is not working!

PR #767, opened 2025-05-18
ALSA PCM plugin: use transfer() callback

borine:
As discussed in https://github.com/arkq/bluez-alsa/pull/745, using the ALSA PCM transfer() callback instead of allowing the ioplug to create and manage the ring buffer fixes the TSAN test failure.

### DIFF
--- a/src/asound/bluealsa-pcm.c
+++ b/src/asound/bluealsa-pcm.c
@@ -90,7 +90,10 @@ struct bluealsa_pcm {
 	int event_fd;
 
 	/* virtual hardware - ring buffer */
-	char * _Atomic io_hw_buffer;
+	char *io_hw_buffer;
+	/* channel areas on top of the ring buffer */
+	snd_pcm_channel_area_t *io_hw_areas;
+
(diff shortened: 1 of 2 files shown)


src/asound/bluealsa-pcm.c
```diff
@@ -622,12 +628,30 @@ static int bluealsa_hw_params(snd_pcm_ioplug_t *io, snd_pcm_hw_params_t *params)
 
 	pcm->frame_size = (snd_pcm_format_physical_width(io->format) * io->channels) / 8;
 
+	if ((pcm->io_hw_buffer = malloc(io->buffer_size * pcm->frame_size)) == NULL) {
```
arkq, line 631: `buffer_size * frame_size` does not seem right...

arkq: OK, forget about that, my bad, everything is OK.

arkq:
I've rebased it on top of master with small style changes to align with the rest of the codebase.

arkq (approved):

PR #768, opened 2025-05-27
Fixes for gcc 14

borine:
I've just completely rebuilt my development machine with ubuntu 25.04, and now find some bluez-alsa build errors. This PR fixes the build (with all codecs and utilities enabled).

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -292,7 +292,6 @@ AC_ARG_ENABLE([systemd],
 	AS_HELP_STRING([--enable-systemd], [enable systemd integration]))
 AM_CONDITIONAL([ENABLE_SYSTEMD], [test "x$enable_systemd" = "xyes"])
 AM_COND_IF([ENABLE_SYSTEMD], [
-	PKG_CHECK_MODULES([SYSTEMD], [systemd >= 200])
 	AC_DEFINE([ENABLE_SYSTEMD], [1], [Define to 1 if systemd is enabled.])
 ])
 
--- a/src/a2dp-faststream.c
+++ b/src/a2dp-faststream.c
@@ -121,13 +121,14 @@ static struct a2dp_caps_helpers a2dp_fs_caps_helpers = {
 void *a2dp_fs_enc_thread(struct ba_transport_pcm *t_pcm) {
 
 	pthread_setcancelstate(PTHREAD_CANCEL_DISABLE, NULL);
-	pthread_cleanup_push(PTHREAD_CLEANUP(ba_transport_pcm_thread_cleanup), t_pcm);
 
 	struct ba_transport *t = t_pcm->t;
 	struct io_poll io = { .timeout = -1 };
 
 	/* determine encoder operation mode: music or voice */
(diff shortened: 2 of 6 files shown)


macmpi:
FWIW Alpine edge branch has moved to GCC 15 (heading to Alpine 3.23), and reports `bluez-alsa` 4.3.1 (for now) as one of many packages [failing to rebuild](https://gitlab.alpinelinux.org/alpine/aports/-/work_items/17311).
Have not checked the details yet, nor if these changes address all issues.

borine:
> reports bluez-alsa 4.3.1 (for now) as one of many packages failing to rebuild.

As noted in https://gcc.gnu.org/gcc-15/porting_to.html the C23 standard makes some significant semantic changes to existing syntax, and this has a widespread impact over many of the dependencies of BlueALSA. So all those dependencies will have to be updated before BlueALSA can be compiled, even before we consider any necessary changes to the BlueALSA code.

In the interim, if anyone is building BlueALSA on a system with gcc 15, a quick workaround is to add `CFLAGS="--std=gnu23"` to the `make` command line.

arkq:
> and reports bluez-alsa 4.3.1 (for now) as one of many packages failing to rebuild.
> and it fails when compiling utils/rfcomm/rfcomm.c because of a warning from the libreadline header. 

Yes, that's the only error that I can see (on Ubuntu 24.04 using upstream gcc-15). However, I guess that the problem is on the libreadline side, because is should be valid to call `rl_message("some const string")`...

@borine as for `-Wclobbered`, I wonder whether it would be possible to rearrange the code so the compiler will not produce any clobbered warnings... I've seen this warning since long time ago (only when building with `-O2` though), but since it was not occurring on CI I've just ignored it. There were more warnings like that, but I've "fixed" them by changing slightly the place where some variables are defined or used (in respect to `pthread_cleanup_push`).

arkq:
> I'm not sure what line is for anyway? bluealsa-rfcomm seems to work fine when I just comment out that call to rl_message("").

There is small difference (maybe it can be achieved in a different way, though) where the "disconnected" message looks like a user command:

#### with `rl_message("")`
```console
$ bluealsa-rfcomm /org/bluealsa/hci0/dev_04_5D_4B_ED_D7_00 
> AT+IPHONEACCEV=1,1,6
04:5D:4B:ED:D7:00> RING
disconnected
```

#### without `rl_message("")`
```console
$ bluealsa-rfcomm /org/bluealsa/hci0/dev_04_5D_4B_ED_D7_00 
> AT+XAPL=054C-0C75-0452,0
04:5D:4B:ED:D7:00> RING
04:5D:4B:ED:D7:00> disconnected
```

borine:
OK, thanks for the explanation, I had not spotted that difference. So rl_message("") is used only for its accidental side-effect of overwriting the prompt buffer (the docs say to save the prompt before using rl_message() and then restore it afterwards). I've pushed a new commit which explicitly clears the prompt, which I think is a more readable solution.

arkq (approved):

PR #769, opened 2025-06-05
aplay: prevent  busy loop

borine:
Handle the situation where a BT device is in the "Running" state but delivers no audio.
Make sure that the worker thread blocks in poll() when there is no work to do.


arkq:
Thanks!

PR #800, opened 2026-05-09
Avoid NULL pointer deref when libsandbox is in LD_PRELOAD

floppym:
libsandbox[1] calls dlopen() inside a constructor function. If this constructor happens to run before the init() function in libaloader.c, dlopen_orig will still be NULL and the program will crash with a segfault.

Avoid this by calling dlsym() inside the dlopen() wrapper itself when necessary.

[1] https://github.com/gentoo/sandbox

Bug: https://bugs.gentoo.org/973030

### DIFF
--- a/test/libaloader.c
+++ b/test/libaloader.c
@@ -67,6 +67,9 @@ void *dlopen(const char *filename, int flags) {
 	if (strstr(filename, "libasound_module_pcm_bluealsa.so") != NULL)
 		filename = strcat(tmp, "/src/asound/.libs/libasound_module_pcm_bluealsa.so");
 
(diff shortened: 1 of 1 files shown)


floppym:
I'm kind of baffled by the test failures; I cannot reproduce them.

I pushed a revert to see if this is some issue with the Github CI environment, but if you have a better idea of what is wrong please let me know.

arkq:
The failure might be just flakiness. I've re run failed tests. I had no time to check that earlier. Change looks OK, so I do not expect any issues with that PR.

arkq (approved):

PR #808, opened 2026-07-24
Use alsa-lib logging for ALSA plugins

borine:
alsa-lib release 1.2.15 (December 2025) improved its logging API to permit filtering of messages using the environment variable LIBASOUND_DEBUG. This PR takes advantage of that to permit filtering of BlueALSA plugin messages. This can be very useful, especially for those distributions whose bluealsa packages are built with debug enabled. It does however mean that debug messaging now has to be explicitly enabled at run time whereas previously a debug build would log by default.

Another bonus of this change is that BlueALSA plugin messages are now written to the same log handler as other ALSA messages, allowing the application to choose its most appropriate logging target.

For earlier alsa-lib releases this PR introduces a new environment variable BLUEALSA_LOG_LEVEL to allow filtering of messages with the same scheme as is used by the `--loglevel=` option of `bluealsad` and `bluealsa-aplay`.

### DIFF
--- a/.github/workflows/build-and-test.yaml
+++ b/.github/workflows/build-and-test.yaml
@@ -29,7 +29,7 @@ jobs:
       fail-fast: false
     runs-on: ubuntu-22.04
     steps:
-    - uses: actions/checkout@v6
+    - uses: actions/checkout@v7
     - name: Install Dependencies
       uses: ./.github/actions/apt-install-deps
     - name: Create Build Environment
@@ -58,7 +58,7 @@ jobs:
   sanitize-prepare:
     runs-on: ubuntu-22.04
     steps:
-    - uses: actions/cache@v5
+    - uses: actions/cache@v6
       id: cache
       with:
         key: sanitize-env
@@ -118,7 +118,7 @@ jobs:
     needs: sanitize-prepare
     runs-on: ubuntu-22.04
     steps:
-    - uses: actions/checkout@v6
+    - uses: actions/checkout@v7
     - name: Install Dependencies
       uses: ./.github/actions/apt-install-deps
     - name: Create Build Environment
@@ -153,7 +153,7 @@ jobs:
         make clean
         SANITIZERS=$(for x in ${{ matrix.sanitize }}; do echo -n " -fsanitize=$x"; done)
         make check CFLAGS="-g -O2 $SANITIZERS -fno-omit-frame-pointer -fno-sanitize-recover=all" TESTS=
-    - uses: actions/cache/restore@v4
+    - uses: actions/cache/restore@v6
       with:
         key: sanitize-env
         path: |
--- a/.github/workflows/code-scanning.yaml
+++ b/.github/workflows/code-scanning.yaml
@@ -21,7 +21,7 @@ jobs:
   cccc:
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v6
+    - uses: actions/checkout@v7
     - name: Run cccc analysis
       uses: sarnold/cccc-action@main
     - uses: actions/upload-artifact@v7
@@ -32,7 +32,7 @@ jobs:
   code-ql:
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v6
+    - uses: actions/checkout@v7
     - name: Install Dependencies
       uses: ./.github/actions/apt-install-deps
     - name: Initialize CodeQL
@@ -77,7 +77,7 @@ jobs:
   doc8-lint:
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v6
+    - uses: actions/checkout@v7
     - name: Run reStructuredText Linter
       uses: deep-entertainment/doc8-action@v5
       with:
@@ -86,7 +86,7 @@ jobs:
   include-what-you-use:
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v6
+    - uses: actions/checkout@v7
     - name: Install Dependencies
       uses: ./.github/actions/apt-install-deps
     - name: Create Build Environment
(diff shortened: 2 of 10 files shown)


arkq:
How the `SNDERR` is handled in alsa-lib >= 1.2.15 ? If it is forwarded to `snd_lib_log(SND_LOG_ERROR, ...)`  I would opt for creating `logging.h` (or something) in the `src/asound` directory which will wrap `SNDERR` into `error` macro for < 1.2.15 and use `snd_lib_log(SND_LOG_ERROR` for >= 1.2.15. Then new logging "system" can be used uniformly in the PCM and CTL plugings.

borine:
OK, I will make changes this weekend. I like the new AI review!

> How the SNDERR is handled in alsa-lib >= 1.2.15 ? If it is forwarded to snd_lib_log(SND_LOG_ERROR, ...)

Yes, see https://github.com/alsa-project/alsa-lib/blob/master/include/error.h

borine:
logging.h still needs a little more work, because it needs to distinguish PCM and MIXER interfaces. I will look at it again later this weekend. The `debug2_params()` macro is also redundant.

arkq:
> logging.h still needs a little more work, because it needs to distinguish PCM and MIXER interfaces.

That's because the ALSA API is different for PCM and mixer? Or maybe there some other reason? My idea about putting logging wrappers in a separate header file was because I thought that these macros can be shared, if that's not the case, skip that idea. Keeping them in the C files is OK for me (if that's the case) :D

arkq:
If the problem is `debug2` which uses `pcm` internally, maybe these "wrappers" should be abandoned... or kept in the C file?

borine:
The thing I forgot is that my code uses `SND_ILOG_PCM` as a fixed argument to snd_lib_log(). That does not make sense for the CTL plugin. Its a shame that there is no `SND_ILOG_IOPLUG` but we have to work with what we are given. As for `debug2()` it is not used in bluealsa-ctl.c, but perhaps it should be renamed `debug_pcm()` just for clarity. I have to leave this now but will return to it in a few days.

arkq:
I've just checked the https://www.alsa-project.org/alsa-doc/alsa-lib/group___error.html documentation. How about using `snd_error()` and other related macros for alsa-lib which supports them, and add a compatibility layer for versions without support? This way our plugins will use native logging API. The problem with missing `SND_ILOG_IOPLUG` still remains, but IMHO from the user point of view it does not matter whether the log comes from a software plugin or a core component. The selected interface should be based on the ALSA subsystem functionality which in our case is SND_ILOG_CONTROL I guess. Also, I can see that there is SND_ILOG_CONFIG, so that also can be used for problems with arguments parsing. All in all, the `snd_error()` seems to be the way, because it allows the user to select the interface when calling the macro: `snd_error(PCM, "Foo")` What do you think?

borine:
> snd_error() seems to be the way

I was thinking the same. Used directly that would lose the abililty to log the pcm path (which I find useful), so my preference would be to use local wrappers around the alsa-lib calls for the pcm plugin at least. I've just noticed that `0` can be used in place of a SND_ILOG_* value to select global scope for the message. This might be useful because, although it's not mentioned in the doxygen docs, ASOUND_DEBUG supports a syntax for filtering by interface as well as priority: `[<level>][,<interface1>:<level1>][,<interface2>:<level2>,...`.

For example:
```
`ASOUND_DEBUG=error,pcm:debug,control:info`
```
sets global logging to error, pcm logging to debug and control logging to info.

So perhaps we could make bluealsa messages all global, that way the specific interface identifiers would apply only alsa-lib internal messages which are generally only of interest to alsa-lib developers?

arkq:
> so my preference would be to use local wrappers around the alsa-lib calls for the pcm plugin at least.

Yes, that's reasonable. Maybe something like `snd_pcm_debug()`. However, I think that it would be better to pass the `pcm` in an explicit way.

> So perhaps we could make bluealsa messages all global, that way the specific interface identifiers would apply only alsa-lib internal messages

Hmm... I'm not sure about that. From the user point of view it does not matter whether the log comes from the plugin or internal library. PCM is a PCM, so if user would like to see all logs related to PCM, then they should select `pcm:debug`, I think.... Anyway, if you think that global messages are better, then go with global messages :)

src/asound/logging.h
```diff
@@ -0,0 +1,67 @@
...
+#if SND_LIB_VERSION >= 0x01020F
+
+# if ! DEBUG
+#  ifdef snd_debug
+#  undef snd_debug
+#  endif
+#  define snd_debug(...) do {} while (0)
+# endif
+
+# define logging_init() do {} while (0);
```
arkq: Is it required to define a "body" for the "empty" macro? If so, the ';' at the end should be removed. The whole idea of using `do {} while()` is for enforcing the usage of `;` after the macro.

src/asound/logging.h
```diff
@@ -0,0 +1,67 @@
...
+#if DEBUG
+# define snd_debug(interface, ...) do { \
+		if (ba_snd_log_level <= BA_LOG_DEBUG) \
+			snd_lib_error(__FILE__, __LINE__, __func__, 0, "[debug] " __VA_ARGS__); \
+	} while (0)
+#else
+# define snd_debug(...) do {} while (0)
+# endif
+
+extern int ba_snd_log_level;
```
arkq: Maybe use a name which aligns better with the `logging_init` and does not suggest that this variable is something that one should use outside of this file. As I see it, it is an internal variable for the logging module, but since it is used in macros, it has to be marked as "external". Am I right?

arkq (dismissed):
It looks nice!

src/asound/logging.h
borine: correct. Ive pushed a fixup commit that includes all the copilot recommendations too.

borine:
Having thought about this a little more, perhaps it would be better if the plugins has the same behaviour (as far as possible) with all versions of alsa-lib. In that case we should make "error" the default message priority threshold for `BLUEALSA_LOG_LEVEL` to match alsa-libs's `LIBASOUND_DEBUG`. Is that a sensible idea?

arkq:
> In that case we should make "error" the default message priority threshold for BLUEALSA_LOG_LEVEL to match alsa-libs's LIBASOUND_DEBUG. Is that a sensible idea?

Yes, that's reasonable. After all for a normal user any message other than error is non essential :D

